Issue #17, opened 2014-12-31
Specifying the output type of an interpolation object


tlycken:
One of the issues with Grid.jl was that it's hard to specify what type the interpolation object should return - Interpolations.jl should be able to do better if we want it to be easily usable with any number type.

I'm leaning toward an interface with an optional first type argument, that specifies the output type, similar to what `round` and friends do in 0.4:

```
julia> itp = Interpolation(Float64, 1:10, Linear(OnGrid()), ExtrapNaN())

julia> typeof(itp[3.2])
Float64
```

Of course, we need to provide some sensible default, and there are a couple of options:

* Use the element type of the input array, i.e. `Interpolation(A, config...) = Interpolation(eltype(A), A, config...)`
  Pros:
    - very predictable for the end user; effectively, `eltype(Interpolation(A,...)) == eltype(A)`

  Cons:
     - will have the same problems as Grid.jl, especially for integer inputs (but these are avoided if the user explicitly sets a type)
     - we have to `convert` each return value, so there must be a conversion defined from whatever our calculations promote to, to the return type.

* Use use some sensible promotion result, e.g. `promote_type(eltype(A), Float64, typeof(x)...)` (where the `Float64` is from the literals in the prefiltering step - we might be able to avoid those, if we can require a `convert` method for converting floats to `eltype(A)`)

  Pros:
    - very similar to what we do today
    - won't require anything extra from the user to work with integers, e.g. `itp = Interpolation(1:10, ...)`

  Cons:
    - might be difficult to get it right, since there are lots of places where numbers enter the calculations
    - might seem unpredictable to the user, especially since e.g. `promote_type(Rational{Int}, Float64) == Float64` will make us return floats instead of rationals for `Interpolation(A::Array{T<:Rational}, ...)`

I honestly don't know what's best here, but before we start implementing this, we need some sort of decisions on:

* What should the API look like? (Is the above suggestion good or ugly?)
* How should we choose a default?

tlycken:
Thinking another round on it after asking for help in #19 just now, I'm starting to believe converting to `TCoefs` isn't such a good idea. For "duck type-based multivalued interpolation", we want to allow e.g. `TCoefs<:AbstractArray`, but indexing with an array doesn't make sense. 

I think we need to do some more bikeshedding...

[...]

PR #18, opened 2014-12-31
Gradient evaluation

tlycken:
I got it working! =)

The interface is currently a little clunky, especially in 1D, but I couldn't think of a way to define an overload of `gradient` for vector interpolations, that wouldn't be recursive. This is how it works:

```
using Interpolations
xg = 1:15
xf = 1:.1:15
f1(x) = sin(2pi/15 * (x-1))
g1(x)= 2pi/15 * cos(2pi/15*(x-1))

f1g = f1(xg)
g1g = g1(xg)

itp1 = Interpolation(f1g, Quadratic(Free(),OnCell()), ExtrapNaN())
f1i = Float64[itp1[x] for x in xf]
g1i = Float64[gradient(itp1, x)[1] for x in xf]

using Gadfly
plot(
	layer(x=xg,y=fg,Geom.point),
	layer(x=xg,y=gg,Geom.point,Theme(default_color=color("red"))),
	layer(x=xf,y=f1i,Geom.path),
	layer(x=xf,y=g1i,Geom.path,Theme(default_color=color("red"))),
)
```

gradient

It works for linear and constant interpolations too, although the results are of course less useful:

gradient-constant
gradient-linear

The meta-programming mumbo-jumbo used to define the correct coefficients for evaluation became more complicated, and as a result (other than the code being a little more opaque), the loading time of the package is bumped up even further; `tic(); using Interpolations; toc()` now reports 31 seconds on my machine. For comparison, `tic(); using Gadfly; toc()` reports 25... It might be possible to reduce the load time slightly by organizing the code into smaller modules where things like `symm`, `sym` and `symp` could be defined as constants, but I guess much of this would be alleviated with staged functions, so I'm not very concerned about it at the moment.

### DIFF
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -3,7 +3,12 @@ module Interpolations
 using Base.Cartesian
 using Compat
 
-import Base: size, eltype, getindex, ndims
+import Base:
+    eltype,
+    gradient,
+    getindex,
+    ndims,
+    size
 
 export 
     Interpolation,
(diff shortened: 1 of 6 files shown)


tlycken:
Good points - we probably do want `gradient(itp{T,1}, x)` to return an 1-element array.

I've now made the bulk method `gradient!`, which calculates into a pre-allocated array, and created a stub for `gradient` just as you suggested. It doesn't handle types very gracefully (it just takes `T` from the interpolation object...) but I'm thinking it's easier to fix that all at once when I take care of #17.

PR #22, opened 2015-01-03
Support quadratic interpolation of Rational types

timholy:
Note this is a pull request against your `handling-of-types` branch, not `master`.

The only real problem with Rationals was we were getting (on julia 0.4) `Overflow` errors. That turns out to be because any little floating-point imprecision was causing conversion to "big" rationals:
```julia
julia> convert(Rational, 0.4)
3602879701896397//9007199254740992
```

### DIFF
--- a/src/quadratic.jl
+++ b/src/quadratic.jl
@@ -36,9 +36,9 @@ function coefficients(q::Quadratic, N, d)
     symm, sym, symp =  symbol(string("cm_",d)), symbol(string("c_",d)), symbol(string("cp_",d))
     symfx = symbol(string("fx_",d))
     quote
-        $symm = convert(TIndex, .5 * ($symfx - .5)^2)
-        $sym  = convert(TIndex, .75 - $symfx^2)
-        $symp = convert(TIndex, .5 * ($symfx + .5)^2)
+        $symm = convert(TIndex, 1//2 * ($symfx - 1//2)^2)
+        $sym  = convert(TIndex, 3//4 - $symfx^2)
+        $symp = convert(TIndex, 1//2 * ($symfx + 1//2)^2)
     end
 end
 
@@ -46,9 +46,9 @@ function gradient_coefficients(q::Quadratic, N, d)
     symm, sym, symp =  symbol(string("cm_",d)), symbol(string("c_",d)), symbol(string("cp_",d))
     symfx = symbol(string("fx_",d))
     quote
-        $symm = convert(TIndex, $symfx - .5)
+        $symm = convert(TIndex, $symfx - 1//2)
         $sym = convert(TIndex, -2 * $symfx)
-        $symp = convert(TIndex, $symfx + .5)
+        $symp = convert(TIndex, $symfx + 1//2)
     end
 end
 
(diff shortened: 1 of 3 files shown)


src/quadratic.jl
```diff
@@ -36,9 +36,9 @@ function coefficients(q::Quadratic, N, d)
     symm, sym, symp =  symbol(string("cm_",d)), symbol(string("c_",d)), symbol(string("cp_",d))
     symfx = symbol(string("fx_",d))
     quote
-        $symm = convert(TIndex, .5 * ($symfx - .5)^2)
-        $sym  = convert(TIndex, .75 - $symfx^2)
-        $symp = convert(TIndex, .5 * ($symfx + .5)^2)
+        $symm = convert(TIndex, 1//2 * ($symfx - 1//2)^2)
```
tlycken: Is specifying these literals as `Rational`s better (than `Float64`s) in some general sense, or does it just happen to work better with `Rational` indices/coefficients? Might there be other cases where we want/need to specify these literals in yet another type to make it work nicely?

test/quadratic.jl
```diff
@@ -63,6 +63,11 @@ for x in 2:xmax-1
     @test_approx_eq A[x] itp3[x]
 end
 
+# Rational element types
+A = Rational{Int}[x^2//10 for x in 1:10]
+itp = Interpolation(A, Quadratic(Free(),OnCell()), ExtrapNaN())
+@test itp[11//10] == (11//10)^2//10
```
tlycken: This test fails on 0.3. Any idea why?

src/quadratic.jl
timholy: It's a good question. Until proven otherwise :smile:, my thinking is that since
```julia
julia> promote_type(Rational{Int}, Float64)
Float64
```
we don't lose anything by specifying them as Rationals.

test/quadratic.jl
timholy: hmm, it's very strange. On julia 0.4:
```
julia> itp.coefs
12-element Array{Rational{Int64},1}:
  -1//40                                                                                                                                                                                    
   3//40                                                                                                                                                                                    
   3//8                                                                                                                                                                                     
   7//8                                                                                                                                                                                     
  63//40
  99//40
 143//40
  39//8 
  51//8 
 323//40
 399//40
 483//40
```
On julia 0.3:
```
julia> itp.coefs
12-element Array{Rational{Int64},1}:
  6899716117846699271//8619614680507807464
  7936969560604716305//5916187127003196984
  2250090984434330863//6323126675786419496
 -4372087249599443041//7183461070236085576
 -7973312573468218199//2373339102190889880
  9203035543489481799//6118447631419504648
   336340586073110987//6949707296401608   
  4469594703982377491//8584938456800293288
 -3056145513736821459//2910278932930701560
  -599620387571391593//1277212073810152584
 -5424053943932394579//786135081844975736 
 -8232532661296565515//432728054263014168 
```
My strong suspicion is that something changed (for the better) in the linear algebra code in julia 0.4, probably specifying constants as Rationals in whatever solver gets called in the prefiltering step.

Also on 0.4, any conversion that causes an overflow will result in an error, whereas on 0.3 it will silently (and incorrectly) succeed.

tlycken: Yet another reason to drop 0.3 support, then...

src/quadratic.jl
tlycken: Probably not. And besides, we're converting to `TIndex` anyway (which currently defaults to `Float64` for everything except `Rational` and `Float32`) so I guess I'm just paranoid.

This might be something to take up in the manual though, if/when there is a "best practices" section. I'll take a look and see what I can find...

tlycken:
Although this currently fails on 0.3, I think we're dropping 0.3 soon enough anyway that this is a better approach than what I have over in `handling-of-types`; merging.

PR #23, opened 2015-01-04
Linear extrapolation

tlycken:
Now that `gradient` works, it was a piece of cake to also fix the broken linear extrapolation.

I've pushed [an update of the plotting examples notebook](http://nbviewer.ipython.org/github/tlycken/Interpolations.jl/blob/linear-extrap/doc/Plotting%20examples.ipynb) too, that showcases this. It even looks nice in 2D =)

I'm not sure that we're handling the corners (i.e. extrapolation in more than one dimension simultaneously) in the best way possible (we're ignoring them and hoping for the best...) but it might be good-enough; in higher dimensions it's quite difficult to define exactly what behavior we'd want anyway.


timholy:
That's pretty sweet!

BTW, some of the text gets truncated in the notebook. I don't know what to do about that, though (I'm not a big IJulia user).

tlycken:
Re: text truncation, that's why [I asked about printing of types](https://groups.google.com/forum/#!topic/julia-users/Vm75QxFjtCs) =) I'll probably update the notebook to print unqualified typenames, and then the text won't be truncated (since it'll fit).

Issue #24, opened 2015-01-05
Vector-valued evaluation


tlycken:
Looking at #5, this is one of very few things left to do before we might call ourselves ready for a version 0.1.

I still think this makes sense only in a multi-linear context; the following is probably a good method definition for that:

```julia
getindex{T,N}(itp::Interpolation{T,N}, NTuple{N,AbstractVector}...)
```

Of course, it's going to have to be more specific in order to avoid ambiguity errors. I'm thinking we might not have to specify the type of the vector elements at all - we'll get method errors in the next stage anyway, if the user supplies some nonsensical index vector.

timholy:
Agreed.

This is really exciting!

tlycken:
Is there any way to do this into a pre-allocated array? (Do we want to?) If so, what should the method signature look like?

timholy:
It could look like this:
```
getindex(itp, I::AbstractVector...) = getindex!(Array(eltype(itp), map(length, I)), itp, I...)
function getindex!(dest, itp, I::AbstractVector...)
    # real function goes here
end
```
There are some models of this in `base/multidimensional.jl`.

tlycken:
Nice! I ended up using a promotion rule like this to define the array:

```julia
eval(ngenerate(:N, :(Array{T, N}), :(getindex{T,N,TCoefs}(itp::Interpolation{T,N,TCoefs,$IT,$EB}, x::NTuple{N,Any}...)),
    N->quote
        Tout = promote_type(T, map(eltype, @ntuple $N x)...)
        dest = Array(T, map(length, @ntuple $N x)...)
        getindex!(dest, itp, x...)
    end
))
```

However, I couldn't use the same expression to define the return type (that's why the above says `Array{T,N}` rather than `Array{<expr for TOut>,N}`) because `N` doesn't seem to be defined there. Does it matter that the declared return type for `ngenerate` isn't exactly correct?

Also, when trying to resolve all the ambiguity errors, I want to `ngenerate` methods for `N>=2` to disallow linear indexing into multi-dimensional interpolation objects, but I can't figure out a way to accomplish that. Any suggestions?

tlycken:
Leaving that here (as a mental note to myself, mostly, so I can see if I can take a stab at this if I get some quality Julia time this weekend :smile: ):

[...]

PR #26, opened 2015-02-17
WIP: convert to stagedfunctions

### LINKED ISSUE #21: Startup times & stagedfunctions
The load time for the package are fairly long, largely because lots of methods need to be generated. The load time could probably be drastically shortened using stagedfunctions, which allow the methods to be generated on-demand.

If you get to the point where you are ready to switch your development to 0.4, I will gladly convert things over to using stagedfunctions.

timholy:
This is a WIP to fix #21. It mostly works, and has a massively-beneficial impact on package loading time, but I am getting a truly bizarre error. I inserted some printing code to illustrate the problem:
```julia
julia> include("runtests.jl")
Testing Linear interpolation in 1D...
Testing Linear interpolation in 2D...
Testing Quadratic interpolation in 1D...
Testing Quadratic interpolation in 2D...
Testing evaluation on grid and boundary in 1D...
Testing evaluation on grid and boundary in 2D...
Testing evaluation on grid and boundary in 3D...
Testing gradient evaluation
g = [-0.0]
g = [-0.0]
g = [-0.0]
g = [-0.0]
g = [-0.0]
g = [-0.0]
g = [-0.0]
g = [-0.0]
g = [0.0]
g = [0.0]
g = [0.0]
g = [0.0]
g = [0.0]
g = [0.0]
g = [0.0]
g = [0.0]
g = [-0.0]
g = [-0.0]
g = [-0.0]
g = [-0.0]
g = [0.15043281484715354]
gradient!(g,itp1,x) = [-0.6782138028608058,-0.99190379965505,-0.8414709848078965,-0.2973045442608362,0.3859739967178289,0.8886510150090671,0.9755183471216421,0.6059298029372212,-0.04718003020117026]
g = [0.15043281484715354]
g = [0.15043281484715354]
ERROR: LoadError: LoadError: assertion failed: |g1(x) - (gradient!(g,itp1,x))[1]| <= 0.015353177135101162
  g1(x) = 0.15353177135101162
  (gradient!(g,itp1,x))[1] = 1
  difference = 0.8464682286489884 > 0.015353177135101162
 in error at error.jl:20
 in test_approx_eq at test.jl:137
 in anonymous at no file:27
 in include at ./boot.jl:249
 in include_from_node1 at ./loading.jl:128
 in include at ./boot.jl:249
 in include_from_node1 at ./loading.jl:128
while loading /home/tim/.julia/v0.4/Interpolations/test/gradient.jl, in expression starting on line 24
while loading /home/tim/.julia/v0.4/Interpolations/test/runtests.jl, in expression starting on line 16
```
Upon further investigation, despite [the code](https://github.com/tlycken/Interpolations.jl/blob/6cea519adfe9af4e7e118cb75f780a2302dc3999/src/Interpolations.jl#L222) `gradient!` is returning `itp1.coefs` rather than `g`, but only from the direct `gradient!` call and not from the `gradient` wrapper (which just calls `gradient!`). Also note that there's an inconsistency in ` (gradient!(g,itp1,x))[1] = 1` with `gradient!(g,itp1,x) = [-0.6782138028608058,-0.99190379965505,-0.8414709848078965,-0.2973045442608362,0.3859739967178289,0.8886510150090671,0.9755183471216421,0.6059298029372212,-0.04718003020117026]`.

I got to the point where I can't work on this any more tonight, and I'm pretty baffled. Any ideas? I suspect this is a bug in julia proper.

### DIFF
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -174,7 +174,7 @@ end
 # Resolve ambiguity with real multilinear indexing
 stagedfunction getindex{T,N,TCoefs,IT,EB}(itp::Interpolation{T,N,TCoefs,IT,EB}, x::Real...)
     n = length(x)
-    n == N || return :(error("Must index $N-dimensional interpolation objects with $(nindexes(N))"))
+    n == N || return :(error("Must index ", $N, "-dimensional interpolation objects with ", $(nindexes(N))))
     getindex_impl(N, IT, EB)
 end
 
@@ -193,7 +193,7 @@ end
 # Allow multilinear indexing with any types
 stagedfunction getindex{T,N,TCoefs,TIndex,IT,EB}(itp::Interpolation{T,N,TCoefs,IT,EB}, x::TIndex...)
     n = length(x)
-    n == N || return :(error("Must index $N-dimensional interpolation objects with $(nindexes(N))"))
+    n == N || return :(error("Must index ", $N, "-dimensional interpolation objects with ", $(nindexes(N))))
     getindex_impl(N, IT, EB)
 end
 
@@ -201,11 +201,11 @@ end
 #   gradient!(g::Vector, itp::Interpolation, NTuple{N}...)
 stagedfunction gradient!{T,N,TCoefs,TOut,IT,EB}(g::Vector{TOut}, itp::Interpolation{T,N,TCoefs,IT,EB}, x...)
     n = length(x)
-    n == N || return :(error("Must index $N-dimensional interpolation objects with $(nindexes(N))"))
+    n == N || return :(error("Must index ", $N, "-dimensional interpolation objects with ", $(nindexes(N))))
     it = IT()
     eb = EB()
     gr = gridrepresentation(it)
-    q = quote
+    quote
         length(g) == $N || error("g must be an array with exactly N elements (length(g) == "*string(length(g))*", N == "*string($N)*")")
         @nexprs $N d->(x_d = x[d])
         $(extrap_transform_x(gr,eb,N))
@@ -218,11 +218,8 @@ stagedfunction gradient!{T,N,TCoefs,TOut,IT,EB}(g::Vector{TOut}, itp::Interpolat
 
             @inbounds g[dim] = $(index_gen(degree(it),N))
         end
-        @show g
         g
     end
-    # @show q
-    q
 end
 
 gradient{T,N}(itp::Interpolation{T,N}, x...) = gradient!(Array(T,N), itp, x...)
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -1,7 +1,7 @@
 module Interpolations
 
 using Base.Cartesian
-using Compat
+using WoodburyMatrices, Compat
 
 import Base:
     eltype,
@@ -10,7 +10,7 @@ import Base:
     ndims,
     size
 
-export 
+export
     Interpolation,
     Constant,
     Linear,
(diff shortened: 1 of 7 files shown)


tlycken:
Thanks a lot for starting this off!

The error is indeed bizarre. I have no idea what could cause this, especially since - as you already demonstrated - it doesn't even seem to be consistent between display and indexing into the gradient vectors. Could it be a problem with re-using `g` for some reason? What happens if you pass `Array(Float64,nx)` instead? (If that is a problem, we need to fix it anyway, but it's nice to have things to rule out...)

I have a few nitpicks and questions about the code too, but I'll submit them as line comments shortly.

src/Interpolations.jl
```diff
@@ -123,131 +123,126 @@ function copy_with_padding(A, it::InterpolationType)
...
-
-                # Calculate coefficient weights based on fx
-                $(coefficients(it, N))
-
-                # Generate the indexing expression
-                @inbounds ret = $(index_gen(degree(it), N))
-                ret
-            end
-        end
+supported(::Type{Constant{OnGrid}}) = true
```
tlycken: Is `supported` used anywhere by us, or is it just a type trait that can be used elsewhere?

src/Interpolations.jl
```diff
@@ -123,131 +123,126 @@ function copy_with_padding(A, it::InterpolationType)
...
+# Resolve ambiguity with colon indexing for 1D interpolations
+#   getindex{T}(A::AbstractArray{T,1}, C::Colon)
+function getindex{T,TCoefs,IT,EB}(itp::Interpolation{T,1,TCoefs,IT,EB}, c::Colon)
+    error("Colon indexing is not supported for interpolation objects")
+end
+
+# Allow multilinear indexing with any types
+stagedfunction getindex{T,N,TCoefs,TIndex,IT,EB}(itp::Interpolation{T,N,TCoefs,IT,EB}, x::TIndex...)
+    n = length(x)
+    n == N || return :(error("Must index interpolation objects with as many indexes as dimensions"))
```
tlycken: It would be nice if these error messages also showed `dims(itp)` (i.e. `N`) and `n`, to make it easier to spot bugs (in the client code) where the user has several interpolation objects of different dimensions, and indexes into the wrong one because of a typo.

src/Interpolations.jl
```diff
@@ -123,131 +123,126 @@ function copy_with_padding(A, it::InterpolationType)
...
+    quote
         ret, pad = copy_with_padding(A, it)
 
         szs = collect(size(ret))
         strds = collect(strides(ret))
 
-        for dim in 1:N
-            n = szs[dim]
+        for dim in 1:$N
+             n = szs[dim]
```
tlycken: Whitespace error (one too many).

test/quadratic.jl
```diff
@@ -49,7 +49,7 @@ xlo, xhi = itp3[.2], itp3[xmax+.7]
 @test_approx_eq xhi A[end]
 
 # Check continuity
-xs = [0:.1:length(A)+1]
+xs = [0:.1:length(A)+1;]
```
tlycken: Is the `;` an idiom for something here (like getting a row vector instead of a col vector)? Why do we need that? As far as I can tell we're only using `xs` one value at a time, so dimensionality shouldn't matter as long as linear indexing works.

src/Interpolations.jl
timholy: No uses yet. I was keying off the `# This creates getindex methods for all supported combinations` and wondered if you wanted to be able to return an error like `"not a supported combination"` from something. So since the list was handy, I turned it into a trait. If we don't end up using it at all, then these should be deleted.

src/Interpolations.jl
timholy: Good idea.

test/quadratic.jl
timholy: https://github.com/JuliaLang/julia/pull/8599

timholy:
Thanks for the speedy review!

There didn't seem to be a magic bullet for the weird error. I'll keep plugging away at it.

test/quadratic.jl
tlycken: Ah. I had completely missed out on that discussion :P

I must admit that I don't like the syntax very much, although I realize I'm likely to be overridden by consensus here (and it might grow on me with time). Would I be considered terribly conservative if I perferred `[i-1 for i in 1:.1:length(x)]` (or `[(0:.1:length(x)-1)...]`) over `[0:.1:length(x)-1;]`? I just think the semicolon is a little too magical...

timholy: `vcat(rng)` or `convert(Vector{Float64}, rng)` are also choices; which do you prefer?

tlycken: We should actually be able to do without either; we could just as well do

```
for x in 0:.1:xmax
    @test_approx_eq_eps itp3[x] itp3[x+.1] .1
end
```

which I think is much less magical. (But I'm being *really* nitpicky here, so I don't blame you if you ignore me...)

timholy:
Got it. It was a bizarre consequence of mixing string interpolation and expression interpolation. For the purposes of base julia, I'll see if I can come up with a reduced test case, but meanwhile here's a version that passes all tests for me locally.

Package loading time is now 0.7 seconds on my laptop. On a different machine still running an older version of 0.4 (from before the great GC merge), package load time was 38.5 seconds. A nice improvement, I think!

timholy:
Before merging I'm happy to squash. I just left them as separate commits so you could see what changed.

timholy:
Now we are green. Turns out I needed to register the WoodburyMatrices package.

tlycken:
Nice find! I've never been nitpicky with git history, so feel free to squash or not whichever you prefer. When I start my Ubuntu machine next I'll tag a version of Interpolations.jl that has the latest changes in a 0.3-compatible state, and then I'll merge this if you're happy with it.

Again, thanks a lot for tackling this rewrite for me! This really does bring the package a huge step toward 1.0.

timholy:
Merged locally (with a rebase) in 6da1207df40329150706d354ec72346f02eda4c4. This way links will persist in case I want to refer to this conversation for seeing if this weird stagedfunction bug is a genuine julia bug.

tlycken:
@timholy I was going to use these improvements to take a new stab at #24, but I can't figure out what to check out to get these changes locally. Did you commit stuff to master locally without pushing?

tlycken:
nvm, figured it out. `git fetch` must have failed for some reason, because the local branch heads weren't updated.

Issue #30, opened 2015-03-29
Developer guide


spencerlyon2:
I've been following developments here and Grid.jl for quite some time now (grid for a couple years, this since its inception). 

I am going to be working on writing some routines for Chebyshev interpolation and cubic B-splines and would like to contribute them to this package as I feel this is where that code should ultimately belong. The only problem is that I don't have enough time to really dig in and learn all the meta-magic works in this package.

Is there a brief developer guide that explains how the system works and how it can be extended? If there isn't one, might I submit a request that one be written? Thanks!

timholy:
That sounds great. There's been a longstanding request to document stagedfunction, and while no one has jumped at it yet clearly that's something important to do. Documentation specific to this package would also be handy, I agree. @tlycken has also published a number of notebooks in the course of developing this, perhaps some links to those would also help.

For myself, the problem is I'm pretty swamped with my own work and https://github.com/JuliaLang/julia/pull/10525#issuecomment-85783671 (which is taking me into parts of julia I've never explored before, so is really slow). I'll see if I can do something, but it could be a few weeks.

tlycken:
@spencerlyon2 That sounds like great contributions to the package!

I have some very large API changes in the pipeline, in order to make room in the API for other types of interpolation than the B-splines we're currently using - that would make it easier to get somewhere with #28, as well as to include your Chebyshev interpolation routines. These changes will probably have quite a large effect on how best to contribute to B-spline functionality, but it should be quite simple to get started on the Chebyshev parts; I already know where I want to go, so you could probably do some work in parallel and then we'll sync up when we're both ready.

Basically, create code for the following functionality:

1. A method on the `interpolate` function that creates an interpolation object

    ```julia
# type_spec is an argument that specifies *how* we should interpolate
itp = interpolate(A, type_spec)
```

For Chebyshev interpolation, you probably want to create one type `Chebyshev`, which is configures the interpolation degree etc, and one type `ChebyshevInterpolation` which is the type of the resulting interpolation objects. You might end up with something like this:

    ```julia
itp = interpolate(A, Chebyshev)
# if you need some parameters, e.g. a boundary condition:
itp = interpolate(A, Chebyshev(Flat))
```

where `itp` is now of type `ChebyshevInterpolation` (maybe with some type parameters).

2. Define indexing on that interpolation object.

    ```julia
getindex(itp::ChebyshevInterpolation, idxs...) = ...
```

[...]

Issue #28, opened 2015-03-02
Incorporation of functionality from GridInterpolations.jl


mykelk:
[GridInterpolations.jl](https://github.com/sisl/GridInterpolations.jl) provides multivariate interpolation on a rectilinear grid with support for both multilinear and simplex interpolation. My lab does interpolation in an inner-inner loop for a dynamic programming application, and so quite a bit of effort was put into making it fast. It would be neat if this functionality can be migrated to Interpolations.jl without compromising on speed or the simplicity of the interface.

tlycken:
Hi @mykelk! I'm sorry I haven't replied to this. I'm sure I at least started writing a reply to this somewhere, but now I can't find it and it obviously didn't get posted.

I tried to skim the paper you referred to from the readme, as well as one of its references, to better understand the math behind your implementation technique, but it was a little too opaque for me to get it without spending more time on it. I hope to be able to do so over the weekend, but if you have easy-to-digest resources that would make this a little easier they would be most welcome.
 
I've also taken a quick look at the code, and while I do appreciate the effort to make everything really fast, I feel that a little too much of the generic nature of Julia's type system has been sacrificed in the process. I think we could preserve most of, if not all, the speed while making the code much more general, with quite a small effort - it would mostly be stuff like removing type arguments from function definitions to allow other types of data than `Float64`.

I would also like to harmonize the API in your code with the API in Interpolations.jl before we merge them into a single package. This might require some changes to Interpolations.jl to make room for your functionality, but that's completely OK with me. Basically, I think the following is probably a good pattern to follow:

```julia
itp = interpolate(data, interpolation_type_definition)
```

where `data` is the array of y/z coordinates in 1/2D (and corresponding quantity for higher dimensions) and `interpolation_type_definition` is a type argument that defines the type of interpolation that should be performed (for the existing linear B-spline interpolation it could be e.g. `BSpline(Linear(OnGrid()))`). We can then wrap this in another method/type to scale the coordinate axes, and maybe add a separate argument for extrapolation.

What do you think?

mykelk:
[...]

Issue #6, opened 2014-11-21
Hermite splines?


tomasaschan:
A [recent post on julia-users](https://groups.google.com/forum/#!topic/julia-users/-bbJ_rVSOjs) made a case for [Cubic Hermite splines](http://en.wikipedia.org/wiki/Cubic_Hermite_spline), that seem to have some properties that regular B-splines might not have (although I must admit that I'm not too familiar with the terminology used in the julia-users thread...).

Cubic Hermite splines are quite similar to B-splines, but instead of using several data points to construct an interpolating polynomial, the derivatives at the data points are used instead. It readily generalizes to irregular grids, which is nice, but I can't find any good resources on whether it generalizes nicely also in dimensions.

lstagner:
I recently had to code up Cubic Hermite Splines (including Monotone Cubic Splines). I also implemented Polyharmonic Splines which work on N-dimensional irregularly gridded data.
 
Heres the code
[Cubic Splines](https://gist.github.com/lstagner/83f91cc138d4767523a3)
[Polyharmonic Splines](https://gist.github.com/lstagner/04a05b120e0be7de9915#file-polyharmonic_splines-jl)

I probably needs a bit of cleanup but its a start.

cauribeteran:
@lstagner, your code looks pretty good. Have you compared your functions in terms of efficiency with pchip or SimplePchip (https://github.com/slabanja/SimplePCHIP)? I'm solving a big scale model and I'm having a hard time in terms of efficiency. Using linear interpolation (Interpolations.jl) each iteration takes around 0.86 minutes. But that approximation is not good enough, and Pchip is required. My matlab code (enhanced with Matlab Coder running C++ code) uses the interp1(...,'pchip') function and each iteration takes about 1.3 minutes. However, using SimplePCHIP in julia each iteration takes almost 13 minutes!! Any recommendations?

tlycken:
@cauribeteran Have you tried higher order B-splines from Interpolations.jl? Are there any reasons you can't apply them to your use case?

The B-spline parts of Interpolations.jl have been very carefully crafted to be performant (most of the actual work in that area was done by Tim). For quadratic and higher order B-splines there's an initial "setup" cost, for solving a system of equations for the spline coefficients, but after that it should be on the same scale as linear interpolation. If you do a lot of evaluations in each iteration, before you change the underlying data (and with 0.83 minutes per iteration for linear splines it sounds like you do) the "setup" cost might be negligible.

[...]

PR #31, opened 2015-04-10
Api rewrite

tlycken:
This is the start of a complete rewrite of this library to open up the API for other types of interpolations than B-splines. So far only constant interpolation (and no extrapolation) is implemented, but I think this shows clear enough where I want this to go, so I figure this is as good a time as any to ask for feedback.

The API will mainly affect how interpolation objects are constructed. What was previously

```julia
itp = Interpolation(A, Constant(OnCell), ExtrapError)
```
will now become

```julia
itp = interpolate(A, BSpline(Constant), OnCell)
exp = extrapolate(itp, Error) # but with a name that doesn't collide - suggestions?
```

I have a couple of reasons I want to make this change:

* We don't hog the name `Interpolation` for B-splines specifically - now, any type of interpolation can add methods to the `interpolate` function, with their own configuration type. Thus, I hope it will be easier to extend the package with things like simplex (see #28), Chebyshev (see #30) or Hermite spline (see #6) interpolation.

* Extrapolation is now even more separated from interpolation, which means that when e.g. linear extrapolation is implemented once, it will be re-usable across all interpolation types, and not only for B-splines. This can hopefully speed things up a little as well; if you *know* you're not going to extrapolate, you never have to bounds-check when indexing into an interpolation object.

* Other things, like scaling uniform grids from the unit grid `1:n` to other coordinate ranges, i.e. what `CoordInterpGrid` does in `Grid.jl`, also become re-usable across interpolation types, with a similar API.

Even if you don't have anything else to say about it than "OK with me", I'd love to hear some feedback on this.

The timeline for this rewrite is a little unclear - I have a lot on my table over the coming month, so progress will be slow. However, most of it is just shuffling code around, so it's perfectly possible to re-use other improvements made to the package in the meantime.


tlycken:
(No, I haven't updated the tests to use this API yet, so they'll all fail miserably... I run some sanity tests with plots in the meantime, and they seem to give plausible results.)

timholy:
Sorry, I completely missed this. (April was really busy.) :+1: for the idea. I'll try to find some time to help with this.

tlycken:
:smiley: 

I have gotten quite far on porting quadratic interpolation as well, but haven't gotten around to pushing the changes yet; testing is the area where I've made absolutely the least progress. So if you do find some time, porting the tests is probably where the risk of duplicated work is the least .)

As you have exposed some areas with room for improvement, I'll try to get to this asap, so we can build from a common platform.

timholy:
OK, I'll wait to do anything until you let me know you've updated this branch.

tlycken:
I've now rebased this onto master. There were a few conflicts, in which I doubt I managed to salvage all the recent changes to master, so I'd be glad if you could take this for a spin.

The old version of this branch is available as `api-rewrite-backup` if we'd want to go back to it.

tlycken:
@timholy, Now that the tests pass, I think this is a solid platform to build upon.

I don't know how widely this package is used, but I could tag a version on master (to get your compatibility fixes there) and then merge this and tag again (minor bump for pre-release breaking changes?) to make collaboration easier. What do you think?

timholy:
This looks awesome. Sure, go ahead and tag master. Meanwhile I'll check out this branch and then start adding to it.

timholy:
Bump. Anything I can do to help merge this? If it's just "tag current master, then merge and re-tag" I can do that for you.

tlycken:
Sorry - I really had hoped I would have gotten to this by now. There are (always) more stuff I'd like to get done here, but for now I think this will have to do so I stop blocking you :)

Yes, a "tag current, then merge, then tag again" is what I had in mind. Given that the API is nowhere near backwards-compatible, I think 0.0.2 - merge - 0.1.0 is reasonable. Fire at will!

tlycken:
Thanks a lot for helping me land this, @timholy. As always, you're invaluable :)

PR #32, opened 2015-05-16
Fixes for 0.4 deprecations

timholy:
This updates for recent julia changes. It works until it doesn't :smile:, but that's because of https://github.com/JuliaLang/julia/issues/11298.


tlycken:
Nice :) I need to fix this in #31 too, but since it's going to overwrite much of the work in your PR's (just because it organizes the code differently) I think we'd better coordinate our efforts.

Since the Julia PR is merged I restarted the Travis build, so we should get a green light on this. Feel free to merge it if you need it in master, but feel even more free to help pushing #31 to completion :smirk:

timholy:
I'm going to merge it to master, but then we can make the same changes in #31.

PR #33, opened 2015-05-16
Make prefiltering faster

timholy:
This PR (which includes #32) makes prefiltering somewhat faster. In many cases the improvement is a bit disappointing, but when Woodbury can be avoided it's quite dramatic.

With current master (well, really with #32, so the deprecations don't slow things down):
```jl
julia> using Interpolations

julia> A = rand(300,301,30);

julia> it = Quadratic(Line(),OnGrid())  # requires Woodbury
Interpolations.Quadratic{Interpolations.Line,Interpolations.OnGrid}()

julia> itp = Interpolation(A, it, ExtrapError());

julia> @time Interpolation(A, it, ExtrapError());
elapsed time: 1.05831119 seconds (637 MB allocated, 8.97% gc time in 29 pauses with 0 full sweep)

julia> it = Quadratic(Flat(),OnCell())   # doesn't require Woodbury
Interpolations.Quadratic{Interpolations.Flat,Interpolations.OnCell}()

julia> itp = Interpolation(A, it, ExtrapError());

julia> @time Interpolation(A, it, ExtrapError());
elapsed time: 0.571626723 seconds (373 MB allocated, 10.14% gc time in 17 pauses with 0 full sweep)
```
With this PR:
```jl
julia> using Interpolations

julia> A = rand(300,301,30);

julia> it = Quadratic(Line(),OnGrid())  # requires Woodbury
Interpolations.Quadratic{Interpolations.Line,Interpolations.OnGrid}()

julia> itp = Interpolation(A, it, ExtrapError());

julia> @time Interpolation(A, it, ExtrapError());
elapsed time: 0.600699271 seconds (44 MB allocated, 15.04% gc time in 2 pauses with 2 full sweep)

julia> it = Quadratic(Flat(),OnCell())   # doesn't require Woodbury
Interpolations.Quadratic{Interpolations.Flat,Interpolations.OnCell}()

julia> itp = Interpolation(A, it, ExtrapError());

julia> @time Interpolation(A, it, ExtrapError());
elapsed time: 0.08264564 seconds (22 MB allocated, 9.55% gc time in 1 pauses with 0 full sweep)
```
It also allocates a lot less memory. With the new GC this isn't a huge issue, but neither is it a bad thing. It might lead to even greater benefits for smaller arrays.


timholy:
I should add that this makes the prefiltering in Interpolations faster than Grid, too. Grid was a bit faster than Interplations master, but not by much; in a 2d case, I got 0.19 s for master Interpolations, .17 s for Grid, and 0.12 for this PR.

timholy:
Note mostly to self: Grid seems to be able to do without Woodbury in a few more cases than Interpolations. Try to figure out why (important since it has a big performance impact).

timholy:
OK, this has been rebased on #31. I'll probably wait until that's merged to work on #34.

timholy:
As an update, once https://github.com/JuliaLang/METADATA.jl/pull/2670 merges, I will be able to contribute a further improvement in the Woodbury prefiltering.

PR #36, opened 2015-05-29
In-place interpolations revisited

timholy:
By comparison with #34, this is probably a better approach: rather than defining a new boundary condition `Interior` and then making the indexing rules treat padding differently for this condition, this PR adds a `pad` parameter to `BSplineInterpolation`: for instance, `pad=0` for anything other than `Quadratic`, for which it is typically 1 unless (1) using `Reflect` or (2) using `interpolate!`. The code generation uses this parameter to compute appropriate index offsets.

The first commit in this PR also improves the type-stability (and adds tests for it) of several key functions. Also, I think the `size` function was broken (it included the consequences of padding, which I assume is _not_ what you wanted?); this is fixed in the 2nd commit by exploiting the new `pad` parameter.

### DIFF
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -2,6 +2,7 @@ module Interpolations
 
 export
     interpolate,
+    interpolate!,
     extrapolate,
 
     OnCell,
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -38,7 +38,8 @@ immutable Periodic <: BoundaryCondition end
 immutable Reflect <: BoundaryCondition end
 typealias Natural Line
 
-size(itp::AbstractInterpolation) = tuple([size(itp,i) for i in 1:ndims(itp)]...)
+# TODO: size might have to be faster?
+size{T,N}(itp::AbstractInterpolation{T,N}) = ntuple(i->size(itp,i), N)::NTuple{N,Int}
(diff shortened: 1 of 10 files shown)


timholy:
With the improvements in filtering, this is nice to see (suitable warmup has been done already):
```jl
julia> A = rand(300,300,30);

julia> Ai = copy(A);

julia> @time itp = interpolate!(Ai, BSpline(Quadratic(Flat)), OnCell);
  54.937 milliseconds (198 allocations: 49757 bytes)

julia> Ag = copy(A);

julia> @time yi = InterpGrid(Ag, BCreflect, InterpQuadratic);
 299.404 milliseconds (108 k allocations: 26187 KB, 20.47% gc time)
```
But this was not so nice:
```jl
julia> function interp_all!(dest, itp)
           for k = 1:size(itp,3)
               for j = 1:size(itp,2)
                   for i = 1:size(itp,1)
                       dest[i,j,k] = itp[i,j,k]
                   end
               end
           end
           dest
       end
interp_all! (generic function with 1 method)

julia> B = similar(A);

julia> @time interp_all!(B, itp);
  10.251 seconds      (2700 k allocations: 42188 KB)

julia> @time interp_all!(B, yi);
   1.069 seconds      (4 allocations: 144 bytes)
```
I'll see if I can fix that.

PR #40, opened 2015-05-29
Start writing developer documentation

### LINKED ISSUE #30: Developer guide

timholy:
This adds some documentation that may be useful for people who want to hack on Interpolations. Should fix #30. CC @spencerlyon2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,6 @@ There's an [IJulia notebook](http://nbviewer.ipython.org/github/tlycken/Interpol
 
 ## Contributing
 
-Work is very much in progress, but and help is always welcome. If you want to help out but don't know where to start, take a look at issue [#5 - our feature wishlist](https://github.com/tlycken/Interpolations.jl/issues/5) =)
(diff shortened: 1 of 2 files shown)


timholy:
(You can see what it looks like by clicking the "View" button in the diff page.)

spencerlyon2:
Excellent. Thanks for doing this and CCing me.

The code that I wrote sits (for now) in spencerlyon2/CompEcon.jl

Currently that repo hosts  a Julia port of the function approximation bits of a well used matlab toolbox for computational economics. When I find time, I would like to comb through that code again, make it more Julian, and try to implement some obvious optimizations. 

The code implements Chebhyshev interpolation as well as arbitrary order b-splines for arbitrary dimensions using a tensor product rule to combine univariate bases. 

Feel free to look around and let me know if you think any of it belongs here.

PR #42, opened 2015-06-01
Faster prefiltering for the Woodbury case

timholy:
This gives an approximate further 2x speed improvement for prefiltering with Woodbury matrices:
```jl
Master:
julia> using Interpolations

julia> A = rand(300,300,30);

julia> Ai = copy(A);

# Delete warmup

julia> @time itp = interpolate!(Ai, BSpline(Quadratic(Periodic)), OnCell);
 465.640 milliseconds (1728 k allocations: 48201 KB, 2.23% gc time)
```
```jl
This branch:
julia> @time itp = interpolate!(Ai, BSpline(Quadratic(Periodic)), OnCell);
 234.042 milliseconds (401 allocations: 127 MB, 5.24% gc time)
```

This uses the new AxisAlgorithms package, and converts the U, V matrices to sparse format (which required updates to WoodburyMatrices.jl).

For comparison:
```jl
julia> @time yi = InterpGrid(Ai, BCperiodic, InterpQuadratic);
 648.083 milliseconds (2700 k allocations: 66754 KB, 10.09% gc time)
```
using Grid.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,5 @@
 julia 0.4-
 
-WoodburyMatrices
+WoodburyMatrices 0.1
 Ratios
+AxisAlgorithms
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -19,7 +19,7 @@ export
     # b-splines/b-splines.jl
     # extrapolation/extrapolation.jl
 
-using WoodburyMatrices, Ratios
+using WoodburyMatrices, Ratios, AxisAlgorithms
 
 import Base: convert, size, getindex, promote_rule
 
--- a/src/b-splines/prefiltering.jl
+++ b/src/b-splines/prefiltering.jl
@@ -34,21 +34,7 @@ function prefilter!{TWeights,TCoefs,N,IT<:Quadratic,GT<:GridType}(
     first = true
     for dim in 1:N
         M, b = prefiltering_system(TWeights, TCoefs, sz[dim], IT, GT)
-        if !isa(M, Woodbury)
-            A_ldiv_B_md!(ret, M, ret, dim, b)
-        else
-            if first
-                buf = Array(eltype(ret), length(ret,))
-                shape = sz
-                retrs = reshape(ret, shape)  # for type-stability against a future julia #10507
-                first = false
-            end
-            bufrs = reshape(buf, shape)
-            filter_dim1!(bufrs, M, retrs, b)
-            shape = (sz[dim+1:end]..., sz[1:dim]...)::NTuple{N,Int}
-            retrs = reshape(ret, shape)
-            permutedims!(retrs, bufrs, ((2:N)..., 1))
-        end
+        A_ldiv_B_md!(ret, M, ret, dim, b)
     end
     ret
 end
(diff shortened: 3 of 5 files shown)


tlycken:
Nice!

Although the execution time is halved, the allocation seems to have gone up by more than a factor 2. Any idea why? Allocating almost 130 MB on a 300x300 matrix seems like it could scale pretty badly when the matrix becomes bigger (what happens for 100x100x100?)...

timholy:
Note it was a 300x300x30 array, which is bigger than 100x100x100.

In the current (master) algorithm, I call `permutedims!` to shift the "working" dimension into the first dimension. I use a [single buffer, reshaping it each time](https://github.com/tlycken/Interpolations.jl/blob/2dc6cd5291e51704e19614fc328d33e3ec14801e/src/b-splines/prefiltering.jl#L40-L50), to make that less memory-consuming. Then I can iterate column-by-column, which doesn't consume memory.

In contrast, this new algorithm doesn't use `permutedims!`. However, the matrix multiplications in the Woodbury formula require temporary allocation, https://github.com/tlycken/Interpolations.jl/blob/5e67f1c81b6acc494167c4dad504fe39a0fd3db4/src/filter1d.jl#L77-L81. Some size comparisons are probably helpful:
`size(src) == size(dest) == size(tmp4) == size(tmp3) >> size(tmp2) == size(tmp1)`. (The "little" ones are of size 2 or 4 along the "working" dimension, stemming from the fact that `V` has 2 or 4 columns.) I could, at the cost of essentially duplicating https://github.com/tlycken/Interpolations.jl/blob/5e67f1c81b6acc494167c4dad504fe39a0fd3db4/src/filter1d.jl#L18-L71 yet again (it's first in `AxisAlgorithms`, but I needed a variant that supported `b`), skip the creation of `tmp4` and fuse that with the `sub!` step. Not only would this allocate one less temporary array, it would also give yet-faster performance, because it's one less "trip" (with attendant cache misses) through an array of size `src`. If you're OK with that much near-duplication, I can go ahead and do that.

timholy:
If memory is a big concern, we could probably also pre-allocate `tmp3` and reuse it for each dimension. It would be a bit uglier (the tridiagonal-only case would have an unused argument), but nothing too horrible.

timholy:
But of course all these temporaries can be gced after each dimension.

tlycken:
As usual, you're on top of things :)

I was only a little worried that there was something fishy going on, but you've convinced me that given the uglification needed to get rid of these extra allocations I think we should wait until we see that it's actually necessary. After all, on most machines 130 MB RAM, temporarily used, isn't so bad, and we can probably expect machine size to scale with data set size too.

timholy:
I did just realize that we didn't need to allocate a fresh array `tmp4`; tridiagonal inversion can be done in-place. Just pushed a new commit that essentially halves the amount of memory allocated:
```jl
julia> @time itp = interpolate!(Ai, BSpline(Quadratic(Periodic)), OnCell);
 228.467 milliseconds (392 allocations: 66807 KB, 4.41% gc time)
```

tlycken:
I wish I could be half as productive as you...

timholy:
Actually, I really respect your design sense---it sets a standard I aspire to. We make a good team :smile:.

Issue #41, opened 2015-06-01
Anisotropic interpolation (i.e. different specs in different dimensions)


timholy:
I've encountered a situation where I want to interpolate along some dimensions but not others. I now wonder whether we should make the `IT` parameter of `BSplineInterpolation` a tuple, one per dimension of the array. One could use, for example, quadratic interpolation for dimensions 1-2 and linear interpolation along dimension 3, or skip interpolation along some dimensions. If this parameter is a tuple-parameter, then the `@generated` functions can generate the appropriate `getindex` code for the type.

Thoughts? We'd of course want there to be an "easy" interface that replicates current behavior, as well as one allowing more control.

tlycken:
I've thought about this too - although I mostly thought about it in the context of boundary conditions and extrapolation. But yes, I think it would be a really nice and powerful feature to be able to specify different behavior in different dimensions.

I imagine a good solution would be to allow *either* arguments as we state them now, *or* tuples of them, and expand the single-argument calls to tuple calls before doing any actual handling of stuff. In other words,

```julia
A = rand(100, 100)
interpolate(A, Quadratic(Flat), Constant)
```

would be equivalent to

```julia
A = rand(100, 00)
interpolate(A, (Quadratic(Flat), Quadratic(Flat)), (Constant, Constant))
```

allowing for variations like

```
A = rand(100, 100)
interpolate(A; (Quadratic(Flat), None), (Constant, ErrorExtrap)```
interpolate(A, Quadratic((Flat,Line)), (Constant, Linear))
# etc...
```

Unfortunately, I won't have time to work on this in the foreseeable future, but I definitely like the idea so if you have a need for it and want to take a stab I'll gladly help out when I do find some time :)

timholy:
Closed by #43

PR #43, opened 2015-07-03
Allow different types of interplation along different dimensions

timholy:
This implements #41

### DIFF
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -28,7 +28,9 @@ abstract GridType
 immutable OnGrid <: GridType end
 immutable OnCell <: GridType end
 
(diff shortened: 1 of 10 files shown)


tlycken:
Amazing work, as usual (both this, and your other PRs in this batch).

I'd like to take them for a spin before I merge them - and I also have a hunch about a few refactorings that could be possible, and might simplify the code a little. I'll have to test them out myself, though, before I know if it works, so I'll get back to you with specifics :)

But seriously, it's kind of amazing that Julias type system is expressive enough for these changes to be so relatively minor, despite the fact that the functionality they provide is awesome.

timholy:
I will be quite interested to see what you come up with. Thanks for taking the time to review.

PR #44, opened 2015-07-03
Re-implement gradient computation

timholy:
This also tests the gradient using DualNumbers, which serves both as a check about our type-generality and provides a stringent test of the gradient computation.

Testing could be more thorough, but this is a start.

EDIT: built on top of #43, the new stuff starts with the 2nd commit.


timholy:
Note to self: whenever I think to myself "testing could be more thorough," bet $1M that I have a bug in my code.

PR #47, opened 2015-07-20
RFC: Scaling of interpolation objects (fixes #25)

### LINKED ISSUE #25: Something equivalent to CoordInterpGrid
In order to reach feature parity with Grid.jl, we need something like `CoordInterpGrid`, that re-scales the interpolation indices to some other coordinate system. This could probably be implemented very similarly to `CoordInterpGrid`, since the underlying interpolation interface is very similar.

tlycken:
This is a start at implementing the functionality provided to Grid.jl by the CoordInterpGrid type, and in fact the code here is heavily based on that contributed by @simonbyrne there.

Simon, do you have any objections to me basing this implementation off of your code? If so, please feel free to object here, and I'll remove this and come up with something from scratch. It's just that borrowing is so easy ;)

The basic idea is that you can do

```julia
f(x) = exp(-x.^2 ./ 4)
xs = -5:.2:5
ys = f(xs)
itp = interpolate(ys, BSpline(Linear), OnGrid)

s = scale(itp, xs)   # now, s[-2.3] ~= f(-2.3)
```

To scale in higher dimensions, you just tuck on more ranges at the end, one range for each dimension (i.e. `scale(itp2, xs, ys)` to scale a 2D interpolation object).

This doesn't fully handle everything we want yet - for one thing, there's no scaling of the coordinates that works for gradients. Also, in Grid.jl @timholy added specializations for dimensions 1-4 that avoided splatting. I'd like to see if it's possible to do this using staged functions instead of explicitly writing them out, but this might not be fully performant yet. It also needs documentation.

However, if anyone has any opinions about the API, this is as good a time as any to raise them.

Todo:

* [x] basic implementation
* [x] gradient evaluation
* [x] make sure it works with `DimSpec`s and `NoInterp` dimensions
* [x] tests for combinations with `extrapolate` (both `extrapolate(scale(...))` and `scale(extrapolate(...))`)
* [x] make sure scaled interpolation objects display nicely
* [x] documentation

### DIFF
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -4,6 +4,7 @@ export
     interpolate,
     interpolate!,
     extrapolate,
+    scale,
 
     gradient!,
 
@@ -22,10 +23,11 @@ export
     # see the following files for further exports:
     # b-splines/b-splines.jl
     # extrapolation/extrapolation.jl
+    # scaling/scaling.jl
 
 using WoodburyMatrices, Ratios, AxisAlgorithms
 
-import Base: convert, size, getindex, gradient, promote_rule
+import Base: convert, size, getindex, gradient, scale, promote_rule
 
 abstract InterpolationType
 immutable NoInterp <: InterpolationType end
@@ -36,7 +38,8 @@ immutable OnCell <: GridType end
 typealias DimSpec{T} Union{T,Tuple{Vararg{Union{T,NoInterp}}},NoInterp}
 
 abstract AbstractInterpolation{T,N,IT<:DimSpec{InterpolationType},GT<:DimSpec{GridType}} <: AbstractArray{T,N}
-abstract AbstractExtrapolation{T,N,ITPT,IT,GT} <: AbstractInterpolation{T,N,IT,GT}
+abstract AbstractInterpolationWrapper{T,N,ITPT,IT,GT} <: AbstractInterpolation{T,N,IT,GT}
+abstract AbstractExtrapolation{T,N,ITPT,IT,GT} <: AbstractInterpolationWrapper{T,N,ITPT,IT,GT}
 
 abstract BoundaryCondition
 immutable Flat <: BoundaryCondition end
@@ -53,7 +56,13 @@ typealias Natural Line
 # TODO: size might have to be faster?
 size{T,N}(itp::AbstractInterpolation{T,N}) = ntuple(i->size(itp,i), N)::NTuple{N,Int}
 size(exp::AbstractExtrapolation, d) = size(exp.itp, d)
- itptype{T,N,IT,GT}(itp::AbstractInterpolation{T,N,IT,GT}) = IT
+bounds{T,N}(itp::AbstractInterpolation{T,N}) = tuple(zip(lbounds(itp), ubounds(itp))...)
+bounds{T,N}(itp::AbstractInterpolation{T,N}, d) = (lbound(itp,d),ubound(itp,d))
+lbounds{T,N}(itp::AbstractInterpolation{T,N}) = ntuple(i->lbound(itp,i), N)::NTuple{N,T}
+ubounds{T,N}(itp::AbstractInterpolation{T,N}) = ntuple(i->ubound(itp,i), N)::NTuple{N,T}
+lbound{T,N}(itp::AbstractInterpolation{T,N}, d) = convert(T, 1)
+ubound{T,N}(itp::AbstractInterpolation{T,N}, d) = convert(T, size(itp, d))
+itptype{T,N,IT,GT}(itp::AbstractInterpolation{T,N,IT,GT}) = IT
 gridtype{T,N,IT,GT}(itp::AbstractInterpolation{T,N,IT,GT}) = GT
 
 @inline gradient{T,N}(itp::AbstractInterpolation{T,N}, xs...) = gradient!(Array(T,N), itp, xs...)
@@ -62,5 +71,6 @@ include("nointerp/nointerp.jl")
 include("b-splines/b-splines.jl")
 include("gridded/gridded.jl")
 include("extrapolation/extrapolation.jl")
+include("scaling/scaling.jl")
 
 end # module
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -8,6 +8,9 @@ export
 
     gradient!,
 
+    AbstractInterpolation,
+    AbstractExtrapolation,
+
     OnCell,
     OnGrid,
 
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -33,7 +33,7 @@ abstract GridType
 immutable OnGrid <: GridType end
 immutable OnCell <: GridType end
 
-typealias DimSpec{T} Union(T,Tuple{Vararg{Union(T,NoInterp)}},NoInterp)
+typealias DimSpec{T} Union{T,Tuple{Vararg{Union{T,NoInterp}}},NoInterp}
 
 abstract AbstractInterpolation{T,N,IT<:DimSpec{InterpolationType},GT<:DimSpec{GridType}} <: AbstractArray{T,N}
 abstract AbstractExtrapolation{T,N,ITPT,IT,GT} <: AbstractInterpolation{T,N,IT,GT}
--- a/src/scaling/scaling.jl
+++ b/src/scaling/scaling.jl
@@ -1,0 +1,69 @@
+type ScaledInterpolation{T,N,ITPT,IT,GT,RT} <: AbstractInterpolationWrapper{T,N,ITPT,IT,GT}
+    itp::ITPT
+    ranges::RT
+end
+ScaledInterpolation{T,ITPT,IT,GT,RT}(::Type{T}, N, itp::ITPT, ::Type{IT}, ::Type{GT}, ranges::RT) =
+    ScaledInterpolation{T,N,ITPT,IT,GT,RT}(itp, ranges)
+function scale{T,N,IT,GT}(itp::AbstractInterpolation{T,N,IT,GT}, ranges::Range...)
+    length(ranges) == N || throw(ArgumentError("Must scale $N-dimensional interpolation object with exactly $N ranges (you used $(length(ranges)))"))
+    for d in 1:N
+        if iextract(IT,d) != NoInterp 
+            length(ranges[d]) == size(itp,d) || throw(ArgumentError("The length of the range in dimension $d ($(length(ranges[d]))) did not equal the size of the interpolation object in that direction ($(size(itp,d)))"))
+        elseif ranges[d] != 1:size(itp,d)
+            throw(ArgumentError("NoInterp dimension $d must be scaled with unit range 1:$(size(itp,d))"))
+        end
+    end
+
+    ScaledInterpolation(T,N,itp,IT,GT,ranges)
+end
+
+@generated function getindex{T,N,ITPT,IT<:DimSpec}(sitp::ScaledInterpolation{T,N,ITPT,IT}, xs...)
+    length(xs) == N || throw(ArgumentError("Must index into $N-dimensional scaled interpolation object with exactly $N indices (you used $(length(xs)))"))
+    interp_types = length(IT.parameters) == N ? IT.parameters : tuple([IT.parameters[1] for _ in 1:N]...)
+    interp_dimens = map(it -> interp_types[it] != NoInterp, 1:N)
+    interp_indices = map(i -> interp_dimens[i] ? :(coordlookup(sitp.ranges[$i], xs[$i])) : :(xs[$i]), 1:N)
+    return :(getindex(sitp.itp, $(interp_indices...)))
+end
+
+getindex{T}(sitp::ScaledInterpolation{T,1}, x::Number, y::Int) = y == 1 ? sitp[x] : throw(BoundsError())
+
+size(sitp::ScaledInterpolation, d) = size(sitp.itp, d)
+lbound{T,N,ITPT,IT}(sitp::ScaledInterpolation{T,N,ITPT,IT,OnGrid}, d) = 1 <= d <= N ? sitp.ranges[d][1] : throw(BoundsError())
+lbound{T,N,ITPT,IT}(sitp::ScaledInterpolation{T,N,ITPT,IT,OnCell}, d) = 1 <= d <= N ? sitp.ranges[d][1] - boundstep(sitp.ranges[d]) : throw(BoundsError())
+ubound{T,N,ITPT,IT}(sitp::ScaledInterpolation{T,N,ITPT,IT,OnGrid}, d) = 1 <= d <= N ? sitp.ranges[d][end] : throw(BoundsError())
+ubound{T,N,ITPT,IT}(sitp::ScaledInterpolation{T,N,ITPT,IT,OnCell}, d) = 1 <= d <= N ? sitp.ranges[d][end] + boundstep(sitp.ranges[d]) : throw(BoundsError())
+
+boundstep(r::LinSpace) = ((r.stop - r.start) / r.divisor) / 2
+boundstep(r::FloatRange) = r.step / 2
+boundstep(r::StepRange) = r.step / 2
+boundstep(r::UnitRange) = 1//2
+
+coordlookup(r::LinSpace, x) = (r.divisor * x + r.stop - r.len * r.start) / (r.stop - r.start)
+coordlookup(r::FloatRange, x) = (r.divisor * x - r.start) / r.step + one(eltype(r))
+coordlookup(r::StepRange, x) = (x - r.start) / r.step + one(eltype(r))
+coordlookup(r::UnitRange, x) = x - r.start + one(eltype(r))
+coordlookup(i::Bool, r::Range, x) = i ? coordlookup(r, x) : convert(typeof(coordlookup(r,x)), x)
+
+gradient{T,N}(sitp::ScaledInterpolation{T,N}, xs...) = gradient!(Array(T,N), sitp, xs...)
+@generated function gradient!{T,N,ITPT,IT}(g, sitp::ScaledInterpolation{T,N,ITPT,IT}, xs...)
+    ndims(g) == 1 || throw(ArgumentError("g must be a vector (but had $(ndims(g)) dimensions)"))
+    length(xs) == count_interp_dims(IT, N) || throw(ArgumentError("Must index into $N-dimensional scaled interpolation object with exactly $N indices (you used $(length(xs)))"))
+
+    interp_types = length(IT.parameters) == N ? IT.parameters : tuple([IT.parameters[1] for _ in 1:N]...)
+    interp_dimens = map(it -> interp_types[it] != NoInterp, 1:N)
+    interp_indices = map(i -> interp_dimens[i] ? :(coordlookup(sitp.ranges[$i], xs[$i])) : :(xs[$i]), 1:N)
+
+    quote
+        length(g) == N || throw(ArgumentError(string("g must be a vector of length ", N, " (was ", length(g), ")")))
+        gradient!(g, sitp.itp, $(interp_indices...))
+        for i in eachindex(g)
+            g[i] = rescale_gradient(sitp.ranges[i], g[i])
+        end
+        g
+    end
+end
+
+rescale_gradient(r::LinSpace, g) = g * r.divisor / (r.stop - r.start)
+rescale_gradient(r::FloatRange, g) = g * r.divisor / r.step
+rescale_gradient(r::StepRange, g) = g / r.step
+rescale_gradient(r::UnitRange, g) = g
--- a/src/scaling/scaling.jl
+++ b/src/scaling/scaling.jl
@@ -4,6 +4,13 @@ type ScaledInterpolation{T,N,ITPT,IT,GT,RT} <: AbstractInterpolationWrapper{T,N,
 end
 ScaledInterpolation{T,ITPT,IT,GT,RT}(::Type{T}, N, itp::ITPT, ::Type{IT}, ::Type{GT}, ranges::RT) =
     ScaledInterpolation{T,N,ITPT,IT,GT,RT}(itp, ranges)
+@doc """
+`scale(itp, xs, ys, ...)` scales an existing interpolation object to allow for indexing using other coordinate axes than unit ranges, by wrapping the interpolation object and transforming the indices from the provided axes onto unit ranges upon indexing.
+
+The parameters `xs` etc must be either ranges or linspaces, and there must be one coordinate range/linspace for each dimension of the interpolation object.
+
+For every `NoInterp` dimension of the interpolation object, the range must be exactly `1:size(itp, d)`.
+""" ->
 function scale{T,N,IT,GT}(itp::AbstractInterpolation{T,N,IT,GT}, ranges::Range...)
     length(ranges) == N || throw(ArgumentError("Must scale $N-dimensional interpolation object with exactly $N ranges (you used $(length(ranges)))"))
     for d in 1:N
@@ -38,6 +45,12 @@ boundstep(r::FloatRange) = r.step / 2
 boundstep(r::StepRange) = r.step / 2
 boundstep(r::UnitRange) = 1//2
 
+@doc """
+Returns *half* the width of one step of the range.
+
+This function is used to calculate the upper and lower bounds of `OnCell` interpolation objects.
+""" -> boundstep
+
 coordlookup(r::LinSpace, x) = (r.divisor * x + r.stop - r.len * r.start) / (r.stop - r.start)
 coordlookup(r::FloatRange, x) = (r.divisor * x - r.start) / r.step + one(eltype(r))
 coordlookup(r::StepRange, x) = (x - r.start) / r.step + one(eltype(r))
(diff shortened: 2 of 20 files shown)


gaabriel:
Sounds good! Presumably it will be possible to make this one call, e.g. `interpolate((x1s, x2s), ys, BSpline(Linear), OnGrid)` or maybe `sinterpolate` to prevent confusion?

tlycken:
@gaabriel, actually, I'm inclined towards *not* implementing that method, for the following reason:

In the long run, I'm hoping to be able to include more types of interpolations than B-splines in this library. It would be nice if things like re-scaling of coordinate axes (at least for uniform grids), extrapolation and maybe other functionality could share code between the different types of interpolation. Currently, this is handled through composition, so for example the following is possible:

```julia
itp = interpolate(ys, BSpline(Linear), OnGrid)
extp = extrapolate(itp, Line) # not implemented yet, but it will be :)
sctp = scale(extp, x1s, x2s)
```

Now, `sctp` interpolates `ys` linearly, extrapolates linearly for OOB coordinates, and has its axes re-scaled to `x1s` and `x2s`. If extrapolation (or scaling) is unimportant for a use case, just leave that row out.

If we were to include all possible combinations of compositions as methods on `interpolate`, the number of methods would explode, and I think it would be much more difficult to maintain.

Of course, if doing it all in one line is important to you, there's nothing preventing you from passing the interpolation object directly to the wrapper:

```julia
sctp = scale(extrapolate(interpolate(ys, BSpline(Linear), OnGrid), Line), x1s, x2s)
```

although admittedly this doesn't read as nicely.

simonbyrne:
Looks great, no objections from me!

src/scaling/scaling.jl
```diff
@@ -0,0 +1,23 @@
...
+end
+
+@generated function getindex{T,N}(sitp::ScaledInterpolation{T,N}, xs...)
+    length(xs) == N || error(string("Must index into ScaledInterpolation{", N, "} with exactly ", N, " indices (you used ", length(xs), ")"))
+    :(getindex(sitp.itp, coordlookup(sitp.ranges, tuple(xs...))...))
+end
+
+size(sitp::ScaledInterpolation, d) = size(sitp.itp, d)
+
+coordlookup(r::FloatRange, x) = (r.divisor * x - r.start) ./ r.step + 1.0
```
timholy: Here and elsewhere, you probably want `one(eltype(r))` rather than `1.0` (think `Float32`).

timholy: Come to think of it, it wouldn't be bad to add some `Float32` tests...

tlycken: Yeah, tests for `Float32` would be nice. Maybe `Rational`s, `Complex` and some other non-trivial types too...

Would e.g. `1//1` do here? Promotion should take care of the rest then, I imagine.

timholy: Probably slower than using `one` directly, I assume. Does that have any failure cases?

tlycken: True; promotion costs. `one` is probably better then; in this case, it's unambiguous what type is the right one, so it'll be difficult to get it wrong. (I was making parallels to the discussions we had about literal constants elsewhere in the library a while ago...)

tlycken:
@timholy I've started working on implementing support for `NoInterp`, and I realized that it would be very useful to have `NoInterp` be independent of all the B-spline stuff, in order to keep `scale` and `extrapolate` agnostic of what interpolation scheme the underlying interpolation object is using.

Currently, the definition is as follows:

```jl
immutable NoInterp <: Degree(0) end
```

but I'd much rather have

```jl
immutable NoInterp <: InterpolationType
```

i.e. to put it on equal level with `BSpline{D<:Degree}` in the type hierarchy. However, that will make it not be a subtype of `DimSpec{BSpline}`, which breaks all sorts of things (because a lot of the type parameter specifications on the internal methods have `IT<:DimSpec{BSpline}` on them, and so `NoInterp` won't meet that requirement anymore), and I'm not quite sure how to resolve that properly. If I just remove the requirement for a subtype of `BSpline` (i.e. `IT<:DimSpec`) non-B-spline interpolation types will match. If I don't, `NoInterp` won't match.

Do you have any suggestions on how best to move forward?

tlycken:
(Re: the current test failure; that's a `@test_approx_eq a b` that should probably have been `@test_approx_eq_eps a b cbrt(eps(a))` - I'll fix that in a later commit.)

src/scaling/scaling.jl
```diff
@@ -0,0 +1,53 @@
...
+end
+
+size(sitp::ScaledInterpolation, d) = size(sitp.itp, d)
+
+coordlookup(r::FloatRange, x) = (r.divisor * x - r.start) ./ r.step + one(eltype(r))
+coordlookup(r::StepRange, x) = (x - r.start) ./ r.step + one(eltype(r))
+coordlookup(r::UnitRange, x) = x - r.start + one(eltype(r))
+coordlookup(i::Bool, r::Range, x) = i ? coordlookup(r, x) : x
+coordlookup{N}(r::NTuple{N}, x::NTuple{N}) = map(coordlookup,r,x)
+coordlookup{N}(i::NTuple{N}, r::NTuple{N}, x::NTuple{N}) = map(coordlookup, i, r, x)
```
tlycken: I'm seeing a type instability in (some of) these methods that I can't pin down and explain ([reproduction script](https://gist.github.com/tlycken/18db1ed677cd9074b8fe)). It seems that Julia is unable to infer the type even of `coordlookup(2:2.3:10, 5.1)`, which is super-weird, since the quantities involved in calculations in that case are `Float64`.

```
julia> @code_warntype(Interpolations.coordlookup(2:2.3:10, 4.1))
Variables:
  r::FloatRange{Float64}
  x::Float64

Body:
  begin  # C:\Users\Tomas Lycken\.julia\v0.4\Interpolations\src\scaling/scaling.jl, line 33:
      return (Base.box)(Base.Float64,(Base.add_float)((Base.box)(Base.Float64,(Base.div_float)((Base.box)(Base.Float64,(
Base.sub_float)((Base.box)(Base.Float64,(Base.mul_float)((top(getfield))(r::FloatRange{Float64},:divisor)::Float64,x::Fl
oat64)::Any)::Float64,(top(getfield))(r::FloatRange{Float64},:start)::Float64)::Any)::Float64,(top(getfield))(r::FloatRa
nge{Float64},:step)::Float64)::Any)::Float64,(Base.box)(Float64,(Base.sitofp)(Float64,1)::Any)::Float64)::Any)::Float64
  end::Float64
```

Note that there's a bunch of boxing operations as well as `Any` inferred. I have no idea what to do about this.

src/scaling/scaling.jl
```diff
@@ -0,0 +1,53 @@
...
+        :(getindex(sitp.itp, coordlookup(sitp.ranges, tuple(xs...))...))
+    end
+end
+
+size(sitp::ScaledInterpolation, d) = size(sitp.itp, d)
+
+coordlookup(r::FloatRange, x) = (r.divisor * x - r.start) ./ r.step + one(eltype(r))
+coordlookup(r::StepRange, x) = (x - r.start) ./ r.step + one(eltype(r))
+coordlookup(r::UnitRange, x) = x - r.start + one(eltype(r))
+coordlookup(i::Bool, r::Range, x) = i ? coordlookup(r, x) : x
```
tlycken: I've changed this line locally to `coordlookup(i::Bool, r::Range, x) = i ? coordlookup(r,x) : convert(typeof(coordlookup(r,x)), x)` to try to fix the type instability (see comment below) but it doesn't help.

KristofferC:
The boxes and Anys are not a problem. For example

```jl
julia> @code_warntype sqrt(5.0)
Variables:
  x::Float64

Body:
  begin  # math.jl, line 144:
      return (Base.Math.box)(Base.Math.Float64,(Base.Math.sqrt_llvm)(x::Float64)::Any)::Float64
  end::Float64
```

What is important is the output which for `coordlookup` is just Float64. I would instead focus on the `map` call in getindex and the splatting of the result. These can be problematic for type inference.

timholy:
You have a run-time anonymous function:
```jl
    # handle dimspecs
    interp_indices = map(it -> IT.parameters[it] != NoInterp, 1:length(IT.parameters))
```
Those are always type-unstable. You could use FastAnonymous to make it type-stable, but it's also always unstable to access the `.parameters` field.

Since `IT.parameters` is available to compile-time portion of the `@generated` function, you can (and should) do all this processing before returning the `Expr`.

tlycken:
@timholy That was intended to be compile-time (note that the quoting is only around the actual `getindex` call, where `interp_indices` have already been calculated and are interpolated into the returned expression).

And indeed; if I add a `@show interp_indices` after the calculation there, it only outputs something on the first execution of `getindex` (and it seems to know what type it is, too).

timholy:
Shoot, I was mixing two things up. Glad you figured it out.

For performance reasons, you probably also don't want `...` in returned expressions: you probably want to build the exprs like this:
```jl
lookup_indices = [:(coord_lookup(sitp.ranges, indx)) for index in interp_indices]
:(getindex(sitp.itp, $(lookup_indices...)))
```

tlycken:
That seems to have solved it. I haven't done any benchmarks, but if we gained a few extra notches on the performance scale thanks to your tips I'm not complaining :)

timholy:
Great to hear. This looks nice, BTW!

tlycken:
I think this is all done now - would gladly have someone look at it and see if there's something more that should be done (e.g. if something more needs documentation - I've only documented the exported API and a couple of the internal helper functions whose purpose I thought wasn't necessarily obvious from the name, but I haven't documented *everything*...)

src/scaling/scaling.jl
```diff
@@ -0,0 +1,87 @@
+type ScaledInterpolation{T,N,ITPT,IT,GT,RT} <: AbstractInterpolationWrapper{T,N,ITPT,IT,GT}
+    itp::ITPT
+    ranges::RT
+end
+ScaledInterpolation{T,ITPT,IT,GT,RT}(::Type{T}, N, itp::ITPT, ::Type{IT}, ::Type{GT}, ranges::RT) =
+    ScaledInterpolation{T,N,ITPT,IT,GT,RT}(itp, ranges)
+@doc """
+`scale(itp, xs, ys, ...)` scales an existing interpolation object to allow for indexing using other coordinate axes than unit ranges, by wrapping the interpolation object and transforming the indices from the provided axes onto unit ranges upon indexing.
```
timholy: In a sense, `(xs,ys,...)` is equivalent to `knots` in the Gridded interpolant, with the very important difference that Range inputs provide very nice extra algorithmic guarantees that make everything more efficient and mesh seamlessly with our `BSpline` infrastructure.

Given that *conceptual* equivalence, however, I suspect we should not have `knots` come first with `Gridded` and `xs, ys, ...` come second with `scale`. I would propose `scale(knots, itp)` but I am comfortable changing `Gridded` instead.

src/scaling/scaling.jl
```diff
@@ -0,0 +1,87 @@
...
+
+The parameters `xs` etc must be either ranges or linspaces, and there must be one coordinate range/linspace for each dimension of the interpolation object.
+
+For every `NoInterp` dimension of the interpolation object, the range must be exactly `1:size(itp, d)`.
```
timholy: Should we allow `Colon` as well?

src/scaling/scaling.jl
```diff
@@ -0,0 +1,87 @@
...
+lbound{T,N,ITPT,IT}(sitp::ScaledInterpolation{T,N,ITPT,IT,OnCell}, d) = 1 <= d <= N ? sitp.ranges[d][1] - boundstep(sitp.ranges[d]) : throw(BoundsError())
+ubound{T,N,ITPT,IT}(sitp::ScaledInterpolation{T,N,ITPT,IT,OnGrid}, d) = 1 <= d <= N ? sitp.ranges[d][end] : throw(BoundsError())
+ubound{T,N,ITPT,IT}(sitp::ScaledInterpolation{T,N,ITPT,IT,OnCell}, d) = 1 <= d <= N ? sitp.ranges[d][end] + boundstep(sitp.ranges[d]) : throw(BoundsError())
+
+boundstep(r::LinSpace) = ((r.stop - r.start) / r.divisor) / 2
+boundstep(r::FloatRange) = r.step / 2
+boundstep(r::StepRange) = r.step / 2
+boundstep(r::UnitRange) = 1//2
+
+@doc """
```
timholy: `@doc` is needed only for 0.3 compatibility or fancy stuff like RST. You can just replace this with `"""`.

I've been meaning to go through and add inline documentation, thanks for starting this effort.

src/scaling/scaling.jl
```diff
@@ -0,0 +1,87 @@
...
+            length(ranges[d]) == size(itp,d) || throw(ArgumentError("The length of the range in dimension $d ($(length(ranges[d]))) did not equal the size of the interpolation object in that direction ($(size(itp,d)))"))
+        elseif ranges[d] != 1:size(itp,d)
+            throw(ArgumentError("NoInterp dimension $d must be scaled with unit range 1:$(size(itp,d))"))
+        end
+    end
+
+    ScaledInterpolation(T,N,itp,IT,GT,ranges)
+end
+
+@generated function getindex{T,N,ITPT,IT<:DimSpec}(sitp::ScaledInterpolation{T,N,ITPT,IT}, xs...)
```
timholy: Change to `xs::Number` so we get the goodness of julia #10525.

src/scaling/scaling.jl
```diff
@@ -0,0 +1,87 @@
...
+This function is used to calculate the upper and lower bounds of `OnCell` interpolation objects.
+""" -> boundstep
+
+coordlookup(r::LinSpace, x) = (r.divisor * x + r.stop - r.len * r.start) / (r.stop - r.start)
+coordlookup(r::FloatRange, x) = (r.divisor * x - r.start) / r.step + one(eltype(r))
+coordlookup(r::StepRange, x) = (x - r.start) / r.step + one(eltype(r))
+coordlookup(r::UnitRange, x) = x - r.start + one(eltype(r))
+coordlookup(i::Bool, r::Range, x) = i ? coordlookup(r, x) : convert(typeof(coordlookup(r,x)), x)
+
+gradient{T,N}(sitp::ScaledInterpolation{T,N}, xs...) = gradient!(Array(T,N), sitp, xs...)
```
timholy: I'd put `xs::Number` here too.

test/scaling/dimspecs.jl
```diff
@@ -0,0 +1,26 @@
...
+f(x,y) = sin(x) * y^2
+
+itp = interpolate(Float64[f(x,y) for x in xs, y in ys], Tuple{BSpline(Quadratic(Periodic)), BSpline(Linear)}, OnGrid)
+sitp = scale(itp, xs, ys)
+
+# Don't test too near the edges until issue #64 is resolved
+for (ix,x0) in enumerate(xs[5:end-5]), (iy,y0) in enumerate(ys[2:end-1])
+    x, y = x0 + 2pi/20, y0 + .05
+    @test_approx_eq sitp[x0, y0] f(x0,y0)
+    @test_approx_eq_eps sitp[x0, y0] f(x0,y0) 0.05
```
timholy: Why both?

test/scaling/dimspecs.jl
```diff
@@ -0,0 +1,26 @@
...
+
+    g = gradient(sitp, x, y)
+    fx = epsilon(f(dual(x,1), y))
+    fy = (f(x, ys[iy+2]) - f(x, ys[iy+1])) / (ys[iy+2] - ys[iy+1])
+
+    @test_approx_eq_eps g[1] fx 0.15
+    @test_approx_eq_eps g[2] fy 0.05 # gradients for linear interpolation is "easy"
```
timholy: Consistency against the corresponding continuous function can of course only be expected to within relatively low precision, but we should insist on *internal* consistency of an interpolant to within machine precision. In other words, someone who uses finite-differencing or, better, DualNumbers to estimate the gradient of an interpolant should get the exact same answer we return from `gradient`---if that doesn't happen, then we've screwed up the gradient formula somehow.

For a 1d interpolant:
```jl
g = gradient(itp, x)
v = itp[dual(x,1)]
@test_approx_eq epsilon(v) g[1]
```
This is very different from (and much more important than) the tests above.

test/scaling/nointerp.jl
```diff
@@ -0,0 +1,38 @@
...
+for (ix,x0) in enumerate(xs[1:end-1]), y0 in ys
+    x,y = x0, y0
+    @test_approx_eq_eps sitp[x,y] f(x,y) .05
+end
+
+# Test error messages for incorrect initialization
+function message_is(message)
+	r -> r.err.msg == message || error("Incorrect error message: expected '$message' but was '$(r.err.msg)'")
+end
+Test.with_handler(message_is("Must scale 2-dimensional interpolation object with exactly 2 ranges (you used 1)")) do
```
timholy: I wasn't familiar with this trick, very nice.

test/scaling/nointerp.jl
```diff
@@ -0,0 +1,38 @@
...
+Test.with_handler(message_is("NoInterp dimension 2 must be scaled with unit range 1:3")) do
+	@test scale(itp, xs, -1:1)
+end
+Test.with_handler(message_is("The length of the range in dimension 1 (8) did not equal the size of the interpolation object in that direction (11)")) do
+	@test scale(itp, -pi:2pi/7:pi, 1:3)
+end
+Test.with_handler(message_is("Must index into 2-dimensional scaled interpolation object with exactly 2 indices (you used 1)")) do
+	@test sitp[2.3]
+end
+end
```
timholy: perhaps worth testing that `gradient!` should expect a vector of length 1 in the `Tuple{BSpline(Quadratic(Periodic)), NoInterp}` case.

test/scaling/scaling.jl
```diff
@@ -0,0 +1,52 @@
...
+
+sitp = scale(itp, -3:.5:1.5)
+
+for (x,y) in zip(-3:.05:1.5, 1:.1:10)
+    @test_approx_eq sitp[x] y
+end
+
+# Verify that it works in >1D, with different types of ranges
+
+gauss(phi, mu, sigma) = exp(-(phi-mu)^2 / (2sigma)^2)
```
timholy: Do remember that if you choose a function that is "bi-quadratic" (i.e., `∂^3f/∂x^3 = 0` in each of its variables) and use the `InPlaceQ` boundary condition, then interpolation should match the function to within machine precision. By choosing something that agrees with the underlying assumptions of the interpolation model, we can build a more rigorous test.

But that can be tackled later, no need to hold up this excellent PR now.

test/scaling/withextrap.jl
```diff
@@ -0,0 +1,40 @@
+
+module ScalingWithExtrapTests
```
timholy: I love how composable this framework is, and that you just get this "for free" :smile:.

timholy:
Very nice! Most importantly, I'm already looking forward to using this! Thanks for tackling it.

src/Interpolations.jl
```diff
@@ -4,6 +4,7 @@ export
     interpolate,
     interpolate!,
     extrapolate,
+    scale,
```
timholy: Maybe also export `ScaledInterpolation`?

test/scaling/dimspecs.jl
tlycken: Yes, that's a good point. There are tests like that elsewhere - I'll add some here too.

test/scaling/nointerp.jl
tlycken: Yes, I was pretty happy when I came up with this myself :smile: 

In fact, I might make a PR against base Julia to add a macro that does this, maybe as a method of `@test_throws` that also verifies the error message.

test/scaling/nointerp.jl
tlycken: Good point. We should probably also test then that `gradient` returns a vector of length 1 for such an interpolant.

test/scaling/withextrap.jl
tlycken: It wasn't completely for free - I had to add functions `bounds`, `lbound` and `ubound` to make this work - but the composability was one major argument for this design in the first place. I'm glad you like it :)

src/Interpolations.jl
tlycken: I don't think there's need for that; there's nothing you can do with that constructor that `scale` can't do (unlike `FilledExtrapolation`, where there's an actual use case for explicitly using the constructor).

We should, though, probably export `AbstractInterpolation`, to make it easier to create functions that use these objects. But the whole point of the composability design is that you shouldn't have to know which the concrete type is.

timholy: 5 minutes ago I wrote this method (see, I'm already using it :smile:):
```jl
function GridDeformation{FV<:FixedVector}(u::ScaledInterpolation{FV})
    N = length(FV)
    ndims(u) == N || throw(DimensionMismatch("Dimension $(ndims(u)) incompatible with vectors of length $N"))
    GridDeformation{eltype(FV),N,typeof(u),typeof(u.ranges[1])}(u)
end
```
But of course I can just add the module name, if you don't want to export it.

src/scaling/scaling.jl
tlycken: I agree that the conceptual similarity is a good argument for similarity in API as well. However, for the same reason, but looking at `extrapolate` instead, I'd like to keep `scale(itp, ranges...)` this way, since both `scale` and `extrapolate` essentially say "take this interpolant (the first argument) and add features to it according to the rest of the parameters".

tl;dr: I actually think it's better to change `Gridded` instead, even though having `knots` as the first argument is probably more intuitive in isolation.

src/scaling/scaling.jl
tlycken: For `NoInterp` dimensions, `Colon` actually does make sense, but it doesn't for any other types of interpolation objects. I'd be OK with supporting the cases that make sense, but that might be out of scope for this PR (since I don't think we handle that in the underlying interpolation implementation either).

src/scaling/scaling.jl
tlycken: Ah, nice. (I just copied your syntax from the `filled.jl`, because I was too lazy to actually go to the documentation... :wink: )

test/scaling/dimspecs.jl
tlycken: I think this is actually residue from trying to figure out good ways to test the scaling. I ended up doing just one test, on the grid points, since if that works then it's reasonably safe to assume that the coordinate scaling has worked correctly. Interpolating between grid points is then tested elsewhere.

src/Interpolations.jl
tlycken: Nvm, then - I stand corrected :)

timholy: Or maybe better is a new function `knots` that applies to any `itp`.

Also, now that I think about it, `scaled(knots, ::GriddedInterpolant)` should probably return another `GriddedInterpolant`, one that shares the data array but has new knots.

tlycken: In that case, I'd argue for `scale(itp, knots...)` as the common interface, returning a `ScaledInterpolant` in most cases, but where `scale(itp::GriddedInterpolation, knots...)` can special-case for re-use of the data array.

src/scaling/scaling.jl
timholy: Doesn't have to be part of this PR (or happen ever). But it's fine even for interpolating dimensions; after all, you let me supply `1:size(A,d)` for one of the knot vectors.

tlycken:
There, I think I addressed all the comments (at least where I'm not of the opinion that nothing should change - you may still change my mind... ;)).

The things I haven't done anything about (and where there are thus still line comments open):

* Using a bi-quadratic function for testing of 2D scaling
* `Colon` indexing
* Argument ordering for `scale(itp, knots...)`

The two first ones I think can wait until later, and the third I'm pretty sure I want as it is.

Any further comments, or is it hit-the-button-time?

src/scaling/scaling.jl
timholy: Fine with me. I chose the first mostly because Matlab [does it that way](http://www.mathworks.com/help/matlab/ref/griddedinterpolant-class.html?refresh=true).

timholy: Although now that I think about it, for `Gridded` it's not adding features; it's a different algorithm. For arbitrary knot vectors you have to use `searchsorted` to find the right spot in the data array.

Also note [scipy](http://docs.scipy.org/doc/scipy-0.15.1/reference/generated/scipy.interpolate.griddata.html).

timholy:
Just posted a couple of other comments above.

To be clear, while I advocate the `knots` first, in the end I don't really care. Hit the button at will :smile:.

src/scaling/scaling.jl
tlycken: Yes - I agree that `interpolate(knots, data, Gridded)` is a better interface, as is probably `scale(knots, itp)` if that was the only thing we did (and the fact that both MATLAB, SciPy and, I assume, a whole bunch of others do it that way is fuel on the fire).

In the context of composition of various "add-ons", though, I think keeping the API as close to `do_extra_stuff(itp, args...)` has a lot of value. `scale` and `extrapolate` are both examples of `do_extra_stuff` in this context. If we don't adhere to `(itp, args...)` for all of them, it will be a lot more difficult for users of *this* library to know (without resorting to docs) in which order the arguments should come.

timholy:
Hooray! Again, thanks for another fantastic addition.

tlycken:
:tada: :balloon: :beers: 

Next up: #50... ;)

timholy:
This package quite desperately needs tagging. I am happy to do it, but the question is: "patch" or "minor." I think the only breaking change is the `FilledInterpolation->FilledExtrapolation`, but neither type was available in the previously-tagged version. So perhaps "patch" is sufficient?

(Alternatively, we could use the fancy new `@deprecate_binding`, but that will introduce a dependency on julia-0.4-rc2.)

tlycken:
I think minor could be justified anyway, given all the new functionality we're bringing in. I'll tag when I'm done with #50, but feel free to do it yourself if you don't want to wait. 

I don't see the need for deprecation of features we've never actually released :)

PR #48, opened 2015-07-22
Simplify indexing

tlycken:
This removes some code that's no longer needed, thanks to JuliaLang/julia#10525. Hopefully, this will make it a little easier to extend the package in the future.

@timholy I assume that the method taking a `CartesianIndex` is an optimization. Is this somehow also provided by base Julia now?


timholy:
It was more likely there to avoid an ambiguity warning. Should not be necessary anymore, I assume.

:+1:

tlycken:
There's still some work to do here, though, as noted in https://github.com/JuliaLang/julia/pull/12273#issuecomment-123996106. Now that I know that you (probably) don't need the method for something specific, I can just remove it if I find that there's another with cartesian indexes provdided by base :)

PR #49, opened 2015-07-22
Latex portability

tlycken:
@timholy You mentioned that you didn't have `biber`, and thus couldn't compile the docs properly. With this change, I get a correct pdf with all references by running `pdflatex` (twice) and without ever running `biber` (or any other bibliography tool).

I feel quite confident that this should work in your end too, but please let me know if you have any problems if you're based on commits after this one.

Obviously, someone with `biber` installed will have to re-generate the `.bbl` if we want to add or change a reference, but I think that's a quite rare case - mostly we'll just write our own stuff :)


timholy:
Yep, works perfectly. Thanks!

tlycken:
:+1:

PR #51, opened 2015-07-23
RFC: Convert integer inputs to float before interpolating

### LINKED ISSUE #38: Fix integer-valued array inputs
Currently this is broken (and it may be my fault):
```
julia> using Interpolations

julia> A = rand(1:20, 3, 3)
3x3 Array{Int64,2}:
 19  20   5
 17  19  14
 19   9   9

julia> itp = interpolate(A, BSpline(Quadratic(Flat)), OnCell);
ERROR: InexactError()
 in invert_column! at /home/tim/.julia/v0.4/Interpolations/src/b-splines/../filter1d.jl:38
 in _A_ldiv_B_md! at /home/tim/.julia/v0.4/Interpolations/src/b-splines/../filter1d.jl:21
 in A_ldiv_B_md! at /home/tim/.julia/v0.4/Interpolations/src/b-splines/../filter1d.jl:14
 in prefilter! at /home/tim/.julia/v0.4/Interpolations/src/b-splines/prefiltering.jl:38
 in prefilter at /home/tim/.julia/v0.4/Interpolations/src/b-splines/prefiltering.jl:26
 in interpolate at /home/tim/.julia/v0.4/Interpolations/src/b-splines/b-splines.jl:36
 in interpolate at /home/tim/.julia/v0.4/Interpolations/src/b-splines/b-splines.jl:39
```
Fix this once ~~#34~~#36 and ~~#36~~#37 get merged.

tlycken:
This should fix #38, but I'm quite sure the solution is not optimal, for the following reasons:

* It converts integer inputs to floats for *all* inputs, but it shouldn't be necessary if no interpolations are to be done in higher dimensions than linear, since then there's no matrix inversion involved.
* I'm quite sure there's an extra copy now - first from `A` to `float(A)`, then from `float(A)` before prefiltering, but if I forward to `interpolate!` instead of `interpolate` I get garbage results in my tests. (Maybe I just misunderstood how in-place interpolation is implemented, and tested too close to the edges...)

### DIFF
--- a/src/b-splines/b-splines.jl
+++ b/src/b-splines/b-splines.jl
@@ -49,6 +49,7 @@ end
 interpolate{IT<:DimSpec{BSpline},GT<:DimSpec{GridType}}(A::AbstractArray, ::Type{IT}, ::Type{GT}) = interpolate(Float64, A, IT, GT)
 interpolate{IT<:DimSpec{BSpline},GT<:DimSpec{GridType}}(A::AbstractArray{Float32}, ::Type{IT}, ::Type{GT}) = interpolate(Float32, A, IT, GT)
 interpolate{IT<:DimSpec{BSpline},GT<:DimSpec{GridType}}(A::AbstractArray{Rational{Int}}, ::Type{IT}, ::Type{GT}) = interpolate(Rational{Int}, A, IT, GT)
+interpolate{T<:Integer,IT<:DimSpec{BSpline},GT<:DimSpec{GridType}}(A::AbstractArray{T}, ::Type{IT}, ::Type{GT}) = interpolate(float(A), IT, GT)
 
 interpolate!{TWeights,IT<:DimSpec{BSpline},GT<:DimSpec{GridType}}(::Type{TWeights}, A, ::Type{IT}, ::Type{GT}) = BSplineInterpolation(TWeights, prefilter!(TWeights, A, IT, GT), IT, GT, Val{0}())
 interpolate!{IT<:DimSpec{BSpline},GT<:DimSpec{GridType}}(A::AbstractArray, ::Type{IT}, ::Type{GT}) = interpolate!(Float64, A, IT, GT)
--- a/src/b-splines/b-splines.jl
+++ b/src/b-splines/b-splines.jl
@@ -42,14 +42,17 @@ padextract(pad::Tuple{Vararg{Integer}}, d) = pad[d]
     end
 end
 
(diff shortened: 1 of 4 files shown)


timholy:
Thanks for reminding me about #38, I had completely forgotten about it. And even more for tackling it yourself!

I think we should try to avoid making two copies; perhaps the right place is in the padding step?

`interpolate!` should probably just throw an error if given an array of `Integer` eltype.

tlycken:
> I think we should try to avoid making two copies; perhaps the right place is in the padding step?

I thought so too, but found myself unable to write the code that actually did that. Perhaps one of the methods that take a `TWeights` argument can be utilized - I'll have to trial-and-error some more :)

tlycken:
This now takes a more thorough approach, and, IIUC, avoids copying the array more than once. The approach is the following (from inside to out):

1. In `prefilter`, calls to `copy(A)` were changed to `copy!(similar(A, TCoefs), A)`. This mandated an extra type argument `TCoefs` to be sent when calling `prefilter.
2. All methods of `copy_with_padding` were updated to also take the type argument `TCoefs`, in order to forward it
3. The main implementation of `interpolate` also takes the type parameter `TCoefs` and forwards it
4. If no type arguments are given to `interpolate`, the following defaults are chosen:

| `eltype(A)` | `TWeights` | `TCoefs` |
| ----------- | ---------- | -------- | 
|    `T`      |  `Float64` | `eltype(A)` |
|  `Float32`  |  `Float32` |  `Float32`  |
| `Rational{Int}` | `Rational{Int}` | `Rational{Int}` |
| `T<:Integer` | `TFloat = typeof(float(zero(T)))` | `TFloat` |

For linear interpolation (which copies but does no matrix inversion) I think this indicates that the array is only copied once:

```
julia> A = rand(1:20, 100, 100);
# warmup of @time, copy! and interpolate omitted
julia> @time copy!(similar(A,Float64),A);
  49.465 microseconds (6 allocations: 80208 bytes)
julia> @time itp = interpolate(A, BSpline(Linear), OnGrid);
  72.627 microseconds (7 allocations: 80224 bytes)
```

For quadratic splines I get much more allocations, but I'm going to assume that those are due to the matrix inversion algorithms and not copying the matrix one extra time:

```
julia> @time itp = interpolate(A, BSpline(Quadratic(Flat)), OnGrid);
   1.097 milliseconds (277 allocations: 301 KB)
```

Good enough? :)

timholy:
Simply gorgeous. Approved wholeheartedly.

PR #53, opened 2015-08-06
Revamp the README

timholy:
To me it seems time to "advertise" this package a little more broadly, so I'll have a few PRs tweaking things. This one brings the README into sync with current syntax. Any and all suggestions welcome, of course.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,25 @@
 
 [![Build Status](https://travis-ci.org/tlycken/Interpolations.jl.svg?branch=master)](https://travis-ci.org/tlycken/Interpolations.jl)
 
-This package is the continuation of Tim Holy's work on [Grid.jl](https://github.com/timholy/Grid.jl), with the explicit goal of reaching feature parity for [B-splines](https://en.wikipedia.org/wiki/B-spline) in the near future (PRs are more than welcome!).
+This package implements a variety of interpolation schemes for the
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -14,33 +32,77 @@ Pkg.add("Interpolations")
...
 ```
+where `g` is a pre-allocated vector.
+
+`A` may have any element type that supports the operations of addition
+and multiplication.  Examples include scalars like `Float64`, `Int`,
+and `Rational`, but also multi-valued types like `RGB` color vectors.
 
-## Options
+Positions `(x, y, ...)` are n-tuples of numbers. Typically these will
+be real-valued (not necessarily integer-valued), but can also be types
```
tlycken: I'd say the positions can be *of* types - we support sending in `1.0 + 1.0du`, but not `Dual{Float64}`... It's a small  difference (in wording - bigger semantically), but I think it pays to be nit-picky in documentation :)

tlycken:
Wonderful writing, as always :)

I had only one nit-pick (see the line comment) - not sure if I can edit that directly on your branch or not. If you don't have time to edit that, just merge as-is and I'll fix it afterwards.

Thanks a lot for taking the time to shape this up!

README.md
timholy: Agreed.

PR #54, opened 2015-08-06
Restrict scalar indexing to Numbers

timholy:
This lets us exploit Julia's new indexing utilities.

For the `linspace` thing to work, it will require a to-be-submitted julia PR, to which I'll link shortly.


timholy:
https://github.com/JuliaLang/julia/pull/12486

tlycken:
LGTM! Thanks for taking the time! :)

Referencing #48 here too, for completeness.

timholy:
> we might want to add a couple of sanity tests for this, too.

Ah, right; I should have waited to hit merge. I'll file an issue to remind myself.

PR #59, opened 2015-08-08
Improve performance by reducing bounds-checking

timholy:
This puts us on top of the heap for every case. If the user supplies a `Range` `r` for the knots, it hurts performance because `r[i]` always checks bounds. It's worth `collect`ing `r` to an `Array` at construction time. (Doing this has another benefit: it's equivalent to making a copy of the knot-vectors, which ensures that the user won't accidentally bork the interpolation object.)


tlycken:
Simply beautiful.

Issue #60, opened 2015-08-09
Combining knots and NoInterp


matthieugomez:
I'm trying to use `interpolate` with knots & `NoInterp` along some dimension. 

- With BSpline, using knots as a first argument returns an error
 
 ```julia
 A = [0.0593468  0.73381 ; 0.298259   0.135074]
 itp = interpolate((1:2, 2:3), A, BSpline{Linear}, OnGrid)
 ```
I'm not sure whether `knots` are available for `Spline` 
- With Grid, combining `knots` and `Tuple{Gridded{Linear}, Gridded{NoInterp}` returns unexpected results

 ```julia
 A = [0.0593468  0.73381 ; 0.298259   0.135074]
 itp = interpolate((1:2, 2:3), A, Tuple{Gridded{Linear}, Gridded{NoInterp}})
 # itp[1, 2] returns A[1, 2] but should return A[1, 1]
 itp[1, 2]
 # 0.7338095900816122
  ```

timholy:
`knots` are only available with Gridded. What's called `BSpline` is really "uniform B-spline."

tlycken:
For this use case, though, it looks like you could use the scaling approach that will be added by #47.

matthieugomez:
Yes I only need the scaling, so that would work. The Gridded example is a bug though, right?

timholy:
> The Gridded example is a bug though, right?

No, `NoInterp` means "just pass the index through." You might be looking for `Gridded{Constant}`.

timholy:
Might need some API fixing though---we should probably throw an error for those knots. Care to submit a pull request?

matthieugomez:
I've looked into the (terse!) code for `Grid` but I'm not sure how to modify it. 
I think there should be an error when, along some dimension `d`, the type is `Gridded{NoInterp}` but the knots is not `1:size(A, d)`.

timholy:
Yep, that's how I'd do it. Nice digging!

timholy:
Fixed by #61.

matthieugomez:
Thanks!

PR #61, opened 2015-08-10
solves #60

matthieugomez:
- Returns error when using Grid{NoInterp} without the range 1:size(A, d)
- moves discussion of knots in the subsection for `Gridded`


timholy:
Looks great. I might still mention early in the docs that one can pass knot vectors, but I can tweak that later. (Or feel free to make a 2nd PR yourself, if you prefer.)

Thanks!

PR #63, opened 2015-09-09
Define length

timholy:
`size` is quite slow because it constructs the NTuple and uses
anonymous functions, so this makes a pretty big difference in certain loops.

### DIFF
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -25,7 +25,7 @@ export
 
 using WoodburyMatrices, Ratios, AxisAlgorithms
 
-import Base: convert, size, getindex, gradient, promote_rule
+import Base: convert, length, size, getindex, gradient, promote_rule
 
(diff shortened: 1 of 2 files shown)


tlycken:
This is just a way of getting at the count of elements in the wrapped matrix, right? It seems to me that the need to know that indicates that we have some leaky abstractions here. The `size` of an interpolation object can still be defined as the size of the domain in which it does not (need to) extrapolate (don't know if that's what we're doing now, but it seems reasonable to me anyway), but the fact that we're wrapping a matrix is very much an implementation detail.

Is the underlying problem rather that `size(::AbstractInterpolation)` is slow? Can we make it faster?

Out of curiosity, what's the use case for `length` on interpolation objects?

sglyon:
@timholy what is the status here?

timholy:
Not necessary anymore.

PR #66, opened 2015-09-17
RFC: Make NoInterp scheme-agnostic

tlycken:
This pull request makes `NoInterp <: InterpolationType` rather than `<: Degree{0}`.

The purpose of this is to pave way for supporting `NoInterp` in scheme-agnostic add-ons like scaling and extrapolation, without having to consider types like `BSpline{NoInterp}` and `Gridded{NoInterp}` separately.

As a side-effect, the API for non-interpolating dimensions is also made a little simpler. Now it's just e.g. `itp = interpolate(A, Tuple{BSpline(Quadratic(Flat)), NoInterp}, OnGrid)` to create an interpolation object where each column of `A` is its own interpolant.

I'm planning on rebasing #47 and #50 onto this, to make sure scaling and extrapolation actually support `NoInterp` correctly.


tlycken:
Oh, shoot - apparently I made some mistakes when rebasing to make this commit come before the scaling stuff I was working on... Give me a minute...

src/Interpolations.jl
```diff
@@ -28,11 +28,12 @@ using WoodburyMatrices, Ratios, AxisAlgorithms
 import Base: convert, size, getindex, gradient, promote_rule
 
 abstract InterpolationType
+immutable NoInterp <: InterpolationType end
 abstract GridType
 immutable OnGrid <: GridType end
 immutable OnCell <: GridType end
 
-typealias DimSpec{T} Union(T,Tuple{Vararg{T}})
+typealias DimSpec{T} Union(T,Tuple{Vararg{Union(T,NoInterp)}},NoInterp)
```
timholy: Sometime we should change this to `Union{T,...}` (braces instead of parens)

timholy:
Looks perfect. I'm unreasonably excited about this change, thanks!

timholy:
Oops, sorry, didn't mean to delete your branch. I do that out of habit with my own branches, but generally I try to let people make that decision themselves. Feel free to hit the "restore" button.

tlycken:
Re: branch deletion: No problem at all this time. Now that it's merged, I might as well base #47 and #50 on current master instead :)

src/Interpolations.jl
tlycken: I didn't know that was a thing, but when I re-read the manual now it seems parentheses aren't used anywhere anymore :P

I guess this is because we now prefer using the parametric type directly rather than through a constructor? (Will the constructor be deprecated soon-ish?)

timholy: On the latest release-0.4 branch, it already is. See, e.g., the slew of warnings in https://travis-ci.org/timholy/Images.jl/jobs/80997593. If you update your julia, you'll find a lot here, too.

tlycken: Then I guess I'll notice next time I'm on my Linux laptop. Working on Windows at the moment, so building from source is too much of a hassle...

PR #67, opened 2015-09-18
Re-include tests with other types

tlycken:
I guess these were excluded when even before we realized we needed `SimpleRatio`s, but I think it's time to include them again.

Perhaps we should add some more in the process, but my mind is all on getting #47 ready to merge...


timholy:
:+1:

PR #68, opened 2015-09-18
Documentation changes regarding Gridded

SloanLindsey:
Updated the documentation to give functional use examples for the Gridded mode.

Presently the visible front level documentation does not present the proper syntax to use the Gridded mode. This update makes the functionality more accessible for new users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ In 1D
 A = rand(20)
 A_x = collect(1.:40.)
 knots = (a,)
-itp = interpolate(knots,A, Tuple{Gridded{Linear}})
+itp = interpolate(knots,A, Gridded{Linear})
 itp[2.0] 
 ```
 
@@ -148,9 +148,14 @@ For example:
 ```jl
 A = rand(8,20)
 knots = ([x^2 for x = 1:8], [0.2y for y = 1:20])
-itp = interpolate(knots, A, Tuple{Gridded{Linear},Gridded{Linear}})
+itp = interpolate(knots, A, Gridded{Linear})
 itp[4,1.2]  # approximately A[2,6]
 ```
+One may also mix modes, by specifying a mode vector in the form of an explicit tuple:
+```jl
+itp = interpolate(knots, A, Tuple{Gridded{Linear},Gridded{Constant}})
+```
+
 
 Presently there are only three modes for gridded:
 ```jl
@@ -164,7 +169,12 @@ whereby nearest neighbor interpolation is used on the applied axis,
 ```jl
 Gridded{NoInterp}
 ```
-
+whereby the coordinate of the selected input vector MUST be located on a grid point. Requests for off grid
+coordinates results in the throwing of an error.
+```jl
+ERROR: InexactError()
+ in getindex at C:\dev\julia\.julia\v0.4\Interpolations\src\gridded\indexing.jl:14
+```
 
 
 ## Extrapolation
--- a/README.md
+++ b/README.md
@@ -125,27 +125,47 @@ which destroys the input `A` but also does not need to allocate as much memory.
 
 These use a very similar syntax to BSplines, with the major exception
 being that one does not get to choose the grid representation (they
-are all `OnGrid`).
+are all `OnGrid`). As such one must specify a set of coordinate arrays
+defining the knots of the array.
+
+In 1D
 ```jl
-A = rand(8,20)
-itp = interpolate(A, Gridded{Linear})
-itp[4,1.2] 
+A = rand(20)
+A_x = collect(1.:40.)
+knots = (a,)
+itp = interpolate(knots,A, Tuple{Gridded{Linear}})
+itp[2.0] 
 ```
 
-If the spacing between adjacent samples differs, then you can use the syntax
+The spacing between adjacent samples need not be constant, you can use the syntax
 ```jl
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -125,28 +125,48 @@ which destroys the input `A` but also does not need to allocate as much memory.
...
+
+In 1D
 ```jl
-A = rand(8,20)
-itp = interpolate(A, Gridded{Linear})
-itp[4,1.2] 
+A = rand(20)
+A_x = collect(1.:40.)
+knots = (a,)
+itp = interpolate(knots,A, Tuple{Gridded{Linear}})
```
tlycken: This `Tuple` shouln't be needed - does it not work for you without it? The minimum modification of the old example I need to get it working, is to add knot vectors:

```jl
A = rand(8,20)
knots = (collect(1:8.), collect(1:.5:10.5))
itp = interpolate(knots, A, Gridded{Linear})
itp[4, 1.2]
```

README.md
```diff
@@ -125,28 +125,48 @@ which destroys the input `A` but also does not need to allocate as much memory.
...
-each axis at which the array `A` is sampled.
+where `knots = (xknots, yknots, ...)` to specify the positions along
+each axis at which the array `A` is sampled for arbitrary ("rectangular") samplings.
 
 For example:
 ```jl
 A = rand(8,20)
 knots = ([x^2 for x = 1:8], [0.2y for y = 1:20])
-itp = interpolate(knots, A, Gridded{Linear})
+itp = interpolate(knots, A, Tuple{Gridded{Linear},Gridded{Linear}})
```
tlycken: This tuple shouln't be needed either; does it not work for you with just `Gridded{Linear}`?

README.md
```diff
@@ -125,28 +125,48 @@ which destroys the input `A` but also does not need to allocate as much memory.
...
+```jl
+Gridded{Linear}
+```
+whereby a linear interpolation is applied between knots,
+```jl
+Gridded{Constant}
+```
+whereby nearest neighbor interpolation is used on the applied axis,
+```jl
+Gridded{NoInterp}
```
tlycken: This syntax recently (earlier today!) changed to just `NoInterp` (see #66). It would be nice with a small description of what it does, too.

tlycken:
Very nice! Help with documentation is always extremely welcome! :)

I've added a few comments and questions; my main one being the fact that you've added `Tuple{...}` around the interpolation type specification. This shouldn't be necessary, so if it doesn't work for you without it we might have a bug somewhere. Could you post a complete example, with both input and output, where it doesn't work with just `Gridded{Linear}`?

tlycken:
Also, I notice that you've just opened your Github account to contribute here - welcome to the OSS community! :dancers:

README.md
SloanLindsey: Hmm, yup this works now. 
however if you wish to mix the types you can't do an implicit tuple creation
```jl
itp = interpolate(knots, A, (Gridded{Linear},Gridded{Linear}))
```
does not work.

SloanLindsey:
Things work with Gridded{Linear}, sorry I got caught up in earlier bugs and went to the syntax from #60 .
So the changes you've proposed work on my 0.4.0-rc1 release with no issues.
I didn't describe the case of NoInterp because I don't really understand what its useful for. We are producing some mapping that I assume points to the underlying grid of the interpolator. I don't understand what the functional difference is between NoInterp and Constant. 

amendment:
NoInterp forces the coordinate to exist on the grid points throwing an error if an inexact location is requested.
This functionality looks to be useful for error checking in some interesting use cases where for instance we want to force the interpolation input vector to contain a well defined integer value.
I'll add the description there too.

Also, thanks for the warm welcome. Though it seems like participation is not really optional in the Julia community, its small enough that going to the source control is the most practical way of addressing issues.

README.md
tlycken: No, it doesn't because the interface is a little fuzzy (I have plans for fixing this, but I haven't gotten there yet). You should supply a *type* (i.e. `Tuple{Gridded{Linear}, Gridded{Constant}}`) when you want to have different behavior in different directions. For same behavior in all directions, though, just `Gridded{Linear}` should do.

`(Gridded{Linear}, Gridded{Linear})` will not work with the current interface, since `(...)` is a *tuple* and not a *tuple type*. (To be specific, this particular tuple is of the type `Tuple{DataType, DataType}`.)

Does that make things clearer?

tlycken:
Thanks for the updates - looks good :)

For reference, the use case for `NoInterp` is when you have a bunch of unrelated data sets in the same array, and want to interpolate them (but it makes no sense to interpolate between them). For example, consider the array `xs = linspace(-pi, pi, 10); A = hcat(cos(xs), sin(xs))` - it is nice to be able to create an interpolation object `itp = interpolate(A, Tuple{Gridded{Linear}, NoInterp})` (note that it's not `Gridded{NoInterp}` as of #66) that can interpolate both `sin(x)` and `cos(x)`, but that doesn't try to interpolate something between them just because the initial data happened to be in one array.

timholy:
@tlycken, OK to rebase and merge this as-is?

tlycken:
Except for [this syntax change](https://github.com/tlycken/Interpolations.jl/pull/68#discussion-diff-39878420R165), go right ahead :)

PR #69, opened 2015-09-20
Fix Union(...) -> Union{...}

tlycken:
This should be fairly straightforward...


tlycken:
I don't know why the build won't start, but since this will be merged with #47 I'm not going to spend time figuring it out.

PR #70, opened 2015-09-20
RFC: Rename `FilledInterpolation` -> `FilledExtrapolation`

tlycken:
When I rebased #47 I put this first, so you can have a say in whether it's a good or bad idea, @timholy. But since this functionality is essentially extrapolation, I think this name is better :)


tlycken:
I don't know why the build won't start (I thing Travis might have been down or something just when I pushed this) but this will be merged with #47 anyway, so I don't think it's a big deal.

PR #72, opened 2015-09-27
Allow manual override of return type for fillvalue

timholy:
I'm playing with ForwardDiff as a way of computing gradients, and needed this to handle the fact that sometimes the same `FilledExtrapolation` object gets called with two different types of indexes (`Float64` and `GradientNumber{Float64,...}`). I added these lines (the last one being the relevant one here) to my own code:
```jl
# Some necessary ForwardDiff extensions
Base.real(v::ForwardDiff.GradientNumber) = real(v.value)
Base.ceil(::Type{Int}, v::ForwardDiff.GradientNumber)  = ceil(Int, v.value)
Base.floor(::Type{Int}, v::ForwardDiff.GradientNumber) = floor(Int, v.value)
# Type-stability in gradient computation
@generated function Interpolations.fillvalue{T<:ForwardDiff.GradientNumber}(val, args::T...)
    :(ForwardDiff.GradientNumber(val, args[1].partials.data))
end
```
and now ForwardDiff works fine.

### DIFF
--- a/src/gridded/indexing.jl
+++ b/src/gridded/indexing.jl
@@ -86,3 +86,11 @@ function getindex{T,N,TCoefs,IT<:DimSpec{Gridded},K,P}(itp::GriddedInterpolation
     dest = Array(T, map(length, x))::Array{T,N}
     getindex!(dest, itp, x...)
 end
+
+function getindex_return_type{T,N,TCoefs,IT<:DimSpec{Gridded},K,P}(::Type{GriddedInterpolation{T,N,TCoefs,IT,K,P}}, argtypes)
+    Tret = TCoefs
+    for a in argtypes
+        Tret = Base.promote_op(Base.MulFun, Tret, a)
+    end
+    Tret
+end
--- a/test/extrapolation/runtests.jl
+++ b/test/extrapolation/runtests.jl
@@ -27,9 +27,7 @@ etpf = @inferred(extrapolate(itpg, NaN))
 @test_throws BoundsError etpf[2.5,2]
 @test_throws ErrorException etpf[2.5,2,1]  # this will probably become a BoundsError someday
 
-etpf = @inferred(FilledExtrapolation(itpg, 'x'))
-@test_approx_eq etpf[2] f(2)
-@test etpf[-1.5] == 'x'
+@test isa(@inferred(getindex(etpf, dual(-2.5,1))), Dual)
 
 etpl = extrapolate(itpg, Linear)
 k_lo = A[2] - A[1]
--- a/src/b-splines/indexing.jl
+++ b/src/b-splines/indexing.jl
@@ -69,6 +69,15 @@ function gradient_impl{T,N,TCoefs,IT<:DimSpec{BSpline},GT<:DimSpec{GridType},Pad
     end
 end
 
+function getindex_return_type{T,N,TCoefs,IT<:DimSpec{BSpline},GT<:DimSpec{GridType},Pad}(::Type{BSplineInterpolation{T,N,TCoefs,IT,GT,Pad}}, argtypes)
+    Tret = TCoefs
+    for a in argtypes
+        Tret = Base.promote_op(Base.MulFun, Tret, a)
+    end
+    Tret
+end
+
+
 @generated function gradient!{T,N}(g::AbstractVector, itp::BSplineInterpolation{T,N}, xs::Number...)
     length(xs) == N || error("Can only be called with $N indexes")
     gradient_impl(itp)
--- a/src/extrapolation/extrapolation.jl
+++ b/src/extrapolation/extrapolation.jl
@@ -1,5 +1,4 @@
-export Throw,
-       FilledExtrapolation   # for direct control over typeof(fillvalue)
+export Throw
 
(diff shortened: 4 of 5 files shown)


tlycken:
I thought this was the purpose of [exporting the `FilledExtrapolation`](https://github.com/tlycken/Interpolations.jl/blob/master/src/extrapolation/extrapolation.jl#L2) constructor, or am I misunderstanding something? Can all the goals of these two things be accomplished with just one of them?

timholy:
It was, but that's a static mechanism. What if you call the same `FilledExtrapolation` object with two different types of indexes? For example, `itp[-3.2]` might give you `NaN`, but `itp[dual(-3.2,1.0)]` (with the same `itp`) should give you `dual(NaN, 1.0)`.

timholy:
Though I suppose we could use the `promote_op` mechanism instead?

tlycken:
My question was not as much "do we need this?" as "do we need both, or can this replace the other?" but if both are valuable I see no reason not to keep them both anyway. So it was more a curious question than criticism :)

But using `promote_op` sounds like a good idea to me - otherwise we'll basically re-implement a mechanism for the user to do the same thing, but hard-coded (and thus error-prone). Perhaps we can accomplish something similar to [how we calculate `T` when creating B-spline interpolation objects](https://github.com/tlycken/Interpolations.jl/blob/master/src/b-splines/b-splines.jl#L23-L27), and/or the array type for `gradient` (see below, soon to be included in the PR for extrapolation behaviors)?

```
@generated function gradient{T,N}(itp::AbstractInterpolation{T,N}, xs...)
    n = count_interp_dims(itp, N)
    Tg = promote_type(T, [x <: AbstractArray ? eltype(x) : x for x in xs]...)
    :(gradient!(Array($Tg,$n), itp, xs...))
end
```

Regardless of what the actual solution ends up being, I think we should try to a) make sure that we are compatible with the promotion mechanisms in base Julia as much as possible, and b) make the user-interface as seamless ass possible. Leveraging `promote_op` to accomplish a) can probably help with b), so that's a good place to start :)

tlycken:
Btw, the test failure should be fixed once #73 is merged.

timholy:
OK, see if you like this better (I do).

I deleted the docs for `FilledExtrapolation`, since they focused on controlling the return type (which has proven to be of very limited utility). Should we un-export the type, or leave it?

test/extrapolation/runtests.jl
```diff
@@ -27,6 +27,8 @@ etpf = @inferred(extrapolate(itpg, NaN))
 @test_throws BoundsError etpf[2.5,2]
 @test_throws ErrorException etpf[2.5,2,1]  # this will probably become a BoundsError someday
 
+@inferred(getindex(etpf, dual(-2.5,1)))
+
 etpf = @inferred(FilledExtrapolation(itpg, 'x'))
```
tlycken: Should this initialization be done some other way?

src/gridded/indexing.jl
```diff
@@ -86,3 +86,11 @@ function getindex{T,N,TCoefs,IT<:DimSpec{Gridded},K,P}(itp::GriddedInterpolation
     dest = Array(T, map(length, x))::Array{T,N}
     getindex!(dest, itp, x...)
 end
+
+function getindex_return_type{T,N,TCoefs,IT<:DimSpec{Gridded},K,P}(::Type{GriddedInterpolation{T,N,TCoefs,IT,K,P}}, argtypes)
+    Tret = TCoefs
+    for a in argtypes
+        Tret = Base.promote_op(Base.MulFun, T, a)
```
tlycken: Is this loop body intended to make use of `Tret = TCoefs` above somehow, or is that just a default in case `length(argtypes) == 0`?

tlycken:
> Should we un-export the type, or leave it?

Since exporting the type was mainly to make the type-controlling constructor available, I think we should probably un-export it. (That's a very simple breaking change to work around, should there be anyone who has already started using it - just say `FilledExtrapolation = Interpolations.FilledExtrapolation` and you're done...) Un-exporting also gives symmetry to other types, where almost no types - except the different configuration "flags" - are exported from the rest of the library.

test/extrapolation/runtests.jl
timholy: This test can probably just be dropped, since it was really testing explicit control.

timholy:
Thanks for the review (and catches). Should be addressed now. There's some weird SSL error on nightly (I restarted, and that didn't fix it), but I doubt it reflects a problem with this PR.

timholy:
All package PRs are failing this way today.

tlycken:
> All package PRs are failing this way today.

:crying_cat_face:

timholy:
Merge anyway? (I can't, it won't let me.)

tlycken:
I assume this is since JuliaLang/julia#11196 merged, and that it'll keep being broken until the regressions are fixed. I also assume this might take a while..:

I'll disable the build on nightly for now; after that, restarting the build should make it pass and allow merging (I'm using Github's semi-new "protect branch" feature on master, to prevent merging before checks pass, but it probably doesn't make sense to require that everything passes on a very unstable Julia master...).

tlycken:
See #75.

tlycken:
Rebasing this onto current master should fix the build (by not trying on 0.5...).

PR #82, opened 2015-10-17
Prefiltering: create Arrays, don't use similar

timholy:
Since a `BSplineInterpolation` requires an `Array` for `coefs`, it seems sensible to create an `Array` directly.

This fixes a bug for Images, for which `similar` creates an Image, but `convert(Array, img)` might execute a transpose depending on `img`'s storage order. (`convert` gets called implicilty [here](https://github.com/tlycken/Interpolations.jl/blob/544bf16665bf8f6ab06cf373dab86954bf512b0b/src/b-splines/b-splines.jl#L29).)


tlycken:
Yeah, I guess this was coded under the assumption that `similar` would actually create an `Array` (which, I guess, was true once, but can of course never be guaranteed for types outside of Base).

Good catch!

PR #83, opened 2015-10-21
remove cases of `symbol(string(`

spencerlyon2:
Minor change, but there is no need to create a string, just to turn it into a symbol


tlycken:
Thanks!

PR #86, opened 2015-10-23
Allow AbstractArrays for BSplineInterpolation

timholy:
I have a multi-terabyte file that corresponds to a single array (image data). I'd like to be able to perform interpolation on it and share the results with multiple workers on the same machine. Consequently, it's very desirable to allow one to construct a `BSplineInterpolation` from a `SharedArray`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Interpolations
 
 [![Build Status](https://travis-ci.org/tlycken/Interpolations.jl.svg?branch=master)](https://travis-ci.org/tlycken/Interpolations.jl)
+[![PkgEval Status](http://pkg.julialang.org/badges/Interpolations_0.4.svg)](http://pkg.julialang.org/?pkg=Interpolations)
 
 This package implements a variety of interpolation schemes for the
 Julia langauge.  It has the goals of ease-of-use, broad algorithmic


tlycken:
I'm all for making this possible, but I'm not sure tacking on another type argument is the best way to implement it - I think it adds unnecessary complexity. 

Given the argument A you added, we should be able to get rid of TCoefs, since the previous assumption was A=Array{N, TCoefs}, which allows for more functionality with the same number of parameters as today. Or am I missing something?

timholy:
No, you're right, that's just a bigger change. Given that you're enthusiastic, I'll do that. Do you want `A` in the current spot of `TCoefs`, or do you want it last? For the Interpolations code, I suspect there's some advantage in having it last, because we'd rarely dispatch on it.

tlycken:
Conceptually, I think it makes sense to have it early (i.e. in place of `TCoefs`), although I see the point about not dispatching on it. But I think [indexing.jl](https://github.com/tlycken/Interpolations.jl/blob/e250fced8cf3fd56168d5a345a028570c522908b/src/b-splines/indexing.jl) is the only place where we actually dispatch on the type parameters of `BSplineInterpolation`, and there we include all of them anyway (in other places, we dispatch on subtypes of `DimSpec{T<:InterpolationType}`...).

I'd actually vote for keeping the name `TCoefs` in the place were it is, but change its meaning - it's more descriptive than `A`. That would just entail changing the type definition to 

```
immutable BSplineInterpolation{T,N,TCoefs<:Array,...}
    coefs::TCoefs
end
```

and changing most (all?) instances of `TCoefs` to `eltype(TCoefs)` in other parts of the code.

tlycken:
Tim, do you have any work on this that you haven't pushed, or has this been bumped down on your (long, I don't doubt) priority list? If you want, I could take a stab at refactoring this PR to use just one type argument, so you don't have to worry about it :)

timholy:
Your timing is impecable. Some git snafu happened, though, I'll try a fresh PR.

PR #87, opened 2015-10-31
Decouple extrapolation bounds from element type

mbauman:
Extrapolation bounds for gridded interpolations are effectively indices and not coupled with the array eltype.  So this changes it to use integers (for `OnGrid()`) and ~~SimpleRatios~~ Float64 (for `OnCell()`) directly, instead of trying to convert the values to the eltype.  This is important because some array values may not support conversion from integers or floats.


tlycken:
LGTM.

@timholy, you have a better eye for potential type instabilities than me - do you see this causing problems anywhere?

mbauman:
I've added tests and fixed an issue that they uncovered with the OnCell implementation.  `@inferred` passes on getindex, so I think this should be good as far as type stability goes.

timholy:
LGTM. Thanks, Matt.

PR #88, opened 2015-11-10
Enable gradient method for Gridded(Constant()) and Gridded(Linear())

spencerlyon2:
This was mostly an exercise to confirm that I understand how all the meta-programming fits together.

One thing I don't like is that when I tested the gradient on a non 1-spaced grid I had to raise the multiplicative factor on the "equality tolerance" from 0.1 to 0.5. Does this bother anyone else or suggest I have made a mistake somewhere?

### DIFF
--- a/src/b-splines/quadratic.jl
+++ b/src/b-splines/quadratic.jl
@@ -183,4 +183,25 @@ function prefiltering_system{T,TCoefs,GT<:GridType}(::Type{T}, ::Type{TCoefs}, n
     Woodbury(lufact!(Tridiagonal(dl, d, du), Val{false}), rowspec, valspec, colspec), zeros(TCoefs, n)
 end
 
+
+"""
+    prefiltering_system{T,TCoefs,GT<:GridType,D<:Degree}m(::T, ::Type{TCoefs}, n::Int, ::Type{D}, ::Type{GT})
+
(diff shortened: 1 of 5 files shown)


spencerlyon2:
Oh oops, looks like I left a docstring change in this branch. If you are ok with that then I'll leave it, otherwise I can cherry-pick the gradient stuff and push to this pr branch

tlycken:
Nah, I'm not going to be that nitpicky.

timholy:
@spencerlyon2, sorry not to notice this earlier.

> One thing I don't like is that when I tested the gradient on a non 1-spaced grid I had to raise the multiplicative factor on the "equality tolerance" from 0.1 to 0.5. Does this bother anyone else or suggest I have made a mistake somewhere?

For writing tests, my personal preferred style is to pick test functions for which interpolation is exact, so that there is no question about the source of any discrepancies. An example is shown [here](https://github.com/tlycken/Interpolations.jl/blob/37fee94f74764f8e02dc33c4cc6639ddcb29647f/test/gradient.jl#L68-L107), where for testing quadratic interpolation I chose an "underlying" function that was, in fact, quadratic.

For example, for `Gridded{Linear}` you could do something like
```
x = sort(rand(5))
y = 0.2*x
```
and then test that the gradient is approximately 0.2 (with high precision).

PR #89, opened 2015-11-12
Allow AbstractArrays

timholy:
Redo of #86


tlycken:
This looks good to me. Thanks for taking the time!

Issue #55, opened 2015-08-06
Add tests for vector-indexing


timholy:
ref #54

tlycken:
Referencing #24 here too; it should be closed when we fix this.

tlycken:
@timholy I took the liberty of fleshing out the body of this issue a little :)

timholy:
Very good, thanks!

PR #90, opened 2015-11-13
WIP on cubic

spencerlyon2:
This is an early version of enabling cubic splines for the uniform case (non-gridded).

Features that need to be implemented:

- [x] Other boundary conditions
- [x] OnGrid vs OnCell distinction
- [ ] Gradient

Other things that need to be done

- [x] tests
- [x] Fix bug in pre-filtering -- right now `BSpline(Cubic(Line()))` is not interpolating. 
- [x] docs
- [ ] Add cubic to  benchmarks
- [x] Did I mention tests?


spencerlyon2:
Ok I think this might be ready for the early stages of review.

I have attempted to implement `Line`, `Free`, `Periodic`, and `Flat` for both `OnCell` and `OnGrid` specifications. I am not 100% confident that I did everything correctly (especially in how I derived the difference between `OnCell` and `OnGrid`), but early tests pass and suggest that what I did wasn't too crazy.

I included a mathematica notebook with all my derivations in a file `doc/Interpolations_algebra.nb`. If you can't open this (perhaps because you don't have mathematica) let me know and I'll export a different format.

src/b-splines/cubic.jl
```diff
@@ -0,0 +1,241 @@
...
+end
+
+# TODO: implement versions of `define_indices_d` for different Bc's if necessary
+
+padding{BC<:Flag}(::Type{BSpline{Cubic{BC}}}) = Val{1}()
+padding(::Type{BSpline{Cubic{Periodic}}}) = Val{0}()
+
+"""
+In this function we assume that `fx_d = x-ix_d` and we produce `cX_d` for
+`X ⋹ {m, _, p, pp}` such tha
```
timholy: This is so nice, I'm reduced to pointing out typos ("that")

timholy:
Gotta quit for the night, but this is looking very promising! Nice work.

src/b-splines/cubic.jl
```diff
@@ -0,0 +1,241 @@
...
+    d[1] = d[end] = -one(T)  # set first and last diagonal to -1
+    du[1] = dl[end] = zero(T)  # set just off diagonal to 0
+
+    # Now Woodbury correction to set `[1, 3], [n, n-2] ==> 1`
+    specs = _build_woodbury_specs(T, n, (1, 3, one(T)), (n, n-2, one(T)))
+
+    Woodbury(lufact!(Tridiagonal(dl, d, du), Val{false}), specs...), zeros(TC, n)
+end
+
+# TODO: check to see if above method is the same as Quadratic
```
timholy: Seems like it might be.

src/b-splines/cubic.jl
```diff
@@ -0,0 +1,241 @@
...
+                                  (n, n-3, one(T))
+                                  )
+
+    Woodbury(lufact!(Tridiagonal(dl, d, du), Val{false}), specs...), zeros(TC, n)
+end
+
+# NOTE: Line, OnGrid is implemented in quadratic.jl
+
+"""
+`Line` `OnCell` amounts to setting `y_0'(x) = 0` at `x = -1/2`. Applying this
```
timholy: `y_0''`?

spencerlyon2: yep, definitely a typo. However, the condition on the coefficients below is correct (it does use `y''`)

timholy:
I haven't checked the mathematics, but the rest looks great!

One general comment is that it looks like our tests could check the boundary conditions more stringently. For example, that for `Flat` the `gradient` is zero at the specified location. But this isn't specific to this PR.

src/b-splines/cubic.jl
spencerlyon2: Haha glad you like it! Will fix the typo

src/b-splines/cubic.jl
spencerlyon2: Oh I remember why I didn't just use the one from Quadratic...

I haven't implemented `Reflect` yet and wasn't sure if `(Flat, OnGrid)` and `(Reflect, OnGrid)` will be the same for Cubic as it is for quadratic

spencerlyon2:
I definitely agree about the boundary conditions. Testing for that would be great.

spencerlyon2:
So I just did another look through the code and have a few questions:

- Am I handling the `pad` right in `define_indices_d`? I did what made sense, but could be wrong
- I'm not totally confident that I did `Periodic` correctly. I followed what was done for quadratic, but didn't totally understand it
- I am using the `_build_woodbury_specs` helper function to add off tridiagonal elements, but am happy to just build the `___spec` object directly as we have done elsewhere.

timholy:
> Am I handling the pad right in `define_indices_d`? I did what made sense, but could be wrong

I haven't worked it through carefully, but I suspect you need to define `padding{BC<:Flag}(::Type{BSpline{Cubic{BC}}}) = Val{2}()` and bump the offset in `define_indices_d` from both sides. The padding has to be symmetric.

> I'm not totally confident that I did `Periodic` correctly. I followed what was done for quadratic, but didn't totally understand it

LGTM. The idea basically is that if your base point is 1, the "previous" coefficient is found at position `n` due to the periodic boundary conditions.

> I am using the `_build_woodbury_specs` helper function

Seems fine to me.

I'll give @tlycken a chance to review, too.

test/b-splines/cubic.jl
```diff
@@ -0,0 +1,48 @@
+# module CubicTests
+
+using Base.Test
+using Interpolations
+
+for (constructor, copier) in ((interpolate, identity), (interpolate!, copy))
+    f(x) = sin((x-3)*2pi/9 - 1)
```
timholy: IMO it would be great to use a cubic polynomial for testing, too, so you can be sure your values are exact. In the LaTeX docs, see the section about `InPlaceQ` to be sure the boundary conditions aren't destroying the exact equality.

tlycken:
I'd love to take a closer look at this, but I'm abroad atm (with a ridiculous network speed - "up to 256kbit/s"...!) so it will probably take a while before I get the chance. Give me a week or so :)

timholy:
Bump.

src/b-splines/b-splines.jl
```diff
@@ -73,9 +74,16 @@ function interpolate!{IT<:DimSpec{BSpline},GT<:DimSpec{GridType}}(A::AbstractArr
     interpolate!(tweight(A), A, it, gt)
 end
 
+immutable Cubic{BC<:Flag} <: Degree{3} end
+Cubic{BC<:Flag}(::BC) = Cubic{BC}()
+
+immutable Quadratic{BC<:Flag} <: Degree{2} end
+Quadratic{BC<:Flag}(::BC) = Quadratic{BC}()
```
tlycken: I'm not sure why these moved here, rather than staying in `quadratic.jl` and `cubic.jl`. I have no strong opinion on the matter, but *if* they move, `Linear` and `Constant` should move here too.

src/b-splines/cubic.jl
```diff
@@ -0,0 +1,235 @@
...
+
+For future reference, this expands out to the following polynomial:
+
+    y_i(x) = 1/6 cm (1+i-x)^3 + c (2/3 - (x-i)^2 + 1/2 (x-i)^3) +
+             cp (2/3 - (1+i-x)^2 + 1/2 (1+i-x)^3) + 1/6 cpp (x-i)^3
+
+When we derive boundary conditions we will use derivatives `y_0'(x)` and
+`y_0''(x)`
+"""
+Cubic
```
tlycken: I'm definitely a fan of moving documentation this way - it keeps it with the implementation and makes it easy to access. Nicely done!

src/b-splines/cubic.jl
```diff
@@ -0,0 +1,235 @@
...
+
+function define_indices_d{BC}(::Type{BSpline{Cubic{BC}}}, d, pad)
+    symix, symixm, symixp = symbol("ix_",d), symbol("ixm_",d), symbol("ixp_",d)
+    symixpp, symx, symfx = symbol("ixpp_",d), symbol("x_",d), symbol("fx_",d)
+    quote
+        # ensure that all three ix_d, ixm_d, and ixp_d are in-bounds no matter
+        # the value of pad
+        # TODO: Not sure if I'm handling the pad correctly, but relative to
+        #       quadratic case I did move upper clamp from `_ - 1` to `_ - 2`
+        #       to make room for the `pp` values
```
tlycken: This seems correct to me.

src/b-splines/cubic.jl
```diff
@@ -0,0 +1,235 @@
...
+    quote
+        $symix = clamp(round(Int, real($symx)), 1, size(itp,$d))
+        $symfx = $symx - $symix
+        $symixm = mod1($symix - 1, size(itp,$d))
+        $symixp = mod1($symix + 1, size(itp,$d))
+        $symixpp = mod1($symix + 2, size(itp,$d))
+    end
+end
+
+# TODO: implement versions of `define_indices_d` for different Bc's if necessary
```
tlycken: This shouldn't be necessary; indices are the same for all BC's except periodic, which you already handle.

src/b-splines/cubic.jl
```diff
@@ -0,0 +1,235 @@
...
+        $sym_1m_fx_cub = cub(1-$symfx)
+        $symm = SimpleRatio(1,6)*$sym_1m_fx_cub
+        $sym  = SimpleRatio(2,3) - sqr($symfx) + SimpleRatio(1,2)*$symfx_cub
+        $symp = SimpleRatio(2,3) - sqr(1-$symfx) + SimpleRatio(1,2)*$sym_1m_fx_cub
+        $sympp = SimpleRatio(1,6)*$symfx_cub
+    end
+end
+
+# This assumes integral values ixm_d, ix_d, and ixp_d,
+# coefficients cm_d, c_d, and cp_d, and an array itp.coefs
```
tlycken: This comment seems like it's copied from the comment above the corresponding quadratic method. Should probably be either updated (and converted to a docstring?) or removed entirely.

src/b-splines/cubic.jl
tlycken: The mathematical root would of course be the same - `∇f(x_0) = 0` for `x_0` at the boundary - however that doesn't necessarily mean the same thing for the coefficient equations. I *think* we need to handle cubic separately.

If you work out the equations and find that they are indeed the same as for the quadratic case, I would prefer a method here that forwards to the quadratic (to make the re-use explicit in this file) rather than just updating the parameter types on the quadratic method.

src/b-splines/cubic.jl
```diff
@@ -0,0 +1,235 @@
...
+```julia
+out = zeros(n, n)
+
+for (i, j, v) in args
+    out[i, j] = v
+end
+```
+
+"""
+function _build_woodbury_specs{T}(::Type{T}, n::Int, args::Tuple{Int, Int, Any}...)
```
tlycken: I do agree that this interface is nicer and makes it easier to see what's going on, but it feels a little out-of-place. Perhaps it would fit better in a PR to [WoodburyMatrices.jl](https://github.com/timholy/WoodburyMatrices.jl)?

In the meantime, moving this function to the bottom of the file gets it out of the way from the interpolation-related code.

src/b-splines/cubic.jl
```diff
@@ -0,0 +1,235 @@
...
+                                  (1, n, SimpleRatio(1, 6)),
+                                  (n, 1, SimpleRatio(1, 6))
+                                  )
+
+    Woodbury(lufact!(Tridiagonal(dl, d, du), Val{false}), specs...), zeros(TC, n)
+end
+
+
+#=
+NOTE: the following are the same as Quadratic and are implemented there:
```
tlycken: I would prefer if these were actually explicitly implemented here, but just forwarding to the implementations in quadratic.jl, e.g.

```
prefiltering_system{T,TC}(::Type{T}, ::Type{TC}, n::Int, ::Type{Cubic{Line}}, ::Type{OnGrid}) =
    prefiltering_system(T, TC, n, Quadratic{Line}, OnGrid)
```

src/b-splines/quadratic.jl
```diff
@@ -129,42 +126,46 @@ function prefiltering_system{T,TC,BC<:Union{Flat,Reflect}}(::Type{T}, ::Type{TC}
...
+        colspec[1,3] = colspec[2,n-2] = 1
+        valspec = zeros(T,2,2)
+        # [1,3]         [n,n-2]
+        valspec[1,1] = valspec[2,2] = 1
+
+        Woodbury(lufact!(Tridiagonal(dl, d, du), Val{false}), rowspec, valspec, colspec), zeros(TC, n)
+    end
 
-    Woodbury(lufact!(Tridiagonal(dl, d, du), Val{false}), rowspec, valspec, colspec), zeros(TC, n)
+    @eval function prefiltering_system{T,TC,GT<:GridType}(::Type{T}, ::Type{TC}, n::Int, ::Type{$BST{Free}}, ::Type{GT})
```
tlycken: Even if we don't do this by explcit forwarding, there's no need for `@eval` here. Instead, you can do

```
function prefiltering_system{T,TC,BST<:Union{Quadratic{Free},Cubic{Free}},GT<:GridType}(..., ::Type{BST}, ...)
```

to get the same methods defined.

tlycken:
Sorry I took so long on this. I've added a bunch of line comments, I hope you don't find them too nitpicky...

In general, this is great work! :+1: 

I have two main objections to the approach you've taken in some cases, mentioned in the comments too:

* I would prefer if all the method definitons related to `Cubic` were explicitly defined in `cubic.jl`, to make it easier to read and document the code.
* I don't think improvements on the interface for creating Woodbury matrices really belong in this package; I'd much prefer if they were submitted to WoodburyMatrices.jl instead (I like them, though, so don't take this as criticism to the idea itself!). If you don't want to keep this PR waiting for another PR there, I think we should update our woodbury matrix contructions at a later date.

Regarding the yet missing features:

* We should implement both `gradient` and [`Hessian`](https://en.wikipedia.org/wiki/Hessian_matrix) for cubic interpolations (continuous Hessians is one of the major arguments for using a cubic spline rather than a quadratic one). If this feels daunting at the moment, I'm fine with doing that as a separate PR (which doesn't *have* to be written by you :) ). The basic principle I followed when I did it for Grid.jl was to just take derivatives of the coefficient polynomials, and the rest follows from there.

* Including cubic splines in the benchmarks is definitely something we want to do, but it could also be done separately (I think Tim knows the inner workings of them better than any of us).

src/b-splines/b-splines.jl
spencerlyon2: They moved here so that I could use `Cubic` and `Quadratic` when re-using some boundary condition implementations across the two degrees. 

If I didn't do this and in `quadratic.jl` I referenced `Cubic`, then I would get an undefined error.

However, this won't be necessary if we re-use implementations from `quadratic.jl` by explicitly forwarding them within `cubic.jl` as suggested above

spencerlyon2:
Thanks for all the comments!

I will go through and address the line comments. I think making the Woodbury spec PR shouldn't take too long to push through (hint @timholy  :wink:)

I think that we can add cubic to benchmarks in a different PR.

Also, given my current time constraints I think I'd vote for putting derivatives in a different PR also, is that ok?

tlycken:
I'm totally fine with derivatives in a separate PR. This is new functionality - it's not broken just because it can be extended :)

lstagner:
Does this PR support [Monotonic Cubic Splines](https://en.wikipedia.org/wiki/Monotone_cubic_interpolation)?

spencerlyon2:
Hi @lstagner this does not implement monotnoic cubic splines, sorry. I'd love to see them too.

@timholy and @tlycken I _think_ I have resolved most of the comments from above. Some things that still need cleaning up:

- [ ] `InPlace` and `InPlaceQ`. I haven't even tried to touch them.
- [ ] Could one of you take a look at how I handle padding? Tim made a comment above and I didn't change the code at all after that. 
- [ ] I've moved the `_build_woodbury_specs` to the bottom of this file. @timholy would you be open to a PR over at WoodburyMatrices?
- [ ] As suggested by @timholy, more stringent tests of the boundary conditions would be great; although this applies to both Quadratic and Cubic and could perhaps be addressed in another PR? Happy to follow your lead on this one

Please let me know what else needs to be done to get this merged. I would suggest that I squash some of the commits before then, so please let me know before pressing that big green button ;)

src/b-splines/quadratic.jl
```diff
@@ -1,6 +1,7 @@
 immutable Quadratic{BC<:Flag} <: Degree{2} end
 Quadratic{BC<:Flag}(::BC) = Quadratic{BC}()
 
+
```
tlycken: This whitespace change is superfluous.

src/b-splines/quadratic.jl
```diff
@@ -129,6 +130,8 @@ function prefiltering_system{T,TC,BC<:Union{Flat,Reflect}}(::Type{T}, ::Type{TC}
     Woodbury(lufact!(Tridiagonal(dl, d, du), Val{false}), rowspec, valspec, colspec), zeros(TC, n)
 end
 
+# quadratic and cubic share the same woodbury correction for `Line` type. No need
+# to re-implement
```
tlycken: What does this comment add here? This idea is expressed in the relevant docstrings for cubic interpolation, so I think this comment can be removed entirely.

src/b-splines/cubic.jl
```diff
@@ -0,0 +1,248 @@
...
+        $symix = clamp(round(Int, real($symx)), 1, size(itp,$d))
+        $symfx = $symx - $symix
+        $symixm = mod1($symix - 1, size(itp,$d))
+        $symixp = mod1($symix + 1, size(itp,$d))
+        $symixpp = mod1($symix + 2, size(itp,$d))
+    end
+end
+
+padding{BC<:Flag}(::Type{BSpline{Cubic{BC}}}) = Val{1}()
+padding(::Type{BSpline{Cubic{Periodic}}}) = Val{0}()
```
tlycken: This looks good to me :+1:

The idea with padding is to give room for the "ghost cells" needed to implement the boundary conditions. For quadratic splines, where the i:th interpolation function is based on the coefficients at i-1, i and i+1, and yields values from i-1/2 to i+1/2. Thus, one needs to pad with 1 at the bottom (since at i==1 we also need the 0:th coefficient) and 1 at the top (since at i==N we also need the N+1:th).

Similarly, for cubic splines we use the i:th function to interpolate between i and i+1 (note that this is *not* centered on i, like the quadratic case), and uses four points from i-1 to i+2. Thus, when i==1 we need the 0:th coefficient, and at i==N we can use the N-1:th function and get by with padding only to N+1 here, i.e. a padding of 1 on each end here too.

It strikes me as I'm writing this that this only applies for `OnGrid` cubic interpolation; for `OnCell`, we're either continuing the interpolant on [1,2) for an extra half-cell below (this seems to be what we're doing now) or we need to add another padding coefficient to be able to interpolate the half-cell with the interpolant valid on [0,1). I think it's fine to do it this way, but we might want to document it explicitly.

src/b-splines/cubic.jl
```diff
@@ -0,0 +1,248 @@
...
+        $symix = clamp(round(Int, real($symx)), 2-$pad, size(itp,$d)+$pad-2)
+        $symfx = $symx - $symix
+        $symix += $pad # padding for oob coefficient
+        $symixm = $symix - 1
+        $symixp = $symix + 1
+        $symixpp = $symixp + 1
+    end
+end
+
+# TODO: not sure if I am doing any of this Periodic stuff correctly
```
tlycken: This seems good to me; this comment should be removed.

tlycken:
@spencerlyon2: I went over this again; most of it looks really good now! :)

There's a couple of minor edits (removing obsolete comments and superfluous whitespace changes) that might be nice to get done before we merge this.

The bullets you mentioned

* [ ] `InPlace` and `InPlaceQ`: they are, AFAIK, specifically designed to facilitate testing, so I think we can leave these untouched unless you want to address the testing too (see below)
* [x] Padding looks good to me (although do read my line comment about it)
* [x] `_build_woodbury_specs` LGTM too - we'll update the usage if/when WoodburyMatrices.jl updates
* [ ] More stringent testing. It would be nice, but it can be done separately if you want. (But if so, please file an issue so we keep track of the fact that it's lacking.)

tlycken:
Also, I haven't been able to view the notebook you added anywhere (nbviewer just gives me a 404 page). Could you post a link to somewhere I can view the notebook without having to download and install IJulia etc (I've been struggling with getting it to work under 0.4 on Windows, and I don't really have the time to fix that kind of problem at the moment...)

timholy:
Regarding Woodbury, naturally I'm happy to accept a PR. I've given you both commit access.

spencerlyon2:
@tlycken thanks for reviewing and getting this code up to the quality bar!

The notebook I added is a _Mathematica_ notebook, not a Jupyter one. Would you like me to export a pdf or other so you can have a look?

Regarding `InPlace` and `InPlaceQ`: I haven't added them at all to cubic. Should we add them before merging?

I agree that better testing is needed, but I prefer that we do that later if you are ok with it. I just don't think I have the extra cycles to work on that right now.

tlycken:
> The notebook I added is a Mathematica notebook, not a Jupyter one. Would you like me to export a pdf or other so you can have a look?

Ah, that explains it. I think it would be valuable to export a pdf (or some other widely readable format) and check that in as well, much as has been done with the LaTeX documentation.

> Regarding InPlace and InPlaceQ: I haven't added them at all to cubic. Should we add them before merging?
>
> I agree that better testing is needed, but I prefer that we do that later if you are ok with it. I just don't think I have the extra cycles to work on that right now.

That's fine by me; I think this is merge-able without the testing. We'll file an issue to keep track of the fact that it needs to be done (or update #55 to cover this too).

So, todo before we can merge this: check in a readable export of the Mathematica notebook, and squash the commits.

Great work, I really appreciate the effort you've put into this!

spencerlyon2:
OK great! I have exported to pdf and squashed.

Just a note, I think the comment about better testing of boundary conditions applies to quadratic also. The cubic tests are pretty much a copy/paste of the existing quadratic tests. We should probably open an issue dedicated to better testing of boundary conditions for both quadratic and cubic.

tlycken:
Thanks for your great work - and your patience with my slow reviews - to put this together! I really do appreciate the effort.

I'm not sure I really like introducing yet another format of documentation (your Mathematica notebook looks quite like the LaTeX documentation for the lower degrees) but I don't think that is enough of a reason to hold up this PR any longer.

As I see it, when I merge this in a moment, I have three issues to file and/or update:

* [ ] Implementation of `gradient` for cubic b-splines, and `hessian` for all b-splines (just 0 for all lower degree)
* [ ] Improved testing, both for cubic splines and for quadratic. This include testing with polynomials that can be interpolated exactly, as well as testing the various boundary conditions more explicitly.
* [ ] Including cubic b-splines in the benchmarks
* [ ] Centralizing the documentation of the mathematics behind this library to some location that supports plain-text

spencerlyon2:
I totally agree with you regarding adding the mathematica notebook as a form of documentation. I simply used mathematica to do the derivations so I didn't have to worry about algebra errors creeping into my work. Most (all?) of the content from that notebook is included in the docstrings in cubic.jl anyway, so it might not be necessary to have another form of docs for cubic.

If you want I could ask mathematica to export to tex and just append that output to the existing latex documentation.  What do you think about that?

PR #97, opened 2016-01-12
Remove the call to `real` when rounding to integer indexes

timholy:
The deprecation of `real` for DualNumbers broke the tests that used `@inferrable`. More long-term, I think it's better to just perform the operation we want without an extra call to `real`. This places the onus on the relevant number types to support `round(Int, x)`, but the two major use-cases for automatic differentiation (ForwardDiff and DualNumbers) now support this (see https://github.com/JuliaDiff/DualNumbers.jl/pull/33 and https://github.com/JuliaDiff/ForwardDiff.jl/pull/84).

CC @mlubin, since I pestered him incessantly and he should know that I'm making good use of his exemplary diligence as a maintainer!

### DIFF
--- a/src/b-splines/constant.jl
+++ b/src/b-splines/constant.jl
@@ -2,7 +2,7 @@ immutable Constant <: Degree{0} end
 
 function define_indices_d(::Type{BSpline{Constant}}, d, pad)
     symix, symx = symbol("ix_",d), symbol("x_",d)
-    :($symix = clamp(round(Int, real($symx)), 1, size(itp, $d)))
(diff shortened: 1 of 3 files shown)


tlycken:
Sounds like a good enough approach that I've already merged it :)

timholy:
Since this fixes deprecation warnings, and deprecation warnings kill performance, I suspect we should tag a new release. Note that this would also release @spencerlyon2's work on cubic, assuming everyone is fine with that. (I am!)

tlycken:
Yes, that is a good idea. In #90 we left a few things off, so the implementation of cubic splines is not yet quite as feature-rich as for lower orders, but I'm fine with that.

PR #98, opened 2016-01-12
remove trailing - from julia version in REQUIRE

tkelman:
unless you have a need to still support pre-releases of julia 0.4

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.4-
+julia 0.4
 
 WoodburyMatrices 0.1
 Ratios


tlycken:
Thanks.

The Travis failure is unrelatedm; @timholy, is this test failure what #97 is intended to address?

tkelman:
most likely, ref http://pkg.julialang.org/detail/Interpolations.html

tlycken:
Yup, noticed that the other day too (I wonder why PkgEval didn't file an issue here about it...). I've merged #97 now, where tests pass.

This change really shouln't be a problem in any way, but I feel a little uneasy merging things with failing tests. Do you mind rebasing? If you have more important things to do, I'll rebase and check myself when I get home from work.

tkelman:
I did this from the web interface. I can ssh into something that's not a phone, clone and rebase but I think if you click restart on travis it would fetch a new merge commit to latest master.

timholy:
Thanks, @tkelman.

tlycken:
Ah, I see. Thanks!

PR #101, opened 2016-01-14
Extrapolation of gridded

### LINKED ISSUE #100: extrapolation of Gridded interpolation is wrong
```julia
itp = interpolate(([0;10],),[100;110], Gridded(Linear()))
```
By the way, what is  REPL response [101.0 102.0] ?
```julia
etp = extrapolate(itp, Flat())
itp[5] # 105.0 OK
etp[5] # 102.0 Buggy
```
5 is considered as out of [1,2] interval (as documented but not as expected by common users)

tlycken:
Implements `lbound` and `ubound` for `GriddedInterpolation`s, to enable cooperation with the `extrapolate` functionality.

Fixes #100

### DIFF
--- a/src/gridded/gridded.jl
+++ b/src/gridded/gridded.jl
@@ -60,6 +60,9 @@ function interpolate!{Tel,N,IT<:DimSpec{Gridded}}(knots::NTuple{N,GridIndex}, A:
     interpolate!(tweight(A), tcoef(A), knots, A, it)
 end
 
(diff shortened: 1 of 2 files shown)


tlycken:
Fun fact: this PR was produced by testing a few things on a JuliaBox instance, and then creating a branch, editing the two code files and creating the PR all through the web interface at Github.com. I don't have Julia installed on the computer from which I did this, yet I feel fairly confident that I did it right (and that the tests will fail if I didn't) :)

timholy:
Thanks!

Issue #46, opened 2015-07-18
Documentation overhaul


tlycken:
This is to discuss changes to, and completion of, documentation for this package. Major questions are:

* How do we want to organize the documentation? Formats, locations etc...
* What should be documented (more than just by writing clear code)? Devdocs? Implementation details?
* On what level of detail do we want to document things?

Some discussion started in #45 already, where the following points [were raised](https://github.com/tlycken/Interpolations.jl/pull/45#issuecomment-122535090):

> In a sense the README could be a summary of what's available, now that
>
>     julia> using Interpolations
>     julia> ?Interpolations
>
> dumps the README contents. So I've begun to think that it might be good to keep the README as an "API reminder" for people who already understand the package, and the IJulia notebook as an introduction to the package.

I agree that it's probably a good idea to make the readme minimalistic. I'd also like to start documenting most other things in a `help>`-compatible way, so we might want to move off from the IJulia notebooks alltogether. And I feel like I have to strip down the LaTeX stuff to something with minimal dependencies, so anyone can at least compile it, but it is a very handy way to document the mathematical background of any methods we implement...

tlycken:
This was raised again in discussion in #90. I'd like to find a better place to keep documentation of math that's too lengthy and/or not relevant for docstrings.

Requirements:

* Readable and editable in any text editor
* It should be possible to render the equations to something nice-looking
* It should be relatively easy to have the rendered documentation shown on the web somewhere (could be inside the repo if we choose markdown - could be something like readthedocs.org...) and update it as part of the bulid/test process.

Nice-to-haves:

* The notation to render mathematics should, preferrably, be easy enough on the eyes to be able to follow even if looking at the plaintext version.

Any suggestions are welcome.

tlycken:
I asked a question on julia-users to get some input for tools and platforms: https://groups.google.com/forum/?nomobile=true#!topic/julia-users/mRk0igz_uHQ

argelouski:
I have seen in the docs some obscure infromation about how to select the behaviour outside the domain, i.e, decide wether or not to extrapolate. I haven't figured how to use it, however, i can see that ExtrapNaN or ExtrapError are no longer valid as arguments to extrapolate. I think that part of the docs could be improved.

tlycken:
That would be most appreciated!

tlycken:
[...]

PR #103, opened 2016-01-18
Improve docs

tlycken:
This builds on #99, so once that's merged, this should only list the last two commits.

Replicated much of @spencerlyon2's wonderful docstrings from `Cubic` for lower-order b-splines as well as fixed some things in the originals (mostly to make it easier to tell them all apart when listing e.g. `help?> Interpolations.prefiltering_system`).

I also found basically that all the information that was in the LaTeX document I wrote long ago is now in docstrings instead, so I removed it. (It's still in the history, if one wants to get it back.)

I haven't touched documentation for any non-B-spline interpolation code.


tlycken:
ref #46

src/b-splines/cubic.jl
```diff
@@ -173,75 +258,62 @@ function prefiltering_system{T,TC}(::Type{T}, ::Type{TC}, n::Int,
...
-    1 cm -2 c + 1 cp = 0 = 0
-
-This is the same system as `Quadratic{Line}`, `OnGrid` so we reuse the
-implementation
+    1 cm -2 c + 1 cp = 0
 """
-function prefiltering_system{T,TC,GT<:GridType}(::Type{T}, ::Type{TC}, n::Int,
-                                                ::Type{Cubic{Line}}, ::Type{GT})
-    prefiltering_system(T, TC, n, Quadratic{Line}, OnGrid)
+function prefiltering_system{T,TC}(::Type{T}, ::Type{TC}, n::Int,
```
spencerlyon2: Did we decide not to rely on the `Quadratic{Line}, OnGrid` implementation here?

spencerlyon2:
Excellent pull request! I love the docstrings. I think this will lower the bar to mostly metaprogramming stuff for people who want to understand how Interpolations.jl works by reading the package.

src/b-splines/cubic.jl
tlycken: Yes, I realized when documenting this that the systems aren't the same after all. `prefiltering_system` relies on `inner_system_diags`, which will yield different coefficients for quadratic and cubic (of course), and we lose that information if we just forward the call.

spencerlyon2: Oh man ok. Good catch. Sounds good, it's nice to have it below he description also

src/b-splines/constant.jl
```diff
@@ -1,20 +1,50 @@
 immutable Constant <: Degree{0} end
 
+"""
+Constant b-splines are *nearest-neighbor* interpolations, and effectively
+return `A[round(Int,x)]` when interpolating
+
+Also, although the implementation is slightly different in order to re-use
```
tlycken: This is a copy-paste mistake; this remark is about linear, not constant, B-splines.

src/b-splines/prefiltering.jl
```diff
@@ -78,3 +78,35 @@ function prefilter!{TWeights,TCoefs<:AbstractArray,IT<:Tuple{Vararg{Union{BSplin
 end
 
 prefiltering_system(::Any, ::Any, ::Any, ::Any, ::Any) = nothing, nothing
```
tlycken: @timholy is this method actually good for anything? It seems to me that it would be better to get MethodErrors on prefiltering_system than on whatever happens to use it...

src/b-splines/quadratic.jl
```diff
@@ -1,6 +1,33 @@
 immutable Quadratic{BC<:Flag} <: Degree{2} end
 Quadratic{BC<:Flag}(::BC) = Quadratic{BC}()
 
+"""
+Assuming uniform knots with spacing 1, the `i`th piece of quadratic spline
+implemented here is defined as follows:
+
+    y_i(x) = cm p(x-i) + c q(x) + cp(1-(x-i))
```
tlycken: Missing a p in the third term.

PR #104, opened 2016-01-18
WIP: Remove `_build_woodbury_specs`

tlycken:
This builds on #99, and only the last commit is unique to this PR.

I thought this would be uncontroversial, but as the test log will soon show, one gets an error from WoodburyMatrices stating

>ERROR: LoadError: LoadError: LoadError: The inverse of a sparse matrix can often be dense and can cause the computer to run out of memory. If you are sure you have enough memory, please convert your matrix to a dense matrix.
> in call at /home/tlycken/.julia/v0.4/WoodburyMatrices/src/WoodburyMatrices.jl:42
> in prefiltering_system at /home/tlycken/.julia/v0.4/Interpolations/src/b-splines/quadratic.jl:144

The line 144 referred to is the call to the `Woodbury` constructor below:

    dl,d,du = inner_system_diags(T,n,Quadratic{BC})
    d[1] = d[end] = -1
    du[1] = dl[end] = 0

    specs = WoodburyMatrices.sparse_factors(T, n,
                                  (1, 3, one(T)),
                                  (n, n-2, one(T))
                                 )

    Woodbury(lufact!(Tridiagonal(dl, d, du), Val{false}), specs...), zeros(TC, n) # error
and I don't see anything wrong with it.

@timholy, @spencerlyon2, can either of you see what's going on here?

Before this merges, we should

* [ ] fix the failing tests
* [x] update REQUIRE to require a version of WoodburyMatrices that includes `sparse_factors`


src/b-splines/cubic.jl
```diff
@@ -210,38 +271,12 @@ This is the same system as `Quadratic{Free}` so we reuse the implementation
...
-    out[i, j] = v
-end
-```
-
-"""
-function _build_woodbury_specs{T}(::Type{T}, n::Int, args::Tuple{Int, Int, Any}...)
-    m = length(args)
-    rowspec = spzeros(T, n, m)
-    colspec = spzeros(T, m, n)
-    valspec = zeros(T, m, m)
```
spencerlyon2: So this is where the problem is.

In `_build_woodbury_specs` we make `valspec` dense, but `WoodburyMatrices.sparse_factors` makes it sparse. 

This works without a problem:

```julia
julia> using WoodburyMatrices

julia> lupart = lufact!(Tridiagonal(ones(3), ones(4), ones(3)), Val{false});

julia> specs = WoodburyMatrices.sparse_factors(Float64, 4, (1, 3, 1.0), (4, 2, 1.0));

julia> Woodbury(lupart, specs...)
ERROR: The inverse of a sparse matrix can often be dense and can cause the computer to run out of memory. If you are sure you have enough memory, please convert your matrix to a dense matrix.
 in call at /Users/sglyon/.julia/v0.4/WoodburyMatrices/src/WoodburyMatrices.jl:41

julia> Woodbury(lupart, specs[1], full(specs[2]), specs[3])
4x4 WoodburyMatrices.Woodbury{Float64,Base.LinAlg.LU{Float64,Tridiagonal{Float64}},SparseMatrixCSC{Float64,Int64},SparseMatrixCSC{Float64,Int64},Array{Float64,2},Array{Float64,2}}:
 #undef  #undef  #undef  #undef
 #undef  #undef  #undef  #undef
 #undef  #undef  #undef  #undef
 #undef  #undef  #undef  #undef
```

@timholy what do you suggest we do here? Should we adjust `WoodburyMatrices.sparse_factors` to return a dense `valspec`?

timholy: I'd be fine with that option.

spencerlyon2: Done, if we tag a new release of WoodburyMatrices this will be taken care of

PR #109, opened 2016-04-08
Updated readme

spencerlyon2:
- Added notes about `Cubic`
- Fixed gridded example so it runs without errors
- Added note about Hessian

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ can be obtained as
 ```jl
 h = hessian(itp, x, y, ...)
 ```
-or, if you're evaluating the gradient repeatedly, a somewhat more
+or, if you're evaluating the hessian repeatedly, a somewhat more
 efficient option is
 ```jl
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,6 +63,18 @@ gradient!(g, itp, x, y, ...)
 ```
 where `g` is a pre-allocated vector.
 
+Some interpolation objects support computation of the hessian, which
+can be obtained as
+```jl
+h = hessian(itp, x, y, ...)
+```
+or, if you're evaluating the gradient repeatedly, a somewhat more
```
tlycken: I assume this is meant to read "Hessian" too? :)

tlycken:
Nice, thanks!

spencerlyon2:
Ahh yes good catch! 

Fixed

tlycken:
LGTM - thanks again for taking the time!

Issue #105, opened 2016-01-20
Monotonic interpolating splines


sebastien-villemot:
Thanks for this great package.

It would be nice to add monotonic splines. This is very useful in some situations, e.g. when optimizing over an interpolated function.

See for example [splinefun](http://www.inside-r.org/r-doc/stats/splinefun) from R for an implementation and some references.

tlycken:
Monotonic splines would be a great addition! However, we can't look at R's implementation, since it's GPL licenced. The references listed are these:

> Becker, R. A., Chambers, J. M. and Wilks, A. R. (1988) The New S Language. Wadsworth & Brooks/Cole.
>
> Dougherty, R. L., Edelman, A. and Hyman, J. M. (1989) Positivity-, monotonicity-, or convexity-preserving cubic and quintic Hermite interpolation. Mathematics of Computation 52, 471--494.
>
> Forsythe, G. E., Malcolm, M. A. and Moler, C. B. (1977) Computer Methods for Mathematical Computations. Wiley.
>
> Fritsch, F. N. and Carlson, R. E. (1980) Monotone piecewise cubic interpolation, SIAM Journal on Numerical Analysis 17, 238--246.
> 
> Hyman, J. M. (1983) Accurate monotonicity preserving cubic interpolation. SIAM J. Sci. Stat. Comput. 4, 645--654.

I don't have any of those available, but someone else might :) Without having read more than the titles, it seems to me that [this paper](http://epubs.siam.org/doi/abs/10.1137/0720057) and [this paper](http://epubs.siam.org/doi/abs/10.1137/0717021) might also be good starting points.

My focus at the moment is on wrapping up B-splines (see e.g. #99, and in particular [this comment](https://github.com/tlycken/Interpolations.jl/pull/99#issuecomment-172659204)) and I will unfortunately not have much time to spend on this library in a while, so I cannot promise a timeline for when this could be included.

Of course, if you'd like to give it a try yourself, a PR is certainly welcome!

sebastien-villemot:
Thanks for your feedback. Currently I am also quite busy, but I may have a look later.

nicknack23:
There is a nice Javascript implementation of monotonic cubic splines on [Wikipedia](https://en.wikipedia.org/wiki/Monotone_cubic_interpolation#Example_implementation), but licensing is not mentioned. The code was added by "Olathe" in 2013, perhaps he is still available and can release it under MIT?

timholy:
[...]

PR #110, opened 2016-04-21
WIP: Cubic Hermite Splines (and monotone splines)

cc7768:
Work in progress with @spencerlyon2. This PR (when it is finished) would address #105 

**Plan**: Cubic Hermite splines use information about both the value and derivative of the function. The plan is to allow for users to give both pieces of information if they would like, but if they only give values then we can construct an approximation of the derivative for them (using a specific approximation that preserves monotonicity).

**Progress**: A long way to go, but I thought I would open PR so no work got duplicated (and as a commitment device to myself). We have a (very) crude 1d implementation that seems to work for interpolating data.

**Next**: Our next step is that we need to plug this into the `Interpolations.jl` framework i.e. make sure we write the right indexing functions etc... Will keep people updated.

### DIFF
--- a/src/cubichermite/ch_ex.jl
+++ b/src/cubichermite/ch_ex.jl
@@ -1,0 +1,41 @@
+using Interpolations
+include("cubichermite.jl")
+
+
+# Use points 1, 2, 3, 4, 5
+x = collect(1:5)
+
+# Test with exponential function
+f(x) = log(x)
+fp(x) = 1.0 ./ x
+
+# Create arrays we need
+y = f(x)
+m = 1.0 ./ fp(x)
+
+# Create the cubic hermite interpolation coeffs
+# Note: m is the derivatives, so this has more info
+c = pre_solve(y, m)
+c_nd = pre_solve(y)
+
+# Create an interpolations cubic b-spline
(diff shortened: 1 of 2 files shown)


tlycken:
Nice!

Hermite splines are not inherently cubic, right? How difficult would it be to generalize the implementation to support (at least) lower orders too?

cc7768:
Hermite splines aren't inherently cubic, but that seems to be the main use case. I don't know how hard it would be to generalize the implementation:

* Linear: Not sure that implementing these would be worthwhile because they should be equivalent to linear b-splines
* Quadratic: Might be worth doing, though they seem to be used very infrequently. Relevant paper: [On Shape Preserving Quadratic Spline Interpolation](http://www.business.uzh.ch/professorships/qba/teaching/operationsResearch/lectures/CEF/SchumakerShapePreservingSplineSIAMJNA1983.pdf)

Will let you know how much effort implementing the quadratic version looks like it will be as I have time to work on this.

appleparan:
Why this one not merged at that time? It could be worth to merge and add Linear & quadratic later.

cc7768:
I haven't made time to finish this, so there isn't anything to merge.

Likely won't have a chance to do this before new year. Depends on how some other things progress.

PR #113, opened 2016-07-28
Fix errors/deprecations/warnings in julia 0.5

carlobaldassi:
~~NOTE: this depends on a related PR getting merged in AxisAlgorithms: https://github.com/timholy/AxisAlgorithms.jl/pull/2~~

The package has several errors and warnings and it's basically broken on julia 0.5.
I tried to fix them all and now on my machine all tests pass without any complaint both on 0.4 and 0.5.
I had to [do some very ugly things](https://github.com/tlycken/Interpolations.jl/commit/20558b754647c6cbf065ae2eb3ba18593cbaf9d2) though, which I think is unavoidable since in julia 0.5 ambiguities lead to errors rather than warnings, which means tests were not passing.

This also introduces a dependency on Compat due to the use of `MulFun` which has been deprecated (introducing another error here) and `super` which has become `supertype`.


tlycken:
Thanks for tackling this!

I've been afk for about a month, so I've completely missed the release of 0.5 with all that entails. I'll hopefully be able to give this PR some love later this week - if possible, I'd like to avoid the `@eval` stuff, and enough other small things have changed that I'd like to give all of it some full review.

tlycken:
OK, I've been able to give this a closer look now; I'll add a few line comments in the places where I think we can improve.

test/b-splines/linear.jl
```diff
@@ -28,14 +28,14 @@ for (constructor, copier) in ((interpolate, x->x), (interpolate!, copy))
 
     # 2D
     g2(y) = cos(y/6)
-    f(x,y) = g1(x)*g2(y)
```
tlycken: This should probably be resolved instead by moving the definitions of `f(x,y)` out of the loop. To do so, move `g2`, `ymax` and `A2``too - but all of these should probably have been outside of the loop to begin with...

src/b-splines/indexing.jl
```diff
@@ -98,11 +98,14 @@ end
...
 end
 
-@generated function gradient{T,N}(itp::AbstractInterpolation{T,N}, xs...)
-    n = count_interp_dims(itp, N)
-    Tg = promote_type(T, [x <: AbstractArray ? eltype(x) : x for x in xs]...)
-    xargs = [:(xs[$d]) for d in 1:length(xs)]
-    :(gradient!(Array($Tg,$n), itp, $(xargs...)))
+# @eval uglyness required for disambiguation with method in Base
+for R in [:Real, :Any]
+    @eval @generated function gradient{T,N}(itp::AbstractInterpolation{T,N}, xs::$R...)
```
tlycken: Using `@eval` here just to avoid code duplication seems to obscure the code more than it helps. It's probably better just to duplicate the method definition - once with `xs::Real...` and once with `xs...`. It's of course entirely possible to lift out the actual implementation into a private helper, i.e. 

```
@generated function gradient(itp::AbstractInterpolation, xs::Real...) = _gradient(itp, xs...)
@generated function gradient(itp::AbstractInterpolation, xs::Any...) = _gradient(itp, xs...)
@generated function _gradient{T,N}(itp::AbstractInterpolation{T,N}(itp::AbstractInterpolation{T,N}, xs...)
    #current implementation
end
```

src/scaling/scaling.jl
```diff
@@ -68,7 +68,12 @@ coordlookup(i::Bool, r::Range, x) = i ? coordlookup(r, x) : convert(typeof(coord
 basetype{T,N,ITPT,IT,GT,RT}(::Type{ScaledInterpolation{T,N,ITPT,IT,GT,RT}}) = ITPT
 basetype(sitp::ScaledInterpolation) = basetype(typeof(sitp))
 
-gradient{T,N,ITPT,IT<:DimSpec}(sitp::ScaledInterpolation{T,N,ITPT,IT}, xs::Number...) = gradient!(Array(T,count_interp_dims(IT,N)), sitp, xs...)
+# @eval uglyness required for disambiguation with method in b-splies/indexing.jl
+# also, GT is only specified to avoid disambiguation warnings on julia 0.4
+for R in [:Real, :Number]
+    @eval gradient{T,N,ITPT,IT<:DimSpec{InterpolationType},GT<:DimSpec{GridType}}(sitp::ScaledInterpolation{T,N,ITPT,IT,GT}, xs::$R...) =
```
tlycken: Same thing here - I'd rather just duplicate the method definition, especially since it's a one-liner. Leaving the comment about resolving a disambiguity should make it obvious enough what's going on when you see the two method definitions side by side.

PR #121, opened 2016-09-19
support precompilation for appropriate Julia versions

### LINKED ISSUE #120: Support precompilation
`Interpolations.jl` doesn't explicitly support Julia precompilation, but it seems like it probably ought to. It should just be a matter of adding 

``` julia
VERSION >= v"0.4.0-dev+6521" && __precompile__()
```

to `src/Interpolations.jl`. 

This is specifically a problem for me because Gadfly requires KernelDensity which requires Interpolations, so the lack of precompile support here is breaking precompilation of Gadfly for me. 

I'll open a PR if this seems reasonable.

rdeits:
fixes #120


src/Interpolations.jl
```diff
@@ -1,3 +1,5 @@
+VERSION >= v"0.4.0-dev+6521" && __precompile__()
```
anriseth: I'd say it's nicer to use
```jl
isdefined(Base, :__precompile__) && __precompile__()
```

rdeits:
Sure, that's also fine. In fact, Interpolations requires julia 0.4 anyway, so we might not need either test and could just have:

```julia
__precompile()__
```

 I'm open to any of these options. What do you think?

anriseth:
Yeah, I guess there's less of a reason to do the compatibility check if we
know it should work.

rdeits:
Okay, I've removed the version check entirely, since it's redundant with what's in REQUIRE

sglyon:
Looks good to me!

timholy:
@rdeits, I updated this and tried to merge it, but GitHub is claiming that required statuses are not passing (even though they are), and I don't have the authority to override the required checks. I'll submit a new PR (credited to you) and see if that has any better luck.

timholy:
See #136.

timholy:
Superseded by #136.

rdeits:
Thanks, @timholy !

PR #136, opened 2017-02-13
Turn on precompilation

timholy:
Often this is important for managing dependents.

Do over of #121.


timholy:
Wow, nightly is something like 5x faster at running the tests compared to 0.5. Nice.

timholy:
Wow, nightly is something like 5x faster at running the tests compared to 0.5. Nice.

PR #137, opened 2017-02-13
Test v0.4 on Travis

timholy:
We claim to support 0.4 but it wasn't being tested on Travis.

### DIFF
--- a/test/b-splines/mixed.jl
+++ b/test/b-splines/mixed.jl
@@ -26,13 +26,18 @@ for (constructor, copier) in ((interpolate, x->x), (interpolate!, copy))
 end
 
 # AbstractArrays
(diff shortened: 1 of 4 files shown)


test/b-splines/mixed.jl
```diff
@@ -26,13 +26,18 @@ for (constructor, copier) in ((interpolate, x->x), (interpolate!, copy))
 end
 
 # AbstractArrays
+if VERSION < v"0.5.0"
+    makesharedarray{T}(::Type{T}, dims; kwargs...) = SharedArray(T, dims; kwargs...)
```
tkelman: this isn't in Compat?

timholy: Nope.

PR #138, opened 2017-02-25
Fix typo in error message

mprat:
The error message when supplying the wrong set of extrapolation dimensions has a typo, causing the REPL to throw an error on the error.


timholy:
Thanks, @mprat! Sorry this went stale; once it gets through CI I'll merge it, and it will be in the next release.

PR #139, opened 2017-03-04
Apply compat updates for Julia 0.6

ararslan:
This should fix the deprecation warnings on 0.6.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,5 +3,5 @@ julia 0.4
 WoodburyMatrices 0.1.5
 Ratios
 AxisAlgorithms
-Compat 0.8.0
+Compat 0.19.0
 DualNumbers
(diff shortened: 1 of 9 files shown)


ararslan:
The only remaining deprecation warning specific to this package is from `Base.FloatRange`, and I'm not sure how best to deal with that; there doesn't appear to be a Compat solution for it.

ararslan:
Ping @timholy -- does this look okay?

timholy:
Thanks, @ararslan!

I wonder if we need some specializations for the new `LinSpace`, but it doesn't seem necessary to get the tests to pass (and that's really more an issue for the previous updates than this one).

ararslan:
My pleasure! Any chance you could tag a patch release with these changes?

timholy:
Sure. @tlycken, a reminder: could you turn on [attobot](https://github.com/attobot/attobot)? No one but you can do that.

tlycken:
@timholy Thanks for the reminder, I think it should be done now!

ararslan:
Thanks all! 🙂

PR #141, opened 2017-03-14
make Extrapolation subtype of AbstractExtrapolation

Evizero:


timholy:
Seems fine to me. I'm not even sure why we have the `Wrapper` version. It seems to have been added in #47, care to comment, @tlycken?

tlycken:
Heh, it's been a while... I *think* what I had in mind was to allow for dispatch on wrappers (scalings, extrapolations, whatevs...) that didn't also work for actual interpolation objects, but it doesn't seem to be used like that anywhere in the library.

Since `AbstractExtrapolation{...} <: AbstractInterpolationWrapper{...}` this change should be totally fine (and is probably more in line with what I intended at some point anyway).

Thanks, @Evizero!

Evizero:
happy to contribute. Would you mind tagging `v0.3.9` ?

PR #147, opened 2017-04-15
add support for Base.parent

Evizero:


sglyon:
Thanks @Evizero, I'm curious what the motivation behind this PR is. (another way of saying "what does `Base.parent` do)

Evizero:
When working with an array decorator (i.e. a subtype of `AbstractArray` that serves as a view around some other `AbstractArray`), `Base.parent` allows to query the underlying "parent" array of that decorator.

I am looking for this functionality specifically while working on this PR https://github.com/JuliaImages/ImageTransformations.jl/pull/13

timholy:
I think this could be OK. One thing bothers me a bit, though: for other array types that I'm familiar with, `parent` returns "the array that was used as an input to create the wrapper." That's not true for `BSplineInterpolation`. The docstring for `parent` specifically mentions view types, and `BSplineInterpolation` can be a view only for `Constant` and `Linear`.

Evizero:
Should we restrict `parent` to only work for those cases of `BSplineInterpolation`?

timholy:
That would be reasonable. It might set up a situation in which we'd be likely to get bug reports about its absence for higher-order interpolants; on the other hand, we could use it as a "teaching opportunity" to emit an informative error message, so that people understand the difference between the `coefs` array in these circumstances.

Evizero:
This should do the trick

PR #148, opened 2017-04-16
use ShowItLikeYouBuildIt for Base.show

Evizero:
This PR makes all `AbstractInterpolation` render their summary string in the REPL using the package [JuliaArrays/ShowItLikeYouBuildIt.jl](https://github.com/JuliaArrays/ShowItLikeYouBuildIt.jl).
This is a display-style that is used throughout the new Image ecosystem and makes the `summary` a little more informative for nested `AbstractArray` types.

for example, with this PR the types render in the REPL as follows:

```julia
julia> itp = interpolate(1:1.0:10, BSpline(Linear()), OnGrid())
10-element interpolate(::Array{Float64,1}, BSpline(Linear()), OnGrid()) with element type Float64:
  1.0
  2.0
  ⋮  
  9.0
 10.0

julia> sitp = scale(itp, -3:.5:1.5)
10-element scale(interpolate(::Array{Float64,1}, BSpline(Linear()), OnGrid()), (-3.0:0.5:1.5,)) with element type Float64:
  9.0
 11.0
  ⋮  
 25.0
 27.0

julia> etp = extrapolate(sitp, Flat())
10-element extrapolate(scale(interpolate(::Array{Float64,1}, BSpline(Linear()), OnGrid()), (-3.0:0.5:1.5,)), Flat()) with element type Float64:
  9.0
 10.0
  ⋮  
 10.0
 10.0
```

instead of what `master` does currently (scroll right to see the whole printout):

```julia
julia> itp = interpolate(1:1.0:10, BSpline(Linear()), OnGrid())
10-element Interpolations.BSplineInterpolation{Float64,1,Array{Float64,1},Interpolations.BSpline{Interpolations.Linear},Interpolations.OnGrid,0}:
  1.0
  2.0
  ⋮
  9.0
 10.0

julia> sitp = scale(itp, -3:.5:1.5)
10-element Interpolations.ScaledInterpolation{Float64,1,Interpolations.BSplineInterpolation{Float64,1,Array{Float64,1},Interpolations.BSpline{Interpolations.Linear},Interpolations.OnGrid,0},Interpolations.BSpline{Interpolations.Linear},Interpolations.OnGrid,Tuple{FloatRange{Float64}}}:
  9.0
 11.0
  ⋮
 25.0
 27.0

julia> etp = extrapolate(sitp, Flat())
10-element Interpolations.Extrapolation{Float64,1,Interpolations.ScaledInterpolation{Float64,1,Interpolations.BSplineInterpolation{Float64,1,Array{Float64,1},Interpolations.BSpline{Interpolations.Linear},Interpolations.OnGrid,0},Interpolations.BSpline{Interpolations.Linear},Interpolations.OnGrid,Tuple{FloatRange{Float64}}},Interpolations.BSpline{Interpolations.Linear},Interpolations.OnGrid,Interpolations.Flat}:
  9.0
 10.0
  ⋮
 10.0
 10.0
```

See the added tests for more examples.

## Context

@timholy and I were thinking about ways to make the REPL printout of interpolations and extrapolations a little more readable (see https://github.com/JuliaImages/ImageTransformations.jl/pull/13#issuecomment-294348743). For example a type that we are working with uses some extrapolation internally and thus the summary string gets really out of hand:

more text please

with this change it looks like this (don't mind the difference in the image preview; unrelated):

nice

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,6 @@
 julia 0.5
 
+ShowItLikeYouBuildIt
 WoodburyMatrices 0.1.5
 Ratios
 AxisAlgorithms
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -97,5 +97,6 @@ include("gridded/gridded.jl")
 include("extrapolation/extrapolation.jl")
 include("scaling/scaling.jl")
 include("utils.jl")
+include("io.jl")
 
 end # module
--- a/src/io.jl
+++ b/src/io.jl
@@ -1,0 +1,83 @@
+using ShowItLikeYouBuildIt
+
+Base.summary(A::AbstractInterpolation) = summary_build(A)
+
+function ShowItLikeYouBuildIt.showarg{T,N,TW,ST,GT}(io::IO, A::BSplineInterpolation{T,N,TW,ST,GT})
+    print(io, "interpolate(")
+    showarg(io, A.coefs)
+    print(io, ", ")
+    _showtypeparam(io, ST)
+    print(io, ", ")
+    _showtypeparam(io, GT)
+    print(io, ')')
+end
+
+function ShowItLikeYouBuildIt.showarg{T,N,TC,ST,K}(io::IO, A::GriddedInterpolation{T,N,TC,ST,K})
+    print(io, "interpolate(")
+    _showknots(io, A.knots)
+    print(io, ", ")
+    showarg(io, A.coefs)
+    print(io, ", ")
+    _showtypeparam(io, ST)
+    print(io, ')')
+end
+
+_showknots(io, A) = showarg(io, A)
+function _showknots{N}(io, tup::NTuple{N,Any})
+    print(io, '(')
+    for (i, A) in enumerate(tup)
+        showarg(io, A)
+        i < N && print(io, ',')
+    end
+    N == 1 && print(io, ',')
+    print(io, ')')
+end
+
+function ShowItLikeYouBuildIt.showarg(io::IO, A::ScaledInterpolation)
+    print(io, "scale(")
+    showarg(io, A.itp)
+    print(io, ", ", A.ranges, ')')
+end
+
+function ShowItLikeYouBuildIt.showarg{T,N,TI,IT,GT,ET}(io::IO, A::Extrapolation{T,N,TI,IT,GT,ET})
+    print(io, "extrapolate(")
+    showarg(io, A.itp)
+    print(io, ", ")
+    _showtypeparam(io, ET)
+    print(io, ')')
+end
+
+function ShowItLikeYouBuildIt.showarg{T,N,TI,IT,GT}(io::IO, A::FilledExtrapolation{T,N,TI,IT,GT})
+    print(io, "extrapolate(")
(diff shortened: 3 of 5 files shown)


sglyon:
Looks great, thank you.

timholy:
Branch looks like it could use an update, then I say merge.

Evizero:
There are no merge conflicts I do not think (?). It just shows that this branch isn't based on the current master. This only really makes a difference if we would like to do a fast-forward merge, which is untypical for PRs

Evizero:
I'll take a look just in case

timholy:
Hmm, I haven't seen that before, must have to do with the settings on this repo. OK, once it gets through CI we should be good to go.

Evizero:
Yes it seems to be as I suspected. This new button just helps one to avoid future merge conflicts by updating the branch with the new additions from master (new additions that aren't in conflict). It is not really necessary and in general will only cause additional merge commits that aren't needed.

I think in general it is better to do manual rebases if merge conflicts arise. Makes for a cleaner git history. If no merge conflict is present, updating a branch before merging a PR does not seems to have any benefits.

Evizero:
Not much testing going on in terms of CI. Could be worth to consider enabling OSX on Travis, and AppVeyor for windows?

timholy:
I've poked through the settings on both GitHub and Travis, and I'm not finding where such things are set. @tlycken, any suggestions?

Evizero:
Here is what I meant: https://github.com/JuliaMath/Interpolations.jl/commits/master. For the single commit "use ShowItLikeYouBuildIt for show", two merge commits were created overall (the last two commits). Without pressing the button there would only be one. Not that it makes such a big difference, it is just redundant.

Evizero:
Concerning where the button comes from. Maybe it is just a new GitHub feature?

Evizero:
did the button block the merge? or did it just look like a problem?

timholy:
It allowed me to merge but warned me I was having to exercise "administrative override." Since I'd never seen it before I didn't want to do anything that could be risky.

Evizero:
Well I could also be wrong in my assumption and there was some git weirdness going on. Safe is good. 👍 

Thanks for the reviews

tlycken:
The merge confusion seems to be a recently added GitHub feature, which I personally would probably recommend against using for the reasons @Evizero has already iterated.

Re: automated testing, I think the reasoning behind running only on linux was that the other two platforms took much longer (at least at the time I last did anything about this), often increasing the time from commit to CI feedback from one or two minutes to ten or twenty (depending mostly on how long the build queue on the agents were). Since Interpolations.jl is a pure-Julia package, I figured testing on other platforms wouldn't give enough extra information compared to the Linux builds; they *should* yield exactly the same results in all cases - only reason it wouldn't, is if we happen to stumble on a platform-specific bug in one of our dependencies.

Evizero:
good points 👍

> they should yield exactly the same results in all cases

It does until it doesn't. For example the AppVeyor tests have repeatedly revealed to me subtle issues in my libraries where a `Int64` was hard-coded instead of allowing other integers to work as well (because it tests on a 32 bit platforms).

PR #150, opened 2017-04-23
Make the Extrapolation constructor type-stable

Evizero:
small bugfix. The added `@inferred` in the tests would fail on current master.

### DIFF
--- a/src/extrapolation/extrapolation.jl
+++ b/src/extrapolation/extrapolation.jl
@@ -1,8 +1,8 @@
 type Extrapolation{T,N,ITPT,IT,GT,ET} <: AbstractExtrapolation{T,N,ITPT,IT,GT}
     itp::ITPT
 end
-Extrapolation{T,ITPT,IT,GT,ET}(::Type{T}, N, itp::ITPT, ::Type{IT}, ::Type{GT}, et::ET) =
-    Extrapolation{T,N,ITPT,IT,GT,ET}(itp)
+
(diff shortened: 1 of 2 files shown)


src/extrapolation/extrapolation.jl
```diff
@@ -28,7 +31,7 @@ You can also combine schemes in tuples. For example, the scheme `(Linear(), Flat
 Finally, you can specify different extrapolation behavior in different direction. `((Linear(),Flat()), Flat())` will extrapolate linearly in the first dimension if the index is too small, but use constant etrapolation if it is too large, and always use constant extrapolation in the second dimension.
 """
 extrapolate{T,N,IT,GT,ET<:ExtrapDimSpec}(itp::AbstractInterpolation{T,N,IT,GT}, et::ET) =
-    Extrapolation(T,N,itp,IT,GT,et)
+    Extrapolation(T,Val{N},itp,IT,GT,et)
```
timholy: Seems better to call `Extrapolation{T,N,ITPT,IT,GT,ET}(itp)` directly. We should probably deprecate the method on line 4.

Evizero: done

timholy:
Thanks!

timholy:
Feel free to tag a new version if you want this right away. I usually think that adding a new depwarn is grounds for a minor version bump.

Evizero:
I am not a JuliaMath member or a Package collaborator, so I am afraid I can't create releases

PR #157, opened 2017-05-10
Update README.md

### LINKED ISSUE #152: Add example of using `scale` to readme
ref #124

aaowens:
Add examples for scale. Fixes #152

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,31 @@ itp = interpolate!(A, BSpline(Quadratic(InPlace())), OnCell())
 ```
 which destroys the input `A` but also does not need to allocate as much memory.
 
+### Scaled BSplines 
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -135,6 +135,31 @@ itp = interpolate!(A, BSpline(Quadratic(InPlace())), OnCell())
 ```
 which destroys the input `A` but also does not need to allocate as much memory.
 
+### Scaled BSplines 
+
+BSplines assume your data is uniformly spaced on the grid `1:N`, or its multidimensional equivalent. If you have data of the form `[f(x) for x in A]`, you need to tell Interpolations about the grid `A`. If `A` is not uniformly spaced, you must use gridded interpolation described below. However, if `A` is a collection of ranges or linspaces, you can use scaled BSplines. This is more efficient because the gridded algorithm does not exploit the uniform spacing, and is currently the only way to use quadratic and cubic interpolation on grids.
```
timholy: The last clause, "...and is currently...", is a bit confusing: does it mean the gridded algorithm is the only way? What about for the implicit grids `1:N`?

timholy:
Thanks!

PR #159, opened 2017-05-23
Use zero and oneunit in place of one

timholy:
In Julia 0.6, a distinction is made between one(T) (multiplicative identity) and oneunit(T) ("1" of type T). In essentially every place we mean oneunit rather than one. However, packages like StaticArrays (for multivalued interpolation) don't define either, but `zero` will work for us.

Also deletes Interpolations-old.jl, which I don't think has a purpose.


src/gridded/gridded.jl
```diff
@@ -24,11 +24,11 @@ function GriddedInterpolation{N,TCoefs,TWeights<:Real,IT<:DimSpec{Gridded},pad}(
         issorted(k) || error("knot-vectors must be sorted in increasing order")
         iextract(IT, d) != NoInterp || k == collect(1:size(A, d)) || error("knot-vector should be the range 1:$(size(A,d)) for the method Gridded{NoInterp}")
     end
-    c = one(TWeights)
+    c = zero(TWeights)
```
sglyon: Can you remind me, what's the point of all the `c*c` multiplications when `c = 0`?

timholy: At one point we couldn't trust that `zero(T)*zero(T)` had type `T`: for example, I believe `0x00*0x00` was a `UInt64`. It's not really relevant for most types anymore. If one attached units to the weights, then the multiplications might still make sense, but though we can (or should be able to---I have haven't actually tested) interpolate unitful quantities, I don't see a reason that the weights should be unitful.

sglyon: Oh ok thanks. Makes sense. 

This changes look good to merge whenever you are ready

PR #167, opened 2017-07-12
Fix #166: Add example for parametric splines to README

### LINKED ISSUE #166: Add documentation for parametric 1D splines embedded on 2D grids
First off, thanks for this great package!

I am having difficulty in interpolating a 1D spline embedded on a 2D grid. Suppose I have a 2D Array where some pixels are marked as knots. I can easily interpolate these pixels manually with line segments and get a discretized polyline passing through all of them, however, I need something smooth.

Let's say I have 3 knots in a 100x100 image:

```julia
A = falses(100,100)
A[1,1] = A[30,50] = A[50,30] = A[100,100] = true
```

How can I fit a cubic spline passing through these activated pixels with Interpolations.jl? Ideally I would have a parametrization of the spline with `t in [0,1]` that I could use to find out the index `(i,j)` in the array where the spline is passing through. In the example, `S(0.) = (1,1)` and `S(1.) = (100,100)`.

juliohm:
This PR is documentation only.


timholy:
:+1: Not certain we want the plotting commands in there, but I'm not set against them either. What do others think?

juliohm:
I can remove the plotting pieces @timholy , or alternatively add a preview of the plot illustrating the example. Please let me know how I should proceed.

timholy:
If you show the plotted result then perhaps I'd leave the commands here. I'm fine with anything, really, as long as it clearly distinguishes "commands needed for the Interpolations.jl part" from "extra stuff to visualize the result." If before the `scatter` command you interrupted the quoted-text to say, "now we can visualize the results" perhaps it would be perfectly clear?

juliohm:
Hi Tim, where should I save the preview image? Can I put it in the doc folder?

juliohm:
I have added it to doc/images/ but let me know if it should reside somewhere else and I will update the PR.

sglyon:
Nope, looks good!

Thanks for the contribution!

On an unrelated note we should figure out why CI is failing. I'm just on a phone and am not good at reading travis logs on here so I'll have to check another time.

sglyon:
Hmm, I was going to merge, but it looks like our settings for the repo prevent merging when CI fails. 

@timholy what do you recommend here?

Evizero:
Nightly is failing even though the intention in the travis.yml is clearly to allow for that. I suspect the config has a slightly wrong syntax. The following should work

```
matrix:
  allow_failures:
    - julia: nightly
```

sglyon:
Good catch @Evizero. Is there a way for me to add a commit to this PR so we can get it to pass?

If not. @juliohm would you be willing to edit the .travis.yml file and delete the `-` in front of `allow_failures` so it matches what @Evizero proposed?

Thank you

juliohm:
Thank you @sglyon , fixed the syntax.

sglyon:
Thanks @juliohm judging by the start of the travis log, it doesn't seem to have worked.

Maybe try to copy/paste this directly:

```yaml
julia:
    - 0.5
    - 0.6
    - nightly
matrix:
    allow_failures:
        - julia: nightly
```

Sorry for having you do this extra work that doesn't pertain to your PR. Thank you though

juliohm:
My mistake, the syntax was still wrong. It should be fixed now.

sglyon:
Looks great! Thanks for doing that.

Once we get the green light I will merge.

PR #171, opened 2017-08-24
DOC: fix syntax in ipynb files. Closes #170

### LINKED ISSUE #170: Example  no longer working
```
nx = 10
f(x) = sin.(2pi/(nx-1) * (x-1))

xcoarse = 1:nx
ycoarse = f(xcoarse);

xfine = 1:.1:xcoarse[end]
yfine = f(xfine);
 
using Interpolations

yitp = interpolate(ycoarse, BSpline(Linear), OnGrid)
```
from the documentation does not run with  v0.6. 

LoadError: MethodError: Cannot `convert` an object of type Type{Interpolations.Linear} to an object of type Interpolations.BSpline
This may have arisen from a call to the constructor Interpolations.BSpline(...),
since type constructors fall back to convert methods.
while loading C:\Users\PetrKrysl\.julia\v0.6\OtolithMotion\src\OtolithMotion.jl, in expression starting on line 29
in Type at base\sysimg.jl:24

sglyon:
This fixes the deprecated syntax calls in the jupyter notebooks in the doc folder


sglyon:
Hmm, I'm not sure what is causing the failures here.

If anyone has a chance to take a look that'd be great!

timholy:
Looks like a performance test problem. Travis is not the best place for those, perhaps we should just drop them?

sglyon:
Yeah, I think dropping performance tests is the right call here.

Issue #122, opened 2016-09-30
Hessian fails for function of two variables


simonp0420:
In converting over some code that currently uses Grid.jl, I need to compute the curvature of a surface.  The following REPL transcript illustrates the problem I'm having:

```
julia> A = rand(5,5)
5×5 Array{Float64,2}:
 0.264768   0.405079  0.253461  0.759207  0.673105
 0.0884208  0.528674  0.244601  0.545683  0.751446
 0.403174   0.68685   0.722238  0.746143  0.42529
 0.396175   0.651159  0.909525  0.402493  0.251807
 0.130116   0.986705  0.259471  0.51158   0.850608

julia> itp = interpolate(A, BSpline(Cubic(Natural())), OnGrid())
5×5 Interpolations.BSplineInterpolation{Float64,2,Array{Float64,2},Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}},Interpolations.OnGrid,1}:
 0.264768   0.405079  0.253461  0.759207  0.673105
 0.0884208  0.528674  0.244601  0.545683  0.751446
 0.403174   0.68685   0.722238  0.746143  0.42529
 0.396175   0.651159  0.909525  0.402493  0.251807
 0.130116   0.986705  0.259471  0.51158   0.850608

julia> h = hessian(itp,2.0,2.0)
ERROR: UndefVarError: dim not defined
 in hessian_coefficients(::Type{Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}}}, ::Int64, ::Int64, ::Int64) at c:\home\simonp\.julia\v0.5\Interpolations\src\b-splines\indexing.jl:18
 in hessian_impl(::Type{Interpolations.BSplineInterpolation{Float64,2,Array{Float64,2},Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}},Interpolations.OnGrid,1}}) at c:\home\simonp\.julia\v0.5\Interpolations\src\b-splines\indexing.jl:123
 in hessian!(...) at c:\home\simonp\.julia\v0.5\Interpolations\src\b-splines\indexing.jl:144
 in hessian(::Interpolations.BSplineInterpolation{Float64,2,Array{Float64,2},Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}},Interpolations.OnGrid,1}, ::Float64, ::Float64) at c:\home\simonp\.julia\v0.5\Interpolations\src\b-splines\indexing.jl:153
```

[...]

PR #181, opened 2017-09-20
Correct bugs in the b-spline's hessian calculation (fixes #180)

### LINKED ISSUE #180: ERROR: UndefVarError: TH not defined
I'm getting an error message from a call to `hessian`, though I think I've set everything up correctly. This happens with both the latest released version of Interpolations.jl as well as master.

```
julia> A = randn(10,10);

julia> itp = interpolate(A, BSpline(Cubic(Line())), OnGrid());

julia> hessian(itp, 5., 5.,)
ERROR: UndefVarError: TH not defined
Stacktrace:
 [1] hessian(::Interpolations.BSplineInterpolation{Float64,2,Array{Float64,2},Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}},Interpolations.OnGrid,1}, ::Float64, ::Float64) at /home/jeff/.julia/v0.6/Interpolations/src/b-splines/indexing.jl:160
```

jeff-regier:
Here are a couple of changes that get `hessian` working again for b-splines.

### DIFF
--- a/src/b-splines/indexing.jl
+++ b/src/b-splines/indexing.jl
@@ -16,10 +16,10 @@ function gradient_coefficients(::Type{IT}, N, dim) where IT<:DimSpec{BSpline}
 end
 function hessian_coefficients(::Type{IT}, N, dim1, dim2) where IT<:DimSpec{BSpline}
     exs = if dim1 == dim2
-        Expr[d==dim1==dim2 ? hessian_coefficients(iextract(IT, dim), d) :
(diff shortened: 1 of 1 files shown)


jeff-regier:
Ah, so this PR fixes the obvious syntax errors, but I don't think the Hessians are correct :-(

timholy:
#122. Sadly, someone who uses hessians needs to sit down and work out the math again.

timholy:
Superseded by #226

PR #182, opened 2017-10-28
Fix gradients for ScaledInterpolation with NoInterp

magerton:
when some dimensions had `NoInterp`(olation) and were not at the very end,
they were re-scaled with the incorrect range. This PR provides a fix & test.


timholy:
Great fix and elegant test, many thanks!

timholy:
New release tagged.

magerton:
Thanks, @timholy! :)

PR #195, opened 2018-03-01
use json instead of jld2 param storage for benchmarks

rdeits:
BenchmarkTools 0.2+ uses JSON rather than JLD2 for storing its suite parameters: https://github.com/JuliaCI/BenchmarkTools.jl/blob/30d9aaa0c674815aae5ff5fba6360fb6809869dd/doc/manual.md#caching-parameters

### DIFF
--- a/perf/benchmarks.jl
+++ b/perf/benchmarks.jl
@@ -96,12 +96,12 @@ for A in (collect(Float64, 1:3),
     end
 end
 
-paramspath = joinpath(dirname(@__FILE__), "params.jld")
+paramspath = joinpath(dirname(@__FILE__), "params.json")
 
 if isfile(paramspath)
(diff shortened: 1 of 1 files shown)


rdeits:
Yes, that's right; it looks like this behavior changed in BenchmarkTools v0.2 and higher. As of the current version, v0.2.5, trying to save parameters as `.jld2` causes BenchmarkTools to throw the following error: https://github.com/JuliaCI/BenchmarkTools.jl/blob/30d9aaa0c674815aae5ff5fba6360fb6809869dd/src/serialization.jl#L54

stevengj:
Looks like this should be merged?

rdeits:
I think so, but I haven't tried this out since March

stevengj:
Rebase?

rdeits:
rebased

PR #197, opened 2018-03-01
remove call to Core.Inference to try to fix v0.7

rdeits:
As far as I know, directly calling `Core.Inference` is not generally recommended. `Base.promote_op` should do the same thing in this case and also works correctly on Julia v0.7.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,5 +4,5 @@ ShowItLikeYouBuildIt
 WoodburyMatrices 0.1.5
 Ratios
 AxisAlgorithms
-Compat 0.19.0
+Compat 0.49
 DualNumbers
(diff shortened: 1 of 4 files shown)


lhupe:
there's another call to `Core.Inference` in `src/gridded/gridded.jl` at line 31

rdeits:
Could one of the admins please take a look at this? It's been open for 4 months

rdeits:
@lhupe thanks--I've made the same change to that line now

timholy:
If we don't care about interpolating empty containers, then
```julia
T = typeof(c*first(A))
```
seems a lot safer. Or we could use
```julia
    if isempty(A)
        T = Base.promote_op(*, typeof(c), Tel)
    else
        T = typeof(c*first(A))
    end
```

rdeits:
Ok, done and rebased

rdeits:
Ok, rebased and passing using @timholy's suggestion. Good to go?

PR #208, opened 2018-06-20
move promote_type call inside the generated function body

### LINKED ISSUE #207: Non-pure `@generated` functions cause hard-to-fix MethodErrors
This line: https://github.com/JuliaMath/Interpolations.jl/blob/839c1d3f6aaa855b4e384c40bd36adf2a72dcf93/src/b-splines/indexing.jl#L111

calls `promote_type` within the generator of a `@generated` function, which violates the rule that generated function generators must not observe any non-constant global state (in this case, it's observing the state of the set of existing methods). The problem is that this means that the generated code essentially has the set of `promote_type` methods baked in at the time the generator is first invoked. That creates some errors that are very hard to track down. For example, the following code: 

```julia
using Interpolations
ext = extrapolate(interpolate(eye(2), BSpline(Linear()), OnGrid()), Linear())
using ForwardDiff
x = ForwardDiff.Dual(1.0, 1.0)
ext[x, 2.5]
```

(with ForwardDiff master and Interpolations master) fails with: 

```
ERROR: LoadError: no promotion exists for ForwardDiff.Dual{Void,Float64,1} and Int64
Stacktrace:
 [1] promote_type(::Type{ForwardDiff.Dual{Void,Float64,1}}, ::Type{Int64}) at ./promotion.jl:161
 [2] promote_type(::Type{T} where T, ::Type{T} where T, ::Type{T} where T) at ./promotion.jl:127
 [3] gradient(...) at /home/rdeits/.julia/v0.6/Interpolations/src/b-splines/indexing.jl:111
 [4] macro expansion at /home/rdeits/.julia/v0.6/Interpolations/src/extrapolation/linear.jl:17 [inlined]
 [5] macro expansion at /home/rdeits/.julia/v0.6/Interpolations/src/extrapolation/extrap_prep.jl:89 [inlined]
 [6] macro expansion at /home/rdeits/.julia/v0.6/Interpolations/src/extrapolation/extrapolation.jl:57 [inlined]
 [7] getindex(::Interpolations.Extrapolation{Float64,2,Interpolations.BSplineInterpolation{Float64,2,Array{Float64,2},Interpolations.BSpline{Interpolations.Linear},Interpolations.OnGrid,0},Interpolations.BSpline{Interpolations.Linear},Interpolations.OnGrid,Interpolations.Linear}, ::ForwardDiff.Dual{Void,Float64,1}, ::Float64) at /home/rdeits/.julia/v0.6/Interpolations/src/extrapolation/extrapolation.jl:63
```

but if we just *move the ForwardDiff import earlier*, then everything is fine: 

```julia
using ForwardDiff
using Interpolations
ext = extrapolate(interpolate(eye(2), BSpline(Linear()), OnGrid()), Linear())
x = ForwardDiff.Dual(1.0, 1.0)
ext[x, 2.5]
```

This happens because if we import ForwardDiff first, then the baked-in version of the method table happens to include the promotion rules that ForwardDiff defines. 

The solution is just not to use promote_type inside the generator, but to just call it in the returned expression.

rdeits:
Fixes #207 (at least for the particular case i showed there). 

I also changed the equivalent code in `hessian()`, although i think that code may have been broken already (`Array{TH` should have been `Array{$TH`). I'll see what the test coverage says. 

Anyway, please let me know if you think this is the right thing to do. If so, I should be able to construct a test to make sure #207 is really fixed.

### DIFF
--- a/src/b-splines/indexing.jl
+++ b/src/b-splines/indexing.jl
@@ -160,9 +160,11 @@ end
 
 @generated function hessian(itp::AbstractInterpolation{T,N}, xs...) where {T,N}
     n = count_interp_dims(itp,N)
-    TH = promote_type(T, [x <: AbstractArray ? eltype(x) : x for x in xs]...)
     xargs = [:(xs[$d]) for d in 1:length(xs)]
-    :(hessian!(Array{TH, 2}($n,$n), itp, $(xargs...)))
+    quote
+        TH = $(Expr(:call, :promote_type, T, [x <: AbstractArray ? eltype(x) : x for x in xs]...))
+        hessian!(Array{TH, 2}($n,$n), itp, $(xargs...))
+    end
 end
 
 hessian1(itp::AbstractInterpolation{T,1}, x) where {T} = hessian(itp, x)[1,1]
--- a/src/b-splines/indexing.jl
+++ b/src/b-splines/indexing.jl
@@ -108,9 +108,11 @@ end
 for R in [:Real, :Any]
     @eval @generated function gradient(itp::AbstractInterpolation{T,N}, xs::$R...) where {T,N}
         n = count_interp_dims(itp, N)
-        Tg = promote_type(T, [x <: AbstractArray ? eltype(x) : x for x in xs]...)
         xargs = [:(xs[$d]) for d in 1:length(xs)]
-        :(gradient!(Array{$Tg, 1}($n), itp, $(xargs...)))
+        quote
+            Tg = $(Expr(:call, :promote_type, T, [x <: AbstractArray ? eltype(x) : x for x in xs]...))
+            gradient!(Array{Tg, 1}($n), itp, $(xargs...))
+        end
     end
 end
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,8 @@ julia:
 matrix:
     allow_failures:
         - julia: nightly
+after_success:
+  # push coverage results to Coveralls
+  - julia -e 'cd(Pkg.dir("WebIO")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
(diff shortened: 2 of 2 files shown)


src/b-splines/indexing.jl
```diff
@@ -158,9 +160,11 @@ end
 
 @generated function hessian(itp::AbstractInterpolation{T,N}, xs...) where {T,N}
     n = count_interp_dims(itp,N)
-    TH = promote_type(T, [x <: AbstractArray ? eltype(x) : x for x in xs]...)
     xargs = [:(xs[$d]) for d in 1:length(xs)]
-    :(hessian!(Array{TH, 2}($n,$n), itp, $(xargs...)))
+    quote
+        TH = $(Expr(:call, :promote_type, T, [x <: AbstractArray ? eltype(x) : x for x in xs]...))
```
stevengj: The problem is that this constructs an array and does the type computations at runtime.  It may not be type-stable, either.  It seems like you'd be better off doing something like `TH = hessian_type(T, xs...)` with
```jl
hessian_type(T) = T
hessian_type(T, x::AbstractArray, xs...) = promote_type(eltype(x), hessian_type(T, xs...))
hessian_type(T, x, xs...) = promote_type(typeof(x), hessian_type(T, xs...))
```

rdeits: Are you sure? That array is constructed inside the generator, not in the run-time code, and all the branching likewise happens inside the generator. I'm a fan of the lisp-y recursion trick anyway, but I'm pretty sure this case will work as-is.

stevengj: Oh, right, I misread your code — you're right, ther eshould continue to be no runtime cost here.

rdeits:
Rebased. Good to go?

rdeits:
Re-rebased after #197 :slightly_smiling_face:

PR #217, opened 2018-07-25
A small step towards 0.7 compatibility.

hessammehr:
Tests pass without issue on 0.6 as well.

### DIFF
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -45,8 +45,8 @@ import Base: convert, size, indices, getindex, gradient, promote_rule,
              ndims, eltype, checkbounds
 
 # Julia v0.5 compatibility
-if isdefined(:scaling) import Base.scaling end
-if isdefined(:scale) import Base.scale end
+if isdefined(Base, :scaling) import Base.scaling end
+if isdefined(Base, :scale) import Base.scale end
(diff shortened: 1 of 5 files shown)


halleysfifthinc:
Travis is allowing failures on nightly; there are still a number of deprecation warnings starting for Interpolations.jl at [line 670](https://travis-ci.org/JuliaMath/Interpolations.jl/jobs/408148043#L670).

hessammehr:
Yes, I'll keep the commits coming.

hessammehr:
The big blocking issue is that tridiagonal matrices are `Tridiagonal{T} where T` in 0.6 and `Tridiagonal{T,V where V<:Array{T,1}}` in 0.7. This seems to affect both Interpolations.jl and AxisAlgorithms.jl. Does this seem like something that could be added to Compat.jl? The fix should be fairly simple:

```julia
const Tridiag{T} =  VERSION > v"0.7.0-DEV.1412" ? (Tridiagonal{T,V} where V<:AbstractArray{T} where T) : (Tridiagonal{T} where T)
```

timholy:
Sorry this didn't get attention!

Interpolations is now 0.7-compatible (0.7-requiring, actually)

PR #218, opened 2018-07-26
Fixes for julia 0.7

carlobaldassi:
(Note that this relies on https://github.com/timholy/AxisAlgorithms.jl/pull/11.)
This fixes all errors and deprecation warnings on 0.7, except for the IO part, whose tests are disabled. The objects are simply printed differently on 0.7, e.g. instead of
```
"8×20 interpolate(::Array{Float64,2}, BSpline(Constant()), OnCell()) with element type Float64"
```
you get
```
"8×20 Interpolations.BSplineInterpolation{Float64,2,Array{Float64,2},Interpolations.BSpline{Interpolations.Constant},Interpolations.OnCell,0}" == "8×20 interpolate(::Array{Float64,2}, BSpline(Constant()), OnCell()) with element type Float64"
```
I suspect that the problem there may be with the ShowitLikeYouBuildIt package, but I haven't investigated thoroughly.
In any case, this PR is probably large enough as it is, and the IO problems might be solved at a later time.

The main thing to note about this PR is that it can expose what I believe is a bug in Julia 0.7-beta during the inference of some `getindex` generated functions. To make things work here, I have had to define some specialized methods for `index_gen` which shouldn't really be necessary (everything works fine without them on 0.6, but on 0.7 despite the fact that `index_gen` works correctly by itself, its inference triggers an "unexpected runtime error" in julia). All these extra methods are marked with "FIXME" comments. This should then be reported as a bug in julia, and the methods removed when it's fixed.

I have also removed a deprecated `Extrapolation` constructor since I assume this will be a new minor version. It's easy to put it back in though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: julia
 sudo: false
 julia:
     - 0.6
+    - 0.7
     - nightly
 matrix:
     allow_failures:
--- a/src/b-splines/indexing.jl
+++ b/src/b-splines/indexing.jl
@@ -31,7 +31,7 @@ function getindex_impl(itp::Type{BSplineInterpolation{T,N,TCoefs,IT,GT,Pad}}) wh
     quote
         $meta
         @nexprs $N d->(x_d = xs[d])
-        inds_itp = indices(itp)
+        inds_itp = axes(itp)
 
         # Calculate the indices of all coefficients that will be used
         # and define fx = x - xi in each dimension
@@ -60,7 +60,7 @@ function gradient_impl(itp::Type{BSplineInterpolation{T,N,TCoefs,IT,GT,Pad}}) wh
     # For each component of the gradient, alternately calculate
     # coefficients and set component
     n = count_interp_dims(IT, N)
-    exs = Array{Expr, 1}(2n)
+    exs = Array{Expr, 1}(undef, 2n)
     cntr = 0
     for d = 1:N
         if count_interp_dims(iextract(IT, d), 1) > 0
@@ -74,7 +74,7 @@ function gradient_impl(itp::Type{BSplineInterpolation{T,N,TCoefs,IT,GT,Pad}}) wh
         $meta
         length(g) == $n || throw(ArgumentError(string("The length of the provided gradient vector (", length(g), ") did not match the number of interpolating dimensions (", n, ")")))
         @nexprs $N d->(x_d = xs[d])
-        inds_itp = indices(itp)
+        inds_itp = axes(itp)
 
         # Calculate the indices of all coefficients that will be used
         # and define fx = x - xi in each dimension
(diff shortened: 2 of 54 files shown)


RalphAS:
I suggest incorporating #219 which should obviate the extra `index_gen` methods.

carlobaldassi:
Thanks @RalphAS, I did that and indeed it works.

RalphAS:
I think I have a fix for the IO stuff, which I can put in when this is merged (unless you want me to try a piggy-back PR).

halleysfifthinc:
Is there anything I can do to help this along? Looking at the Travis logs for the last build, I can't see why the MethodError is being thrown, it looks to me like it should be matching the `_A_ldiv_B_md!` from AxisAlgorithms.

carlobaldassi:
@halleysfifthinc I think (hope) it's just a matter of waiting for the new version of AxisAlgorithms to be tagged ([which should happen shortly](https://github.com/JuliaLang/METADATA.jl/pull/16136)) so that it can be retrieved by the Travis CI.

halleysfifthinc:
Ahh. I had seen earlier that the new version of AxisAlgorithms was released but forgot that it had not gotten through METADATA

carlobaldassi:
Ok checks are green now :)
I have updated the requirements on AxisAlgorithms.
I think this is ready to merge so that @RalphAS can fix the IO stuff next.
(There are now some new deprecations from the Ratios package, which are non blocking though.)

.travis.yml
```diff
@@ -10,4 +11,4 @@ after_success:
   # push coverage results to Coveralls
   - julia -e 'cd(Pkg.dir("WebIO")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
   # push coverage results to Codecov
-  - julia -e 'cd(Pkg.dir("WebIO")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
```
sglyon: This looks odd -- why do we cd to WebIO.jl?

I don' think you changed this during this PR, but I just noticed it...

src/b-splines/indexing.jl
```diff
@@ -24,14 +24,21 @@ function hessian_coefficients(::Type{IT}, N, dim1, dim2) where IT<:DimSpec{BSpli
     Expr(:block, exs...)
 end
 
-index_gen(::Type{IT}, N::Integer, offsets...) where {IT} = index_gen(iextract(IT, min(length(offsets)+1, N)), IT, N, offsets...)
+function index_gen(::Type{IT}, N::Integer, offsets...) where {IT}
+    idx = index_gen(iextract(IT, min(length(offsets)+1, N)), IT, N, offsets...)
+    if v"0.7-" ≤ VERSION < v"0.7.0-beta2.119"
```
sglyon: Let's add `@static` to this condition. 

Although this is a meta-programming function and having an explicit branch shouldn't impact performance of evaluating an interpolation, removing the branch at parse time makes sense to me.

Evizero:
Nice! I think the IO part has to do with the fact that I think some of ShowitLikeYouBuildIt moved to base. So its probably just about figuring out where to hook into that machinery

carlobaldassi:
@sglyon I have fixed those two issues.

Evizero:
could someone please tag a new minor version?

Issue #215, opened 2018-07-20
StackOverflowError with julia 0.7


Alexander-Barth:
The following code, make Julia 0.7 beta2 and 0.7.0-beta2.55 crash (under Linux):
```
using Interpolations
x1 = Float64[x for x = 1:5, y = 1:6]
y1 = Float64[y for x = 1:5, y = 1:6]

itp = interpolate((x1[:,1],y1[1,:]),x1,Gridded(Linear()))
itp[3,3]
```

This is with a checkout of Interpolation for the master branch (commit 615f9d7).
The error starts with:
```
┌ Warning: Deprecated syntax `parametric method syntax Extrapolation{T, ITPT, IT, GT, ET}(::Type{T}, N, itp::ITPT, ::Type{IT}, ::Type{GT}, et::ET)` around deprecated.jl:53.
│ Use `Extrapolation(#unused#::Type{T}, N, itp::ITPT, #unused#::Type{IT}, #unused#::Type{GT}, et::ET) where {T, ITPT, IT, GT, ET}` instead.
└ @ deprecated.jl:53
Internal error: encountered unexpected error in runtime:
StackOverflowError()
jl_typemap_insert_generic at /buildworker/worker/package_linux64/build/src/typemap.c:914
jl_typemap_array_insert_ at /buildworker/worker/package_linux64/build/src/typemap.c:936
jl_typemap_level_insert_ at /buildworker/worker/package_linux64/build/src/typemap.c:976
jl_typemap_insert_generic at /buildworker/worker/package_linux64/build/src/typemap.c:915
jl_typemap_array_insert_ at /buildworker/worker/package_linux64/build/src/typemap.c:936
...
```

It could also be, of course, an bug in Julia 0.7. Any help would be greatly appreciated.

PR #219, opened 2018-07-26
workaround for inference bug

RalphAS:
This addresses #215 (cf. JuliaLang/julia#27907) for Julia v0.7- through beta2.87 and should impose no extra cost after the inference bug is fixed.


carlobaldassi:
This was incorporated in #218 and can be closed.

PR #220, opened 2018-08-03
add 0.7-compatible showarg methods

RalphAS:
This change adds versions of the `showarg` methods compatible with the usage in 0.7 `Base`.
The old `ShowItLikeYouBuildIt` versions are still used with older Julia.
Output format is the same as in the earlier versions of Interpolations.
Corresponding tests are re-enabled.

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -1,3 +1,5 @@
+# after this, functionality was incorporated into Base
+if VERSION < v"0.7.0-DEV.1790"
 using ShowItLikeYouBuildIt
 
 Base.summary(A::AbstractInterpolation) = summary_build(A)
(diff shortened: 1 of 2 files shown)


timholy:
Just noting this is a really crucial PR (we have decided we are *not* upgrading `ShowItLikeYouBuildIt` for 0.7), so we need to ditch the dependency and go 0.7+. There are some conflicts here but I am fixing them now.

tlycken:
> we need to ditch the dependency and go 0.7+

Is it possible to specify that it is a dependency for 0.6, but from 0.7 and up? Since there are quite a lot of things you need to update in your code to move between 0.6 and 0.7/1.0, I think it would be nice if we could have as long a time window as possible where we support both.

timholy:
Incorporated into #226

PR #222, opened 2018-08-10
Fix testing failure on julia 1.0

felixcremer:
This removes the import of LinearAlgebra: gradient  and exports the gradient function because it has been fully removed (see https://github.com/JuliaLang/julia/issues/16113).
It also includes the changes of #221 to Random.srand.

### DIFF
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -8,6 +8,7 @@ export
     extrapolate,
     scale,
 
+    gradient,
     gradient!,
     gradient1,
     hessian!,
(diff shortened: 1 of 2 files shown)


tlycken:
Gradient seems to me like it's a too general concept to be "owned" by Interpolations.jl. Is there somewhere else we should import the name from to make sure we're extending the concept, rather than exporting a new one? (Or is import merging solved nowadays?)

timholy:
No, import merging isn't solved. I don't think we should export it.

mkborregaard:
Is it correct that merge of this PR is what keeps this package back from 1.0 compatibility?

tlycken:
@mkborregaard Technically, no. But (one of) the problem(s) addressed by this PR is, namely that we need to decide how to handle that `gradient` is no longer exported from base.

felixcremer:
Everything this tries to solve is handled in #226

PR #223, opened 2018-08-10
Test on Julia 1.0

tlycken:
Now that Julia 1.0 is released (🎉 🎉 🎉 ❗️ ) we should include it in our testing pipeline :)


timholy:
Superseded by #226

PR #225, opened 2018-08-13
BREAKING: Unexport gradient and friends

tlycken:
This is another take on solving the same problems as #222, but instead of exporting a new function `gradient`, I have removed it (and its friends) from the list of exported functions.

By qualifying all usages in the test suite as `Interpolations.gradient`, they now pass on 1.0! 🎉 

However, I'd like to figure out a way to make this transition a little kinder to the users. Is there a way to deprecate usage like
```
using Interpolations
itp = ...
gradient(itp, ...)
```
while still allowing
```
using Interpolations
itp = ...
Interpolations.gradient(itp, ...)
```
?

If not, maybe we just have to accept that this is part of a somewhat painful transition to 1.0.

Edit: I've incorporated #223 here too, because making it work on Julia 1.0 is the main motivation behind this PR. Doesn't matter to me if we merge #223 first and then this (which will effectively rebase that commit away) or close #223 and take the change from here.

---

Side note: It's not so weird to me after making these changes that we have problems with Hessians (se e.g. #122 and #181). There's not a single mention of `hessian` in the test suite...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: false
 julia:
     - 0.6
     - 0.7
+    - 1.0
     - nightly
 matrix:
     allow_failures:
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -8,12 +8,6 @@ export
     extrapolate,
     scale,
 
-    gradient!,
-    gradient1,
-    hessian!,
-    hessian,
-    hessian1,
-
     AbstractInterpolation,
     AbstractExtrapolation,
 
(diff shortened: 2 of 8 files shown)


sglyon:
I agree with the notion of Interpolations not "claiming" the function gradient (can't remember which issue I saw that comment in) and not exporting the function is a good way to accomplish that.

I think users have two options:

```
using Interpolations
itp = ...
Interpolations.gradient(itp, ...)
```

or 

```
using Interpolations
using Interpolations.gradient
itp = ...
gradient(itp, ...)
```

As long as `gradient` has not been loaded/imported by another used package already.

timholy:
If you're eager this can be merged, but I already have this on my local branch and lots of other work too.

If you do merge, I would argue *against* tagging 0.7/1.0 compatible just yet. I am getting very close (<1day?) to having a full rewrite done.

timholy:
Superseded by #226

Issue #224, opened 2018-08-11
Rewrite


timholy:
This is to give heads-up about some big coming changes. I'll check them off as I complete them locally and put it all in one gigantic PR.

In addition to the agenda in #212,

- [X] drop support for 0.6 and get rid of Compat
- [X] stop exporting `gradient`, `gradient!`, `gradient1`, `hessian`, `hessian!`, and `hessian1`.
- [X] deprecate `itp[x]` in favor of `itp(x)`
- [ ] (not done, but underway) get rid of all the generated functions
- [ ] don't use `clamp` for interpolation objects (make them throw `BoundsError`s)
- [ ] consider renaming `OnGrid` to `OnKnot` 
- [ ] improve performance

tlycken:
> stop exporting gradient, gradient!, gradient1, hessian, hessian!, and hessian1.
👍 🎉 
> drop support for 0.6 and get rid of Compat
As noted elsewhere: I think it makes sense to keep the window where we support 0.6 _and_ 0.7/1.0 open for as long as possible, to give people a chance to update their code. (Deprecation warnings for stuff we need to remove when dropping 0.6 is fine, of course.)
> (not done, but underway) get rid of all the generated functions
Nice!
> improve performance
I wouldn't believe it was you coding if you didnt... 😄

tlycken:
I must be really unattentive today. Read this, and yet went ahead and filed #225 anyway. Feel free to close that and move on with your local version instead, if the git-juggling feels unnecessarily complex.

Otherwise, I think #225 might be the quickest way to 1.0-compatibility (at least as far as our test suite knows 😅).

timholy:
I completely disagree with supporting 0.6 at the same time as 0.7. By bumping the minimum julia version in REQUIRE we allow people still on 0.6 to run "old" versions of Interpolations that continue to work on 0.6. Only people running 0.7 and higher will get the new code.

0.7/1.0 is the future, and it's also here now. So if we're doing a big update there is not reason to cruft the code.

What I'm striving for is to do all our breaking at once. It seems we also want to deprecate the `itp[x]` syntax. Right now people are in an upgrade cycle and so will view depwarns from Interpolations as part of background noise; a month from now they might be annoyed ("I thought I was done fixing depwarns!")

tlycken:
That are very good points; I didn't consider the option of pinning older Interpolations versions if you're stuck on 0.6 for some reason.

timholy:
[...]

PR #226, opened 2018-08-14
Break all the things

### LINKED ISSUE #122: Hessian fails for function of two variables

### LINKED ISSUE #127: Throw not exported?
For extrapolation, `Throw` is not exported, but `Flat` is. Any reason for this?

### LINKED ISSUE #151: Quadratic/Cubic Interpolations Not Working for Dimensions > 3
The following works as expected:
```
using Interpolations
V = zeros(10,10,10)
interpV = interpolate(V, BSpline(Cubic(Line())), OnGrid())
```
However, the following throws an error:
```
using Interpolations
V = zeros(10,10,10,10)
interpV = interpolate(V, BSpline(Cubic(Line())), OnGrid())
```
Specifically:
```
ERROR: MethodError: no method matching Array{Float64,4}(::Int64, ::Int64, ::Int64, ::Int64)
Closest candidates are:
  Array{Float64,4}{T}(::Any) at sysimg.jl:53
 in alloc_matmul(::SparseMatrixCSC{Float64,Int64}, ::Array{Float64,4}, ::Int64) at /Users/nhipsman/.julia/v0.5/AxisAlgorithms/src/AxisAlgorithms.jl:42
 in _A_ldiv_B_md!(::Array{Float64,4}, ::WoodburyMatrices.Woodbury{Float64,Base.LinAlg.LU{Float64,Tridiagonal{Float64}},SparseMatrixCSC{Float64,Int64},SparseMatrixCSC{Float64,Int64},Array{Float64,2},Array{Float64,2}}, ::Array{Float64,4}, ::CartesianRange{CartesianIndex{0}}, ::CartesianRange{CartesianIndex{3}}, ::Array{Float64,1}) at /Users/nhipsman/.julia/v0.5/Interpolations/src/b-splines/../filter1d.jl:77
 in A_ldiv_B_md!(::Array{Float64,4}, ::WoodburyMatrices.Woodbury{Float64,Base.LinAlg.LU{Float64,Tridiagonal{Float64}},SparseMatrixCSC{Float64,Int64},SparseMatrixCSC{Float64,Int64},Array{Float64,2},Array{Float64,2}}, ::Array{Float64,4}, ::Int64, ::Array{Float64,1}) at /Users/nhipsman/.julia/v0.5/Interpolations/src/b-splines/../filter1d.jl:15
 in prefilter!(::Type{Float64}, ::Array{Float64,4}, ::Type{Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}}}, ::Type{Interpolations.OnGrid}) at /Users/nhipsman/.julia/v0.5/Interpolations/src/b-splines/prefiltering.jl:58
 in prefilter(::Type{Float64}, ::Type{Float64}, ::Array{Float64,4}, ::Type{Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}}}, ::Type{Interpolations.OnGrid}) at /Users/nhipsman/.julia/v0.5/Interpolations/src/b-splines/prefiltering.jl:40
 in interpolate(::Type{Float64}, ::Type{Float64}, ::Array{Float64,4}, ::Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}}, ::Interpolations.OnGrid) at /Users/nhipsman/.julia/v0.5/Interpolations/src/b-splines/b-splines.jl:64
 in interpolate(::Array{Float64,4}, ::Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}}, ::Interpolations.OnGrid) at /Users/nhipsman/.julia/v0.5/Interpolations/src/b-splines/b-splines.jl:68
```
Unfortunately I'm not quite an experienced enough Julia programmer to get into the weeds of the source code, but hopefully someone here can.  Thanks!

### LINKED ISSUE #158: Scaled interpolation does not automatically extrapolate when called with a vector
The following example is taken from the readme
```
A_x = 1.:2.:40.
A = [log(x) for x in A_x]
itp = interpolate(A, BSpline(Cubic(Line())), OnGrid())
sitp = scale(itp, A_x)
sitp[3.]
sitp[3.5]
```
This can also be called in one go by passing a vector to be evaluated
```
sitp[[3.0, 3.5]]
2-element Array{Float64,1}:
 1.09861
 1.22632
```

If evaluated outside the grid range (e.g. at 42), it seems to automatically extrapolate
```
sitp[42.0]
3.7425272463577794
```
However, that doesn't work for a vector
```
sitp[[42.0,3.5]]

BoundsError: attempt to access 20-element scale(interpolate(::Array{Float64,1}, BSpline(Linear()), OnGrid()), (1.0:2.0:39.0,)) with element type Float64 at index [[42.0,3.5]]
 in throw_boundserror(::Interpolations.ScaledInterpolation{Float64,1,Interpolations.BSplineInterpolation{Float64,1,Array{Float64,1},Interpolations.BSpline{Interpolations.Linear},Interpolations.OnGrid,0},Interpolations.BSpline{Interpolations.Linear},Interpolations.OnGrid,Tuple{FloatRange{Float64}}}, ::Tuple{Array{Float64,1}}) at abstractarray.jl:363
 in checkbounds at abstractarray.jl:292 [inlined]
 in _getindex at multidimensional.jl:272 [inlined]
 in getindex(::Interpolations.ScaledInterpolation{Float64,1,Interpolations.BSplineInterpolation{Float64,1,Array{Float64,1},Interpolations.BSpline{Interpolations.Linear},Interpolations.OnGrid,0},Interpolations.BSpline{Interpolations.Linear},Interpolations.OnGrid,Tuple{FloatRange{Float64}}}, ::Array{Float64,1}) at abstractarray.jl:760
 in include_string(::String, ::String) at loading.jl:441
 in include_string(::String, ::String) at sys.dylib:?
 in eval(::Module, ::Any) at boot.jl:234
 in eval(::Module, ::Any) at sys.dylib:?
 in (::Atom.##65#68)() at eval.jl:102
 in withpath(::Atom.##65#68, ::Void) at utils.jl:30
 in withpath(::Function, ::Void) at eval.jl:38
 in macro expansion at eval.jl:101 [inlined]
 in (::Atom.##64#67{Dict{String,Any}})() at task.jl:60

```

What works is to set the extrapolation explicitely
```
esitp = extrapolate(sitp,Linear())
esitp[[42.0,3.5]]
2-element Array{Float64,1}:
 3.74253
 1.22632
```

This seems a bit inconsistent. Would it make sense to also automatically extrapolate when the input is a vector of points? Possibly related is issue #129.

General information
```
Pkg.status("Interpolations")
- Interpolations                0.6.1

Julia Version 0.5.3-pre+0
Commit 42f8035ccc (2017-05-06 16:34 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin16.5.0)
  CPU: Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT NO_AFFINITY NEHALEM)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, broadwell)
```

### LINKED ISSUE #165: Julia 0.6 breaks scaling with LinSpace grid?
Hello, all! I upgraded to the latest version and now some code is complaining about my use of `LinSpace`. Am I doing something wrong? The error I get is:

```julia
ERROR: MethodError: no method matching coordlookup(::LinSpace{Float64}, ::Float64)
Closest candidates are:
  coordlookup(::UnitRange, ::Any) at /home/gmihalache/.julia/v0.6/Interpolations/src/scaling/scaling.jl:71
  coordlookup(::StepRange, ::Any) at /home/gmihalache/.julia/v0.6/Interpolations/src/scaling/scaling.jl:73
  coordlookup(::StepRangeLen, ::Any) at /home/gmihalache/.julia/v0.6/Interpolations/src/scaling/scaling.jl:76
Stacktrace:
```
Thank you!

### LINKED ISSUE #191: Functions with Arrays of AbstractInterpolations cannot be handled by ForwardDiff
I've discovered an issue when passing Arrays of AbstractInterpolations to functions and then using ForwardDiff to process those functions -- *even if the Interpolations are never used.* Oddly, the error only occurs if you call the ForwardDiff package from within a function. A workaround is possible if you know the precise type of Interpolation you're going to have ahead of time or if you can pass an Array{Any}.

I'm not sure if this is a bug in Interpolations or ForwardDiff.

Code:
```julia
println("Start.")

using ForwardDiff
using Interpolations

function eff(p, thing)
    out = p[1] *  p[2]
    return out
end

function GradientWrapper(func, p, thing)
    f(x) = func(x, thing)
    return ForwardDiff.gradient(f,p)
end

n = 40
z = [i * j for i in 1:n, j in 1:n]
itp = interpolate(z, BSpline(Cubic(Natural())), OnGrid())


p = [5.0, 6.0]

test = [itp, itp]

println("Using type $(typeof(test))")
println("Calling the derivative directly:")
f(x) = eff(x,test)
println(ForwardDiff.gradient(f, p))
println("Calling the wrapper:")
println(GradientWrapper(eff, p, test))

test = Array{Any}(2)
test[1] = itp
test[2] = itp

println("Using type $(typeof(test))")
println("Calling the derivative directly:")
g(x) = eff(x, test)
println(ForwardDiff.gradient(g, p))
println("Calling the wrapper:")
println(GradientWrapper(eff, p, test))


test = Array{AbstractInterpolation}(2)
test[1] = itp
test[2] = itp

println("Using type $(typeof(test))")
println("Calling the derivative directly:")
h(x) = eff(x, test)
println(ForwardDiff.gradient(h, p))
println("Calling the wrapper -- this fails:")
println(GradientWrapper(eff, p, test))

println("Done.")
```

Output:
```
Start.
Using type Array{Interpolations.BSplineInterpolation{Float64,2,Array{Float64,2},Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}},Interpolations.OnGrid,1},1}
Calling the derivative directly:
[6.0, 5.0]
Calling the wrapper:
[6.0, 5.0]
Using type Array{Any,1}
Calling the derivative directly:
[6.0, 5.0]
Calling the wrapper:
[6.0, 5.0]
Using type Array{Interpolations.AbstractInterpolation,1}
Calling the derivative directly:
[6.0, 5.0]
Calling the wrapper -- this fails:
LoadError: MethodError: no method matching string(::Expr)
The applicable method may be too new: running in world age 2758, while current world is 21890.
Closest candidates are:
  string(::Any...) at strings/io.jl:120 (method too new to be called from this world context.)
  string(!Matched::BigInt) at gmp.jl:568 (method too new to be called from this world context.)
  string(!Matched::BigFloat) at mpfr.jl:885 (method too new to be called from this world context.)
  ...
while loading evensmallertest.jl, in expression starting on line 53
in GradientWrapper at evensmallertest.jl:13
in typeinf_ext at base/inference.jl:2622
in typeinf_code at base/inference.jl:2583
in typeinf_frame at base/inference.jl:2504
in typeinf at base/inference.jl:2787
in typeinf_work at base/inference.jl:2722
in abstract_eval at base/inference.jl:1950
in abstract_eval_call at base/inference.jl:1927
in abstract_call at base/inference.jl:1897
in abstract_call_gf_by_type at base/inference.jl:1420
in typeinf_edge at base/inference.jl:2535
in typeinf at base/inference.jl:2787
in typeinf_work at base/inference.jl:2722
in abstract_eval at base/inference.jl:1950
in abstract_eval_call at base/inference.jl:1927
in abstract_call at base/inference.jl:1897
in abstract_call_gf_by_type at base/inference.jl:1420
in typeinf_edge at base/inference.jl:2535
in typeinf at base/inference.jl:2787
in typeinf_work at base/inference.jl:2669
in abstract_interpret at base/inference.jl:2076
in abstract_eval at base/inference.jl:1950
in abstract_eval_call at base/inference.jl:1927
in abstract_call at base/inference.jl:1897
in abstract_call_gf_by_type at base/inference.jl:1420
in typeinf_edge at base/inference.jl:2535
in typeinf at base/inference.jl:2787
in typeinf_work at base/inference.jl:2669
in abstract_interpret at base/inference.jl:2076
in abstract_eval at base/inference.jl:1950
in abstract_eval_call at base/inference.jl:1927
in abstract_call at base/inference.jl:1701
in builtin_tfunction at base/inference.jl:1133
in builtin_tfunction at base/inference.jl:1215
in  at base/inference.jl:899
in getfield_tfunc at base/inference.jl:897
in limit_type_depth at base/inference.jl:692
```

### LINKED ISSUE #198: Calling with function syntax?

### LINKED ISSUE #200: Scaled interpolations failing on vector with grid points below 1.0
Here is a minimum working example.

using Interpolations
grid = linspace(0, 2.0, 10)
values = [-exp(-x) for x in grid]
itp = interpolate(values, BSpline(Cubic(Line())), OnGrid())
sitp = scale(itp, grid)

sitp[0.0]
-0.9999999999999999

sitp[[0.0,0.0]]
ERROR: BoundsError: attempt to access 10-element scale(interpolate(::Array{Float64,1}, BSpline(Cubic(Line())), OnGrid()), (0.0:0.2222222222222222:2.0,)) with element type Float64 at index [[0.0, 0.0]]
Stacktrace:
 [1] throw_boundserror(::Interpolations.ScaledInterpolation{Float64,1,Interpolations.BSplineInterpolation{Float64,1,Array{Float64,1},Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}},Interpolations.OnGrid,1},Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}},Interpolations.OnGrid,Tuple{StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}}}, ::Tuple{Array{Float64,1}}) at ./abstractarray.jl:434
 [2] checkbounds at ./abstractarray.jl:362 [inlined]
 [3] macro expansion at ./multidimensional.jl:494 [inlined]
 [4] _getindex at ./multidimensional.jl:491 [inlined]
 [5] getindex(::Interpolations.ScaledInterpolation{Float64,1,Interpolations.BSplineInterpolation{Float64,1,Array{Float64,1},Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}},Interpolations.OnGrid,1},Interpolations.BSpline{Interpolations.Cubic{Interpolations.Line}},Interpolations.OnGrid,Tuple{StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}}}, ::Array{Float64,1}) at ./abstractarray.jl:883

But the following works:
sitp[1.0]
-0.36789030239286824

sitp[[1.0, 1.0]]
2-element Array{Float64,1}:
 -0.36789
 -0.36789

### LINKED ISSUE #202: Gradients of 3D extrapolations not working.
The following minimal working example fails for me very cryptically:

```
using Interpolations

xs = 1:10
itp = Interpolations.scale(Interpolations.interpolate(zeros(10),BSpline(Cubic(Periodic())), OnGrid()),xs)
extp = Interpolations.extrapolate(itp,Periodic())
gradient(extp,2.,2.)

```
I get an error beginning with
```
ERROR: LoadError: UndefVarError: d not defined
```
Do extrapolation objects support computing gradients?

### LINKED ISSUE #212: Deprecation of array indexing calls `[]`
As suggested in https://github.com/JuliaMath/Interpolations.jl/pull/206#issuecomment-405056722, I propose the following procedural changes to deprecate interpolation evaluation by array indexing `[]`:

- [x] Support interpolation evaluation by function calls `()` (complete from https://github.com/JuliaMath/Interpolations.jl/pull/206)
- [x] Modify`README.md` to let users know that `[]` is deprecated in favour of `()` (complete from https://github.com/JuliaMath/Interpolations.jl/pull/206)
- [ ] Start throwing deprecation warnings for evaluation with `[]`
- [ ] Modify unit tests and files in `doc` directory that evaluate interpolation objects by `[]` to `()` 
- [ ] Modify any script in package that evaluates interpolation objects by `[]` to `()`
- [ ] (After tagging a major release) remove `[]` from the code base entirely

The first two steps will be completed as soon as the pull request https://github.com/JuliaMath/Interpolations.jl/pull/206 is approved and changes are committed accordingly. Completing the rest of the procedure should not be a complex task.

### LINKED ISSUE #213: Interaction with ForwardDiff.jl
I would like to use automatic differentiation on a function that includes an interpolation step. Unfortunately, Interpolations does not seem to play well with ForwardDiff. A (non-sensical) minimum working example is

```
using ForwardDiff, Interpolations

function itp_test(x)
    A_x = collect(1.0:2.0:40.0)
    A = A_x.^2
    knots = (A_x,)
    itp = interpolate(knots, A, Gridded(Linear()))
    return itp[x]
end

j = ForwardDiff.gradient(itp_test, [2.0])

ERROR: MethodError: Cannot `convert` an object of type ForwardDiff.Dual{ForwardDiff.Tag{#itp_test,Float64},Float64,1} to an object oftype Float64
This may have arisen from a call to the constructor Float64(...),
since type constructors fall back to convert methods.
Stacktrace:
 [1] macro expansion at [...]/.julia/v0.6/Interpolations/src/gridded/indexing.jl:56 [inlined]
 [2] getindex!(::Array{Float64,1}, ::Interpolations.GriddedInterpolation{Float64,1,Float64,Interpolations.Gridded{Interpolations.Linear},Tuple{Array{Float64,1}},0}, ::Array{ForwardDiff.Dual{ForwardDiff.Tag{#itp_test,Float64},Float64,1},1}) at [...]/.julia/v0.6/Interpolations/src/gridded/indexing.jl:43
 [3] itp_test(::Array{ForwardDiff.Dual{ForwardDiff.Tag{#itp_test,Float64},Float64,1},1}) at [...]/scratch2.jl:62
 [4] vector_mode_gradient at [...]/.julia/v0.6/ForwardDiff/src/gradient.jl:96 [inlined]
 [5] gradient(::Function, ::Array{Float64,1}, ::ForwardDiff.GradientConfig{ForwardDiff.Tag{#itp_test,Float64},Float64,1,Array{ForwardDiff.Dual{ForwardDiff.Tag{#itp_test,Float64},Float64,1},1}}, ::Val{true}) at [...]/.julia/v0.6/ForwardDiff/src/gradient.jl:17
 [6] gradient(::Function, ::Array{Float64,1}, ::ForwardDiff.GradientConfig{ForwardDiff.Tag{#itp_test,Float64},Float64,1,Array{ForwardDiff.Dual{ForwardDiff.Tag{#itp_test,Float64},Float64,1},1}}) at [...]/.julia/v0.6/ForwardDiff/src/gradient.jl:15 (repeats 2 times)
 [7] macro expansion at [...]/.julia/v0.6/Atom/src/repl.jl:118 [inlined]
 [8] anonymous at ./<missing>:?
```

I guess it has something to do with limitation (4) listed a http://www.juliadiff.org/ForwardDiff.jl/stable/user/limitations.html but I was wondering if there is a fix to Interpolations.jl that could allow us to use ForwardDiff with it. 

There has also been a previous closed issue #155 in which @timholy conjectured that this issue could be fixed but the issue remained dormant. Please close this issue if you think it more appropriate to continue in the existing issue.

Thanks a lot for your help.

### LINKED ISSUE #228: Confused about the true meaning of OnCell and OnGrid
This may sound surprising, but I'm looking for clarification about what OnCell and OnGrid actually mean. Let's take the case of `Constant`: my interpretation is that `OnCell` *should* mean that the array value is constant over `i-0.5` to `i+0.5` but has a step discontinuity at the edges of this region. Conversely, I would expect `OnGrid` to be constant over `i` to `i+(1-eps())`. In other words, something like this:

constant-desired

But that's not what we have, and indeed `OnGrid` and `OnCell` seem to behave identically:

constant_real

Likewise for Quadratic:

quadratic

By the definition I am picturing, I'm not even sure there is an obvious meaning to `Linear`, `OnCell` that is also continuous. (The closest I can come up with is that the slope at `i` is `(y[i+1]-y[i-1])/2`, but that has discontinuities at the half-integers.)

So despite having worked with the package a lot (though not for quite a while, truth be told), I am confused.

timholy:
Here's ~~an early preview of where I'm heading with my attempted~~ a rewrite. Some of the goals are laid out in #224. Despite adding more tests and more docstrings, ~~so far~~ this is ~~a tiny bit~~ shorter than its predecessor, which may be a good sign.

A couple things I'm particularly happy about:
- I got rid of most of the generated functions. These were necessary when we wrote the package, but since then Julia tuples have gotten so fast you can do everything that way. In general things just work better with ordinary functions (e.g., world age issues, tracking with Revise, etc.)
- while there is still some complexity in the "generic" code, the customization that has to be done for a given interpolation mode (e.g., `BSpline(Quadratic(BC()))`) is considerably simpler and smaller than before. This may help attract user/developers.
- I'm using the axes of the returned object to indicate the region at which we "exactly" (up to numerical precision) reconstruct the input array. If you use `interpolate`, this always agrees with the parent, but with `interpolate!` the result may lose its edges depending on the interpolation order and choice of boundary conditions. I am pleased that we can now signal to the user where the values can plausibly be trusted; formerly that was only something gurus knew.
- This turns on bounds-checking for `interpolate`; use `extrapolate` if you want to evaluate out-of-bounds.
- Long ago (#74) we switched our API from types to values, with the major advantage that, in principle, it allows one to store data related to the interpolation mode or boundary conditions. However, because internally most of those settings were tossed and all choices were represented only via the type system, there was no way we could exploit this possibility. Consequently, I've switched to storing the initialization settings as fields in the `BSplineInterpolation` object. The good news is that when they are all empty, there is no storage overhead:

```julia
julia> struct Empty end

julia> struct Wrapper1
           x::Float64
       end

julia> struct Wrapper2{E}
           x::Float64
           y::E
       end

julia> wx = Wrapper1(3.2)
Wrapper1(3.2)

julia> wy = Wrapper2(3.2, Empty())
Wrapper2{Empty}(3.2, Empty())

julia> sizeof(wx)
8

julia> sizeof(wy)
8

julia> struct NotEmpty
           z::UInt16
       end

julia> wz = Wrapper2(3.2, NotEmpty(2))
Wrapper2{NotEmpty}(3.2, NotEmpty(0x0002))

julia> sizeof(wz)
16
```

~~There are a couple of broken tests, and I've temporarily commented out `Gridded`, `Scaled`, and extrapolated objects. These will be addressed soon. But this should give people a chance to start looking it over and deciding whether they like it or not.~~

Left to do:

- [X] add some hessian tests, and errors for anything that doesn't actually work
- [X] reimplement extrapolation
- [x] reimplement scale
- [x] reimplement gridded

Related issues:

- fixes #228 
- fixes #213
- fixes #212 
- fixes #202 
- fixes #200
- fixes #198
- fixes #191 
- fixes #165 
- fixes #158 
- fixes #151 
- fixes #127
- fixes #122


src/b-splines/constant.jl
```diff
@@ -6,48 +6,16 @@ return `A[round(Int,x)]` when interpolating
...
-    if (length(offsets) < N)
-        d = length(offsets)+1
-        sym = Symbol("c_", d)
-        return :($sym * $(index_gen(IT, N, offsets..., 0)))
-    else
-        indices = [offsetsym(offsets[d], d) for d = 1:N]
-        return :(itp.coefs[$(indices...)])
-    end
-end
+padded_axis(ax::AbstractUnitRange, ::BSpline{Constant}) = ax
```
tlycken: ❗️ 

This is friggin' awesome. You are a genious. You just took the most complicated parts of this library to get right, and made them into "copy-from-wikipedia".

Have I told you how much I love the code you write?

src/b-splines/cubic.jl
```diff
@@ -24,145 +24,56 @@ When we derive boundary conditions we will use derivatives `y_0'(x)` and
...
 
-"""
-In `hessian_coefficients` for a cubic b-spline we assume that `fx_d = x-ix_d`
-and we define `cX_d` for `X ⋹ {m, _, p, pp}` such that
-
-    cm_d  = p''(fx_d)
-    c_d   = q''(fx_d)
-    cp_d  = q''(1-fx_d)
-    cpp_d = p''(1-fx_d)
+hessian_weights(::Cubic, δx) = (1-δx, 3*δx-2, 3*(1-δx)-2, δx)
```
tlycken: I think there might be bugs here (and we know from #122 and #181 that there are bugs in the Hessian coefficients as they stood before). These _should_ just be the derivatives of the gradient coefficients, but as far as I can tell they aren't here: the third weight expression should be `3δx - 5`.

As we don't have any tests for Hessians this is difficult to verify, so leave that change out of this PR if you wish. But it's _really_ nice that the mathematics is so much easier to check here, because the code is much closer to the results of the derivation. Well done!

timholy: I think I'm going to add tests for Hessians and then insert deliberate `throw`s into code-paths that don't work, just to prevent users from relying on stuff that's broken. The error message will include a plea for help :smile:.

I've noticed many of our tests follow a general pattern, so I'm going to rework our tests so that there's a common core of test utilities. That should make it much easier to add the Hessian tests. Probably the very next priority.

timholy:
Discussed at JuliaCon with @ChrisRackauckas (now is the time to complain about anything else!)

src/b-splines/cubic.jl
timholy: I may also ultimately fix the Hessians myself, but once the rewrite is done I have higher priorities (other packages to upgrade to 1.0) since I don't use the hessian code myself.

pkofod:
Nobody's going to challenge your word, but do you have benchmarks to back up that it is faster? :)

timholy:
> Nobody's going to challenge your word, but do you have benchmarks to back up that it is faster? :)

You should! I actually don't yet know that it will be. But I did a couple of small tests before on simple problems (e.g., 1d linear interpolation) and was able to get better generated code than that produced by Interpolations. I don't think it will be a big difference.

But truly, my only goal is to make sure I don't (appreciably) slow anything down. If I get approximate performance parity and complete feature parity (fixing a whole bunch of bugs in the process), with more maintainable code, I'll consider it a win.

pkofod:
> But truly, my only goal is to make sure I don't (appreciably) slow anything down. If I get approximate performance parity and complete feature parity (fixing a whole bunch of bugs in the process), with more maintainable code, I'll consider it a win.

A big win, I agree! I have always had a hard time reading the internals of this package, and I really appreciate what you're doing here! I just saw in the associated issue that you said it'd be faster :)

timholy:
OK, this is now extremely close: technically the only thing holding it up is [an inference failure](https://github.com/JuliaLang/julia/issues/29117), which triggers failures in the new `test/core.jl` and in `test/b-splines/constant.jl`. To make sure this doesn't get merged prematurely here's the remaining list:

- [ ] fix or work around https://github.com/JuliaLang/julia/issues/29117
- [ ] run the benchmarks and fix any serious regressions that are discovered

In the OP I've updated the list of bugs that this fixes, it's getting to be quite a long list!

timholy:
At long last, this is ready to merge. Here are the benchmark results (old version execution time is along the horizontal axis, new version execution time is along the vertical axis, and the dashed line represents equal performance; points below the line indicate improved performance, and points above the line represent worsened performance):

usage

construction

You can see we're considerably faster on construction and basically on par or better in terms of run-time performance.

From a user perspective, the main thing to note is that this implements https://github.com/JuliaMath/Interpolations.jl/issues/228#issuecomment-415446012, meaning that now one uses

```julia
interpolate(A, BSpline(Quadratic(Line(OnCell()))))
```

rather than

```julia
interpolate(A, BSpline(Quadratic(Line())), OnCell())
```

The reason is that the old API (and accompanying docs) really weren't correct, see https://github.com/JuliaMath/Interpolations.jl/issues/228. This implementation makes it clearer that all `OnGrid` and `OnCell` do is change the interpretation of the boundary condition. Note that since `Constant` and `Linear` don't need a boundary condition, you should no longer supply `OnGrid` or `OnCell` for them.

lstagner:
I just tried out this branch. Great work. I have a question though. I noticed I can still use the getfield syntax `itp[xval]` if `xval` is an integer. Is this required for your implementation. I was hoping we could use that syntax as a way to access the parent array e.g. `itp[i] = parent(itp)[i]` that way the `itp` could act both as an array and as a interpolating function.

timholy:
It's useful to have these still be `AbstractArrays` and therefore we have to support indexing with integers. 

Keep in mind that the coefficients array may not match the input, if prefiltering was used (quadratic and cubic). Consequently `itp[i]` for integer `i` still performs the calculations of interpolation so as to "undo" the prefiltering. Is that what you're asking?

timholy:
Ah, yes, you're commenting on how it behaves when you have knots (that's an important detail, since `BSpline` works as-is). Basically you're asking for the separation in #227. That's not fully part of this, though the part you're asking for isn't that many additional lines of code. Doable, in other words, though perhaps worth a separate PR.

PR #231, opened 2018-09-20
Add NEWS about release 0.9

timholy:
This may help other packages in their efforts to upgrade.


timholy:
Since this only affects the README I'll merge without running the tests.

Issue #227, opened 2018-08-23
Enhancement proposal: restructure package around knot-vectors and their indices


timholy:
Plugging away at #226 has been the occasion for reconsidering the overall structure of the package. I've been slowly evolving towards one significant internal API change (introducing `WeightedIndex`) which hopefully you'll see soon as part of #226, so I won't bother describing it here. Here, I want to focus on a much more visible form of restructuring that may not be part of #226 but which I think is worth considering. This proposal stems from our accumulated experience thinking about indexing, arrays, and other container types elsewhere in the Julia ecosystem (CCing @mbauman and @andyferris, who have helped shape these views).

We currently have 3 interpolation types: `BSpline`, `Scaled`, and `Gridded`. (We also have extrapolation types, but that's orthogonal to the content of this proposal.) The core point of this proposal is that I believe we can:

- unify all 3 interpolation types into a single type (probably called `BSpline`)
- eliminate the `OnCell` vs `OnGrid` distinction
- support attractive new features like encoding position in [physical units](https://github.com/ajkeller34/Unitful.jl) (CC @ajkeller34)

by centralizing everything around the idea of `knot` vectors.

As you all know, the knots are the places where we shift the parametrization of the underlying polynomial: when you cross a knot you end up accessing different elements of the underlying coefficient array. For an array axis spanning `1:n`, currently we have the following:

- `OnGrid`: the knots are effectively positioned at `1:n`
- `OnCell`: the knots are effectively positioned at `UnitRange(0.5, n+0.5)`

Another important component is padding, which is currently expressed through a type-parameter of the appropriate `AbstractInterpolation` type. Now that we have `OffsetArrays`, a much cleaner approach (which you'll see in #226) is to use the coefficient array's axes to encode the padding: for example, quadratic interpolation requires an extra padding element on either end, and so the natural approach would be to index this axis of the coefficient array with integers that range from 0 to `n+1`, where the 0th and `n+1`st elements are the padding. By using an OffsetArray it's easy to keep everything in alignment.

[...]

PR #238, opened 2018-09-26
Monotonic interpolations

mateuszbaran:
I've prepared an implementation of monotonic interpolations (issue #105 ) based on the code posted there. A few methods are implemented, tested and documented. This code partially covers WIP PR #110 .

Is there anything I should improve?

### DIFF
--- a/src/monotonic/monotonic.jl
+++ b/src/monotonic/monotonic.jl
@@ -1,0 +1,275 @@
+
+#=
+    Interpolate using cubic Hermite splines. The breakpoints in arrays xbp and ybp are assumed to be sorted.
+    Evaluate the function in all points of the array xeval.
+    Methods:
+        "Linear"                yuck
+        "FiniteDifference"      classic cubic interpolation, no tension parameter
+                                Finite difference can overshoot for non-monotonic data
+        "Cardinal"              cubic cardinal splines, uses tension parameter which must be between [0,1]
+                                cubin cardinal splines can overshoot for non-monotonic data
+                                (increasing tension decreases overshoot)
+        "FritschCarlson"        monotonic - tangents are first initialized, then adjusted if they are not monotonic
+                                can overshoot for non-monotonic data
+        "FritschButland"        monotonic - faster algorithm (only requires one pass) but somewhat higher apparent "tension"
+        "Steffen"               monotonic - also only one pass, results usually between FritschCarlson and FritschButland
+    Sources:
+        Fritsch & Carlson (1980), "Monotone Piecewise Cubic Interpolation", doi:10.1137/0717021.
+        Fritsch & Butland (1984), "A Method for Constructing Local Monotone Piecewise Cubic Interpolants", doi:10.1137/0905021.
+        Steffen (1990), "A Simple Method for Monotonic Interpolation in One Dimension", http://adsabs.harvard.edu/abs/1990A%26A...239..443S
+
+    Implementation based on http://bl.ocks.org/niclasmattsson/7bceb05fba6c71c78d507adae3d29417
+=#
+
+export
+    MonotonicInterpolationType,
+    LinearMonotonicInterpolation,
+    FiniteDifferenceMonotonicInterpolation,
+    CardinalMonotonicInterpolation,
+    FritschCarlsonMonotonicInterpolation,
+    FritschButlandMonotonicInterpolation,
+    SteffenMonotonicInterpolation
+
+abstract type MonotonicInterpolationType <: InterpolationType end
+
+struct LinearMonotonicInterpolation <: MonotonicInterpolationType
+end
+
+struct FiniteDifferenceMonotonicInterpolation <: MonotonicInterpolationType
+end
+
+struct CardinalMonotonicInterpolation{T<:Number} <: MonotonicInterpolationType
+    tension :: T # must be in [0, 1]
+end
+
+struct FritschCarlsonMonotonicInterpolation <: MonotonicInterpolationType
+end
+
+struct FritschButlandMonotonicInterpolation <: MonotonicInterpolationType
+end
+
+struct SteffenMonotonicInterpolation <: MonotonicInterpolationType
+end
+
+struct MonotonicInterpolation{T, TCoeffs, Tel, Type<:MonotonicInterpolationType,
+    K<:AbstractVector{<:Number}} <: AbstractInterpolation{T,1,DimSpec{Type}}
+
+    it::Type
+    knots::K
+    A::Vector{Tel}
+    m::Vector{TCoeffs}
+    c::Vector{TCoeffs}
+    d::Vector{TCoeffs}
+end
+
+
+size(A::MonotonicInterpolation) = size(A.knots)
+axes(A::MonotonicInterpolation) = axes(A.knots)
+
+function MonotonicInterpolation(::Type{TWeights}, it::IType, knots::K, A::Vector{Tel},
+    m::Vector{TCoeffs}, c::Vector{TCoeffs}, d::Vector{TCoeffs}) where {TWeights, TCoeffs, Tel, IType<:MonotonicInterpolationType, K<:AbstractVector{<:Number}}
+
+    isconcretetype(IType) || error("The b-spline type must be a leaf type (was $IType)")
+    isconcretetype(TCoeffs) || warn("For performance reasons, consider using an array of a concrete type (eltype(A) == $(eltype(A)))")
+
+    cZero = zero(TWeights)
+    if isempty(A)
+        T = Base.promote_op(*, typeof(cZero), eltype(A))
+    else
+        T = typeof(cZero * first(A))
+    end
+
+    MonotonicInterpolation{T, TCoeffs, Tel, IType, K}(it, knots, A, m, c, d)
+end
+
+function interpolate(::Type{TWeights}, ::Type{TCoeffs}, knots::K,
+    A::AbstractArray{Tel,1}, it::IT) where {TWeights,TCoeffs,Tel,K<:AbstractVector{<:Number},IT<:MonotonicInterpolationType}
+
+    # first we need to determine tangents (m)
+    n = length(knots)
+    m, Δ = calcTangents(TCoeffs, knots, A, it)
+    c = Vector{TCoeffs}(undef, n-1)
+    d = Vector{TCoeffs}(undef, n-1)
+    for k ∈ 1:n-1
+        if IT == LinearMonotonicInterpolation
+            c[k] = d[k] = zero(TCoeffs)
+        else
+            xdiff = knots[k+1] - knots[k]
+            c[k] = (3*Δ[k] - 2*m[k] - m[k+1]) / xdiff
+            d[k] = (m[k] + m[k+1] - 2*Δ[k]) / (xdiff * xdiff)
+        end
+    end
+
+    MonotonicInterpolation(TWeights, it, knots, A, m, c, d)
+end
+
+function interpolate(knots::AbstractVector{<:Number}, A::AbstractArray{Tel,1},
+    it::IT) where {Tel,N,IT<:MonotonicInterpolationType}
+
+    interpolate(tweight(A), tcoef(A), knots, A, it)
+end
+
+function (itp::MonotonicInterpolation)(x::Number)
+    x >= itp.knots[1] || return
+    x <= itp.knots[end] || return
+
+    k = 1
+    n = length(itp.knots)
+    while k < n-1 && x > itp.knots[k+1]
+        k += 1
+    end
+    xdiff = x - itp.knots[k]
+    return itp.A[k] + itp.m[k]*xdiff + itp.c[k]*xdiff*xdiff + itp.d[k]*xdiff*xdiff*xdiff
+end
+
+function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
+    y::AbstractVector{Tel}, method::LinearMonotonicInterpolation) where {TCoeffs, Tel}
+
+    n = length(x)
+    Δ = Vector{TCoeffs}(undef, n-1)
+    m = Vector{TCoeffs}(undef, n)
+    for k in 1:n-1
+        Δk = (y[k+1] - y[k]) / (x[k+1] - x[k])
+        Δ[k] = Δk
+        m[k] = Δk
+    end
+    m[n] = Δ[n-1]
+    return (m, Δ)
+end
+
+function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
+    y::AbstractVector{Tel}, method::FiniteDifferenceMonotonicInterpolation) where {TCoeffs, Tel}
+
+    n = length(x)
+    Δ = Vector{TCoeffs}(undef, n-1)
+    m = Vector{TCoeffs}(undef, n)
+    for k in 1:n-1
+        Δk = (y[k+1] - y[k]) / (x[k+1] - x[k])
+        Δ[k] = Δk
+        if k == 1   # left endpoint
+            m[k] = Δk
+        else
+            m[k] = (Δ[k-1] + Δk) / 2
+        end
+    end
+    m[n] = Δ[n-1]
+    return (m, Δ)
+end
+
+function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
+    y::AbstractVector{Tel}, method::CardinalMonotonicInterpolation{T}) where {T, TCoeffs, Tel}
+
+    n = length(x)
+    Δ = Vector{TCoeffs}(undef, n-1)
+    m = Vector{TCoeffs}(undef, n)
+    for k in 1:n-1
+        Δk = (y[k+1] - y[k]) / (x[k+1] - x[k])
+        Δ[k] = Δk
+        if k == 1   # left endpoint
+            m[k] = Δk
+        else
+            m[k] = (oneunit(T) - method.tension) * (y[k+1] - y[k-1]) / (x[k+1] - x[k-1])
+        end
+    end
+    m[n] = Δ[n-1]
+    return (m, Δ)
+end
+
+function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
+    y::AbstractVector{Tel}, method::FritschCarlsonMonotonicInterpolation) where {TCoeffs, Tel}
+
+    n = length(x)
+    Δ = Vector{TCoeffs}(undef, n-1)
+    m = Vector{TCoeffs}(undef, n)
+    for k in 1:n-1
+        Δk = (y[k+1] - y[k]) / (x[k+1] - x[k])
+        Δ[k] = Δk
+        if k == 1   # left endpoint
+            m[k] = Δk
+        else
+            # If any consecutive secant lines change sign (i.e. curve changes direction), initialize the tangent to zero.
+            # This is needed to make the interpolation monotonic. Otherwise set tangent to the average of the secants.
+            if Δk <= zero(Δk)
+                m[k] = zero(TCoeffs)
+            else
+                m[k] = Δ[k-1] * (Δ[k-1] + Δk) / 2.0
+            end
+        end
+    end
+    m[n] = Δ[n-1]
+    #=
+    Fritsch & Carlson derived necessary and sufficient conditions for monotonicity in their 1980 paper. Splines will be
+    monotonic if all tangents are in a certain region of the alpha-beta plane, with alpha and beta as defined below.
+    A robust choice is to put alpha & beta within a circle around origo with radius 3. The FritschCarlson algorithm
+    makes simple initial estimates of tangents and then does another pass over data points to move any outlier tangents
+    into the monotonic region. FritschButland & Steffen algorithms make more elaborate first estimates of tangents that
+    are guaranteed to lie in the monotonic region, so no second pass is necessary.
+    =#
+
+    # Second pass of FritschCarlson: adjust any non-monotonic tangents.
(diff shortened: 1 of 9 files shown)


src/monotonic/monotonic.jl
```diff
@@ -0,0 +1,351 @@
...
+    return k
+end
+
+function (itp::MonotonicInterpolation)(x::Number)
+    k = findKnot(itp, x)
+    xdiff = x - itp.knots[k]
+    return itp.A[k] + itp.m[k]*xdiff + itp.c[k]*xdiff*xdiff + itp.d[k]*xdiff*xdiff*xdiff
+end
+
+function derivative(itp::MonotonicInterpolation, x::Number)
```
tlycken: Why is this function not named `gradient` (returning an array of length 1) or `gradient1` (returning a scalar)?

src/monotonic/monotonic.jl
```diff
@@ -0,0 +1,347 @@
...
+    x <= itp.knots[end] || error("Given number $x is outside of interpolated range [$(itp.knots[1]), $(itp.knots[end])]")
+    k = searchsortedfirst(itp.knots, x)
+    if k > 1
+        k -= 1
+    end
+    xdiff = x - itp.knots[k]
+    return itp.A[k] + itp.m[k]*xdiff + itp.c[k]*xdiff*xdiff + itp.d[k]*xdiff*xdiff*xdiff
+end
+
+function derivative(itp::MonotonicInterpolation, x::Number)
```
tlycken: Why is this function not named `gradient` (returning an array of length 1) or `gradient1` (returning a scalar)?

mateuszbaran:
Thanks! I still don't understand well how things work in this library, so I'm glad this code is good enough :).

src/monotonic/monotonic.jl
```diff
@@ -0,0 +1,347 @@
...
+Only one pass, results usually between FritschCarlson and FritschButland.
+"""
+struct SteffenMonotonicInterpolation <: MonotonicInterpolationType
+end
+
+"""
+    MonotonicInterpolation
+
+Monotonic interpolation up to third order represented by type, knots and
+coefficients. Type is any concrete subtype of `MonotonicInterpolationType`.
```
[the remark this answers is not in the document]
mateuszbaran: I've removed that sentence and renamed `K` to `TKnots`,

src/monotonic/monotonic.jl
```diff
@@ -0,0 +1,347 @@
...
+    xdiff = x - itp.knots[k]
+    return itp.m[k] + 2*itp.c[k]*xdiff + 3*itp.d[k]*xdiff*xdiff
+end
+
+@inline function check_monotonic(knots, A)
+    axes(knots) == axes(A) || throw(DimensionMismatch("knot vector must have the same axes as the corresponding array"))
+    issorted(knots) || error("knot-vector must be sorted in increasing order")
+end
+
+function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
```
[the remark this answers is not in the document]
mateuszbaran: This code was taken almost straight from the implementation provided by niclasmattsson in discussion for issue #105 . I haven't really looked into the details except for the interpretation of coefficients `A`, `m`, `c` and `d`. I've added some documentation for these vectors in a new commit.

src/monotonic/monotonic.jl
mateuszbaran: I've used the convention from [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/src/derivative.jl) but `gradient1` seems like a more fitting choice. Should I write a new method for the `gradient` function as well (receiving and returning one element arrays)?

tlycken: The behavior of `gradient` should, in order to match how other interpolation schemes work, be

    function gradient(itp, x...)
        g = Vector{eltype(itp)}(undef, lenght(x))
        # fill g with values
        g
    end

I think (correct me if I'm wrong) that the monotonic implementation here only supports 1D interpolations, which means this would effectively be `gradient(itp, x) = [derivative(itp, x)]`. `gradient1` is just for convenience when the user knows it's a 1D interpolation object, but it can also avoid the array allocation: `gradient1(itp, x) = derivative(itp, x)`. In other words, if you rename `derivative` to `gradient1` and then add `gradient(itp, x) = [gradient1(itp, x)]` (with typed arguments) you'll be good.

As of #240 there's a fallback defined for `gradient!` that just calls `gradient`.

timholy: But `gradient` should return an `SVector` rather than `Vector`---there's a ton of overhead for `Vector` but essentially none for `SVector`.

mateuszbaran:
Thank you, these are great suggestions. I'll push a new version soon.
I have one question: the `getindex` function, the way it's defined in the library, does not make much sense here (the nodes are not located at subsequent integers). What should I do about it?

src/monotonic/monotonic.jl
```diff
@@ -0,0 +1,347 @@
...
+end
+
+
+size(A::MonotonicInterpolation) = size(A.knots)
+axes(A::MonotonicInterpolation) = axes(A.knots)
+
+function MonotonicInterpolation(::Type{TWeights}, it::IType, knots::K, A::AbstractArray{Tel,1},
+    m::Vector{TCoeffs}, c::Vector{TCoeffs}, d::Vector{TCoeffs}) where {TWeights, TCoeffs, Tel, IType<:MonotonicInterpolationType, K<:AbstractVector{<:Number}}
+
+    isconcretetype(IType) || error("The b-spline type must be a leaf type (was $IType)")
```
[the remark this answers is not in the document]
mateuszbaran: Right, of course, I've fixed it.

src/monotonic/monotonic.jl
mateuszbaran: I've renamed `derivative` to `gradient1` and introduced new functions: `gradient`, `hessian` and `hessian1`. `hessian` and `gradient` return `SVector`s.

src/monotonic/monotonic.jl
```diff
@@ -180,7 +180,12 @@ function (itp::MonotonicInterpolation)(x::Number)
     return itp.A[k] + itp.m[k]*xdiff + itp.c[k]*xdiff*xdiff + itp.d[k]*xdiff*xdiff*xdiff
 end
 
-function derivative(itp::MonotonicInterpolation, x::Number)
+function gradient(itp::MonotonicInterpolation, x::AbstractArray{<:Number, 1})
```
tlycken: Nice! This is _almost_ what we want, but the type of `x` should probably be `VarArg{<:Number,1}` - see e.g. [how it's done for `Gridded`](https://github.com/JuliaMath/Interpolations.jl/blob/master/src/gridded/indexing.jl#L11).

src/monotonic/monotonic.jl
```diff
@@ -180,7 +180,12 @@ function (itp::MonotonicInterpolation)(x::Number)
...
+    length(x) == 1 || error("Given vector x ($x) should have exactly one element")
```
tlycken: With `x::VarArg{<:Number,1}`, I don't think this check is needed (but I'm not sure).

src/monotonic/monotonic.jl
```diff
@@ -189,13 +194,27 @@ function derivative(itp::MonotonicInterpolation, x::Number)
     return itp.m[k] + 2*itp.c[k]*xdiff + 3*itp.d[k]*xdiff*xdiff
 end
 
+function hessian(itp::MonotonicInterpolation, x::AbstractArray{<:Number, 1})
+    length(x) == 1 || error("Given vector x ($x) should have exactly one element")
```
tlycken: Same thing here: `x` should be `VarArg`, and the length check is then probably redundant.

src/monotonic/monotonic.jl
mateuszbaran: I've just checked it and it looks like `function gradient(itp::MonotonicInterpolation, x::Number)` works. I think `gradient` isn't really a vararg function when it expects exactly two arguments :)

I tried vararg with hessian first and with `function hessian(itp::MonotonicInterpolation, x::Vararg{<:Number,1})` I got stack overflow due to [this function](https://github.com/JuliaMath/Interpolations.jl/blob/67049ef476b5ff848c781dc039e316070e2d6718/src/Interpolations.jl#L357) but with `function hessian(itp::MonotonicInterpolation, x::Number)` it just works.

src/monotonic/monotonic.jl
mateuszbaran: This check isn't needed with `function gradient(itp::MonotonicInterpolation, x::Number)` and it works.

tlycken: Yeah; for monotonic interpolations it doesn't matter very much.

My main concern here is that I want the interface to match `gradient` for other interpolation types _exactly_, with the only exception being the type of `itp`. The idea is that you should be able to have code that initializes an interpolation object of some type, and then (perhaps a lot later in the code) use any part of the common API without caring which type you used. So if you e.g. start out with regular B-splines, and then realize you want monotonic splines, you switch out the initialization part _and nothing else_.

It might be that `x::Number` accomplishes that well enough, and if so it's totally fine by me. I just want to make sure that the signature of this method (as well as the others, e.g. `itp(<whatever>)` and `hessian(itp, <whatever>)`) is designed with the common API in mind.

mateuszbaran:
Thank you too for a great library!

Before merging, I'd still like to know what should I do about `getindex` because right now printing an object of type `MonotonicInterpolation` may results in an error due to evaluation beyond interpolated range. `Gridded` just prints `#undef` in such cases but I don't see how it's done with evaluation outside interpolated range still throwing an error.

src/monotonic/monotonic.jl
mateuszbaran: Yes, I understand, but `Vararg{<:Number,1}` didn't work (I wrote about it here: https://github.com/JuliaMath/Interpolations.jl/pull/238#discussion_r221170008 ). `gradient(itp, [1.0])` still works (`UnexpandedIndexTypes` union contains `AbstractVector`). Is there another use case I should keep in mind?

tlycken: Hum. I'll ping in @timholy here...

tlycken:
@timholy Two outstanding questions here:

1. How should `getindex` handle out-of-bounds evaluation?

2. What's the proper way to declare `gradient`'s signature here?

src/monotonic/monotonic.jl
timholy: `gradient` is a varargs function, because the notation `args::Vararg{T,N}` means "supply `N` separate arguments of type `T`." But in your case since all `MonotonicInterpolation`s are 1-dimensional, this is equivalent to `gradient(itp::MonotonicInterpolation, x::Number)`.

`x::Vararg{Number,1}` should be the same as `x::Number`, but `Vararg{<:Number,1}` may not be. 

n-dimensional positions are encoded as varargs; like the rest of julia, indexing with vectors means something like this:

```julia
julia> A = reshape(1:20, 4, 5)
4×5 reshape(::UnitRange{Int64}, 4, 5) with eltype Int64:
 1  5   9  13  17
 2  6  10  14  18
 3  7  11  15  19
 4  8  12  16  20

julia> A[2:3, 3:4]
2×2 Array{Int64,2}:
 10  14
 11  15
```

This is *not* what you mean here, so definitely get rid of the `AbstractArray`.

mateuszbaran: I'll change `x::AbstractArray{<Number,1}` to `x::Number` then. In this case `x::Vararg{Number,1}` causes a stack overflow while `x::Number` does not:
```
Test threw exception
  Expression: ≈((Interpolations.hessian(itp, [t]))[1], hessval, atol=1.0e-12)
  StackOverflowError:
  Stacktrace:
   [1] hessian(::Interpolations.MonotonicInterpolation{Float64,Float64,Float64,FiniteDifferenceMonotonicInterpolation,Array{Float64,1},Array{Float64,1}}, ::Array{Float64,1}) at /home/mateusz/.julia/dev/Interpolations/src/Interpolations.jl:359 (repeats 80000 times)
```

mateuszbaran: My mistake, I checked again and using `x::Number` leads to the same stack overflow error. I don't know what should I do.

timholy: I just checked out the branch in this PR, and with this diff:
```diff
diff --git a/src/monotonic/monotonic.jl b/src/monotonic/monotonic.jl
index 28d1377..83e2d2d 100644
--- a/src/monotonic/monotonic.jl
+++ b/src/monotonic/monotonic.jl
@@ -180,9 +180,8 @@ function (itp::MonotonicInterpolation)(x::Number)
     return itp.A[k] + itp.m[k]*xdiff + itp.c[k]*xdiff*xdiff + itp.d[k]*xdiff*xdiff*xdiff
 end
 
-function gradient(itp::MonotonicInterpolation, x::AbstractArray{<:Number, 1})
-    length(x) == 1 || error("Given vector x ($x) should have exactly one element")
-    return SVector(gradient1(itp, x[1]))
+function gradient(itp::MonotonicInterpolation, x::Number)
+    return SVector(gradient1(itp, x))
 end
```
I get this:
```julia
julia> using Interpolations
[ Info: Recompiling stale cache file /home/tim/.julia/compiled/v1.0/Interpolations/VpKVx.ji for Interpolations [a98d9a8b-a2ab-59e6-89dd-64a1c18fca59]

julia> x = 0.0:0.1:1.0
0.0:0.1:1.0

julia> y = x.^2;

julia> itp = interpolate(x, y, LinearMonotonicInterpolation());

julia> itp(0.3)
0.09

julia> Interpolations.gradient(itp, 0.3)
1-element StaticArrays.SArray{Tuple{1},Float64,1,1}:
 0.5
```
No StackOverflow.

This one works fine too (not that I recommend it):
```julia
diff --git a/src/monotonic/monotonic.jl b/src/monotonic/monotonic.jl
index 28d1377..5352034 100644
--- a/src/monotonic/monotonic.jl
+++ b/src/monotonic/monotonic.jl
@@ -180,9 +180,8 @@ function (itp::MonotonicInterpolation)(x::Number)
     return itp.A[k] + itp.m[k]*xdiff + itp.c[k]*xdiff*xdiff + itp.d[k]*xdiff*xdiff*xdiff
 end
 
-function gradient(itp::MonotonicInterpolation, x::AbstractArray{<:Number, 1})
-    length(x) == 1 || error("Given vector x ($x) should have exactly one element")
-    return SVector(gradient1(itp, x[1]))
+function gradient(itp::MonotonicInterpolation, x::Vararg{Number,1})
+    return SVector(gradient1(itp, x...))
 end
```

Which version of Julia are you running?

mateuszbaran: Yes, `Interpolations.gradient(itp, 0.3)` works, but `Interpolations.gradient(itp, [0.3])` and `Interpolations.hessian(itp, [0.3])` do not:
```
julia> using Interpolations
[ Info: Recompiling stale cache file /home/mateusz/.julia/compiled/v1.0/Interpolations/VpKVx.ji for Interpolations [a98d9a8b-a2ab-59e6-89dd-64a1c18fca59]

julia> itp = interpolate(0.0:0.1:1.0, 0.0:0.1:1.0, LinearMonotonicInterpolation());

julia> itp(0.0)
0.0

julia> itp(0.2)
0.2

julia> Interpolations.gradient(itp, 0.3)
1-element StaticArrays.SArray{Tuple{1},Float64,1,1}:
 1.0

julia> Interpolations.gradient(itp, [0.3])
ERROR: Given number 2 is outside of interpolated range [0.0, 1.0]
Stacktrace:
 [1] (::Interpolations.MonotonicInterpolation{Float64,Float64,Float64,LinearMonotonicInterpolation,StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}},StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}})(::Int64) at /home/mateusz/.julia/dev/Interpolations/src/monotonic/monotonic.jl:174
 [2] getindex at /home/mateusz/.julia/dev/Interpolations/src/Interpolations.jl:383 [inlined]
 [3] isassigned(::Interpolations.MonotonicInterpolation{Float64,Float64,Float64,LinearMonotonicInterpolation,StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}},StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}}, ::Int64) at ./abstractarray.jl:351
 [4] show_delim_array(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Interpolations.MonotonicInterpolation{Float64,Float64,Float64,LinearMonotonicInterpolation,StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}},StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}}, ::Char, ::String, ::Char, ::Bool, ::Int64, ::Int64) at ./show.jl:659
 [5] show_delim_array at ./show.jl:649 [inlined]
 [6] show_vector(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Interpolations.MonotonicInterpolation{Float64,Float64,Float64,LinearMonotonicInterpolation,StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}},StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}}, ::Char, ::Char) at ./arrayshow.jl:442
 [7] show_vector at ./arrayshow.jl:432 [inlined]
 [8] show at ./arrayshow.jl:418 [inlined]
 [9] print(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Interpolations.MonotonicInterpolation{Float64,Float64,Float64,LinearMonotonicInterpolation,StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}},StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}}) at ./strings/io.jl:31
 [10] #print_to_string#326(::Nothing, ::Function, ::String, ::Vararg{Any,N} where N) at ./strings/io.jl:124
 [11] print_to_string at ./strings/io.jl:112 [inlined]
 [12] string at ./strings/io.jl:143 [inlined]
 [13] gradient(::Interpolations.MonotonicInterpolation{Float64,Float64,Float64,LinearMonotonicInterpolation,StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}},StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}}, ::Array{Float64,1}) at /home/mateusz/.julia/dev/Interpolations/src/Interpolations.jl:349
 [14] top-level scope at none:0

julia> Interpolations.hessian(itp, [0.3])
ERROR: StackOverflowError:
Stacktrace:
 [1] hessian(::Interpolations.MonotonicInterpolation{Float64,Float64,Float64,LinearMonotonicInterpolation,StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}},StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}}, ::Array{Float64,1}) at /home/mateusz/.julia/dev/Interpolations/src/Interpolations.jl:359 (repeats80000 times)

julia>
```
These error messages aren't helpful.

timholy: I just pushed some fixes to your branch.

timholy:
I fixed the `getindex` problem by throwing a `BoundsError`. It's not great for indicating the natural bounds of the object, unfortunately; I'm not quite sure what to do about this, short of creating a special exception type.

timholy:
My vote for `itp[x]` is to interpret `x` with respect to the parent array indexes, see the last paragraph of #227. But I don't think we change the meaning of `[]` without a nice long deprecation period.

tlycken:
Ah, true. I filed #242 to continue this discussion - it's a little OT here...

src/monotonic/monotonic.jl
mateuszbaran: OK, so I'll just remove the failing tests for gradient/hessian with array as an argument.

test/gradient.jl
```diff
@@ -194,7 +194,6 @@ using Test, Interpolations, DualNumbers, LinearAlgebra
                 for t in grid
                     gradval = epsilon(itp(dual(t, 1.0)))
                     @test Interpolations.gradient1(itp, t) ≈ gradval atol = 1.e-12
-                    @test Interpolations.gradient(itp, [t])[1] ≈ gradval atol = 1.e-12
```
timholy: Rather than remove these can you change `[t]` to `t`?

timholy:
Once those gradient/hessian tests are re-inserted without the `[]` then I approve merging this. Thanks in advance, @mateuszbaran!

mateuszbaran:
Tests are re-inserted. Thank you for help :+1: .

timholy:
Just to make sure you understand @mateusbaran, in ForwardDiff a multidimensional position is encoded by a vector but here it is encoded by varargs. This package follows indexing rules where vectors along each dimension means a rectangular sub-region of an array, i.e. the Cartesian product of all positions along each supplied axis.

Whether we stick with that will depend on whether we continue to make these subtypes of AbstractArray.

mateuszbaran:
I understand now. I didn't pay attention to it earlier because I've used this library only for one-dimensional interpolation.

timholy:
Merged as #243. Thanks for the great contribution @mateuszbaran!

tlycken:
Great work here, @mateuszbaran. Big thanks!

PR #247, opened 2018-10-01
Update README.md

yakir12:
Added some info about how to deal with `missing`. If this is cool by you keep it, if not let's let people know how to deal with missing data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,16 @@ NoInterp
 whereby the coordinate of the selected input vector MUST be located on a grid point. Requests for off grid
 coordinates results in the throwing of an error.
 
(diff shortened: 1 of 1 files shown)


timholy:
I think this is a good idea. I'm just wondering if we can phrase it in such a way as to acknowledge the aim of your "workaround": that you don't want `missing` to appear in the interpolated result. (You can just do `itp = interpolate(A, BSpline(Linear()))` but some values will return `missing`.)

yakir12:
I could change it to:
> `missing` data will naturally propagate through the interpolation,
> where some values will become `missing`. To avoid that, one can 
> filter out the missing data points and use a gridded interpolation. 
> For example:
...

PR #267, opened 2018-11-13
test and fix iteration error of ScaledIterator

kdw503:
This fixes errors during the iteration of ScaledIterator object.
Test bench is also included.
Without fixes this causes 'ErrorException("cannot reduce an empty list")' and 'MethodError(Interpolations.ScaledIterState....))'.

### DIFF
--- a/test/scaling/scaling.jl
+++ b/test/scaling/scaling.jl
@@ -78,3 +78,9 @@ using Test, LinearAlgebra
     bar!(rbar, sitp)
     @test rfoo ≈ rbar
 end
+
+@testset "eachvalue iteration" begin
+    A = reshape(reinterpret(SVector{2,Float64}, collect(1.:12.)), 3, 2)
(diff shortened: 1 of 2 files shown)


test/scaling/scaling.jl
```diff
@@ -78,3 +78,9 @@ using Test, LinearAlgebra
     bar!(rbar, sitp)
     @test rfoo ≈ rbar
 end
+
+@testset "eachvalue iteration" begin
+    A = reshape(reinterpret(SVector{2,Float64}, collect(1.:12.)), 3, 2)
+    sitp = scale(interpolate(A, BSpline(Cubic(Line(OnGrid())))), 1:3, 1:2)
+    first(eachvalue(sitp))
```
timholy: Can you change this to
```julia
@test first(eachvalue(sitp)) ≈ @SVector [1.,2.]
```
(When possible, it's a good idea to test for correctness in addition to just whether something runs.)

kdw503: Ok, I did.

timholy:
Thanks so much!

PR #269, opened 2018-11-16
[RFC] Add Hessians for ScaledInterpolation and tests

### LINKED ISSUE #268: Hessian of CubicSplineInterpolation broken?
The following
```
using Interpolations
A = rand(20, 10)
xs = 1:20
ys = 1:10
f = CubicSplineInterpolation((xs,ys), A)
Interpolations.hessian(f, 2.5, 6.7)
```
throws
```
ERROR: StackOverflowError:
[1] hessian(::Interpolations.Extrapolation{Float64,2,ScaledInterpolation{Float64,2,Interpolations.BSplineInterpolation{Float64,2,OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}},BSpline{Cubic{Line{OnGrid}}},Tuple{Base.OneTo{Int64},Base.OneTo{Int64}}},BSpline{Cubic{Line{OnGrid}}},Tuple{UnitRange{Int64},UnitRange{Int64}}},BSpline{Cubic{Line{OnGrid}}},Throw{Nothing}}, ::Float64, ::Float64) 
at /Users/karrasch/.julia/packages/Interpolations/wGqjI/src/Interpolations.jl:371 (repeats 80000 times)
```

dkarrasch:
This adds `hessian` functionality to `ScaledInterpolation`s. The inference issue seems to be due to `symmatrix`, as far as I can tell, but inference hasn't been tested on `hessian` for `BSplineInterpolation` either, so that may not be too bad.

The current implementation avoids unnecessary (by symmetry) rescaling operations and allocation of arrays (as would have been necessary in any recursive implementation I could come up with).

EDIT: (sort of) fixes #268.

### DIFF
--- a/src/extrapolation/extrapolation.jl
+++ b/src/extrapolation/extrapolation.jl
@@ -67,6 +67,25 @@ end
     end
 end
 
+@inline function hessian(etp::AbstractExtrapolation{T,N}, x::Vararg{Number,N}) where {T,N}
+    itp = parent(etp)
+    if checkbounds(Bool, itp, x...)
+        hessian(itp, x...)
+    else
+        error("extrapolation of hessian not yet implemented")
+        # # copied from gradient above, with obvious modifications
+        # # but final part is missing
+        # eflag = tcollect(etpflag, etp)
+        # xs = inbounds_position(eflag, bounds(itp), x, etp, x)
+        # h = @inbounds hessian(itp, xs...)
+        # skipni = t->skip_flagged_nointerp(itp, t)
+        # # not sure if it should be just h here instead of Tuple(h)
+        # # extrapolate_hessian needs to be written
+        # # SVector is likely also wrong here
+        # SVector(extrapolate_hessian.(skipni(eflag), skipni(x), skipni(xs), Tuple(h)))
+    end
+end
+
 checkbounds(::Bool, ::AbstractExtrapolation, I...) = true
 
 # The last two arguments are just for error-reporting
--- a/src/scaling/scaling.jl
+++ b/src/scaling/scaling.jl
@@ -126,6 +126,34 @@ rescale_gradient(r::UnitRange, g) = g
 Implements the chain rule dy/dx = dy/du * du/dx for use when calculating gradients with scaled interpolation objects.
 """ rescale_gradient
 
+@propagate_inbounds function hessian(sitp::ScaledInterpolation{T,N}, xs::Vararg{Number,N}) where {T,N}
+    @boundscheck (checkbounds(Bool, sitp, xs...) || Base.throw_boundserror(sitp, xs))
+    xl = maybe_clamp(sitp.itp, coordslookup(itpflag(sitp.itp), sitp.ranges, xs))
+    h = hessian(sitp.itp, xl...)
+    return symmatrix(rescale_hessian_components(itpflag(sitp.itp), sitp.ranges, Tuple(h), size(h, 1)))
+end
+
+function rescale_hessian_components(flags, ranges, h, n)
+    hs = ()
+    for i=1:n
+        if getfirst(flags) isa NoInterp
+            flags = getrest(flags)
+            ranges = Base.tail(ranges)
+        else
+            s1 = rescale_gradient_components(flags, ranges, Tuple(h[i*(n+1)-n:i*n]))
+            s2 = rescale_hessian(ranges[1], s1)
+            hs = (hs..., s2...)
+            flags = getrest(flags)
+            ranges = Base.tail(ranges)
+        end
+    end
+    return hs
+end
+
+rescale_hessian(r::StepRangeLen, h) = h ./ step(r)
+rescale_hessian(r::StepRange, h) = h ./ r.step
+rescale_hessian(r::UnitRange, h) = h
+
 ### Iteration
 
 struct ScaledIterator{SITPT,CI,WIS}
--- a/src/scaling/scaling.jl
+++ b/src/scaling/scaling.jl
@@ -130,31 +130,23 @@ Implements the chain rule dy/dx = dy/du * du/dx for use when calculating gradien
     @boundscheck (checkbounds(Bool, sitp, xs...) || Base.throw_boundserror(sitp, xs))
     xl = maybe_clamp(sitp.itp, coordslookup(itpflag(sitp.itp), sitp.ranges, xs))
     h = hessian(sitp.itp, xl...)
-    return symmatrix(rescale_hessian_components(itpflag(sitp.itp), sitp.ranges, Tuple(h), size(h, 1)))
+    return rescale_hessian_components(itpflag(sitp.itp), sitp.ranges, h)
 end
 
-function rescale_hessian_components(flags, ranges, h, n)
-    hs = ()
-    idx = 1
-    while idx <= n
-        if getfirst(flags) isa NoInterp
-            flags = getrest(flags)
-            ranges = Base.tail(ranges)
-        else
-            s1 = rescale_gradient_components(flags, ranges, Tuple(h[(idx-1)*n+idx:idx*n]))
-            s2 = rescale_hessian(ranges[1], s1)
-            hs = (hs..., s2...)
-            flags = getrest(flags)
-            ranges = Base.tail(ranges)
-            idx += 1
-        end
+function rescale_hessian_components(flags, ranges, h)
+    steps = SVector(get_steps(flags, ranges))
+    return h ./ (steps .* steps')
+end
+
+function get_steps(flags, ranges)
+    if getfirst(flags) isa NoInterp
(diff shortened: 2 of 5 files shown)


src/scaling/scaling.jl
```diff
@@ -126,6 +126,36 @@ rescale_gradient(r::UnitRange, g) = g
 Implements the chain rule dy/dx = dy/du * du/dx for use when calculating gradients with scaled interpolation objects.
 """ rescale_gradient
 
+@propagate_inbounds function hessian(sitp::ScaledInterpolation{T,N}, xs::Vararg{Number,N}) where {T,N}
+    @boundscheck (checkbounds(Bool, sitp, xs...) || Base.throw_boundserror(sitp, xs))
+    xl = maybe_clamp(sitp.itp, coordslookup(itpflag(sitp.itp), sitp.ranges, xs))
+    h = hessian(sitp.itp, xl...)
+    return symmatrix(rescale_hessian_components(itpflag(sitp.itp), sitp.ranges, Tuple(h), size(h, 1)))
```
timholy: Do you really need to use `symmatrix` again here? The calculation of the interpolation is expensive (`4^N` where `N` is the dimensionality), but the rescaling is only `N^2`. So you probably don't need to worry much about efficiency here. (The value of `symmatrix` is that you don't have to do redundant computations.)

src/scaling/scaling.jl
```diff
@@ -126,6 +126,36 @@ rescale_gradient(r::UnitRange, g) = g
...
+
+function rescale_hessian_components(flags, ranges, h, n)
+    hs = ()
+    idx = 1
+    while idx <= n
+        if getfirst(flags) isa NoInterp
+            flags = getrest(flags)
+            ranges = Base.tail(ranges)
+        else
+            s1 = rescale_gradient_components(flags, ranges, Tuple(h[(idx-1)*n+idx:idx*n]))
```
timholy: This line is presumably a major contributor to non-inferrability.

src/scaling/scaling.jl
```diff
@@ -126,6 +126,36 @@ rescale_gradient(r::UnitRange, g) = g
...
 
+@propagate_inbounds function hessian(sitp::ScaledInterpolation{T,N}, xs::Vararg{Number,N}) where {T,N}
+    @boundscheck (checkbounds(Bool, sitp, xs...) || Base.throw_boundserror(sitp, xs))
+    xl = maybe_clamp(sitp.itp, coordslookup(itpflag(sitp.itp), sitp.ranges, xs))
+    h = hessian(sitp.itp, xl...)
+    return symmatrix(rescale_hessian_components(itpflag(sitp.itp), sitp.ranges, Tuple(h), size(h, 1)))
+end
+
+function rescale_hessian_components(flags, ranges, h, n)
+    hs = ()
```
timholy: The "build-up-the-output" pattern is not usually inferrable, whereas the "process the head and then move on to the tail" pattern is. Reason: inference has to be able to prove that it will converge, and if arguments keep getting bigger then inference starts to worry that it is a non-terminating problem and will bail by returning `Any`. (Better to finish and yield a non-inferrable result than to never terminate.)

timholy:
Yes, I think that's right. And yes, you do have to pick out the `NoInterp` ones, but fortunately that's removing them from `steps` rather than `h` which is hopefully pretty straightforward.

With regards to symmetry, ooh, I see your point, there could be a roundoff issue. Would `h ./ (steps .* steps')` be safe?

dkarrasch:
The current status is as follows:

- Inference for `hessian` works whenever there is no `NoInterp` axis, but doesn't otherwise; comments are very welcome.
- I added a `hessian(::Extrapolation,...)` which works as expected when the point in question is in bounds, and throws an informative error otherwise; this fixes my original issue, since I created an `Extrapolation`object "accidentally" via the convenience function `CubicSplineInterpolation`, but did not intend to evaluate out of bounds.
- Somehow `hessian` gives poor results at the boundaries of the interval in the test; that's why the test is restricted to `xs[6:end-5]` or so. This cannot be due to the rescaling and should be investigated further, perhaps in a different issue?

src/scaling/scaling.jl
```diff
@@ -126,6 +126,28 @@ rescale_gradient(r::UnitRange, g) = g
...
+@propagate_inbounds function hessian(sitp::ScaledInterpolation{T,N}, xs::Vararg{Number,N}) where {T,N}
+    @boundscheck (checkbounds(Bool, sitp, xs...) || Base.throw_boundserror(sitp, xs))
+    xl = maybe_clamp(sitp.itp, coordslookup(itpflag(sitp.itp), sitp.ranges, xs))
+    h = hessian(sitp.itp, xl...)
+    return rescale_hessian_components(itpflag(sitp.itp), sitp.ranges, h)
+end
+
+function rescale_hessian_components(flags, ranges, h)
+    steps = SVector(get_steps(flags, ranges))
+    return h ./ (steps .* steps')
```
timholy: It turns out this is the source of non-inferrability, and it's really the fault of StaticArrays. You can replicate the problem with

```julia
using StaticArrays, Test
h = SMatrix{1,1}([1.0])
steps = SVector(1.0)
testinf(h, steps) = h ./ (steps .* steps')

julia> testinf(h, steps)
1×1 SArray{Tuple{1,1},Float64,2,1}:
 1.0

julia> @inferred testinf(h, steps)
ERROR: return type SArray{Tuple{1,1},Float64,2,1} does not match inferred return type Any
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] top-level scope at none:0
```
I can look into fixing this.

(This also implies that `NoInterp` is a red herring.)

test/scaling/nointerp.jl
```diff
@@ -20,6 +20,38 @@ using Interpolations, Test, LinearAlgebra, Random
...
+    xs = range(-pi, stop=pi, length=60)
+    A = hcat(map(f1, xs), map(f2, xs), map(f3, xs))
+    itp = interpolate(A, (BSpline(Cubic(Periodic(OnGrid()))), NoInterp()))
+    sitp = scale(itp, xs, ys)
+
+    for x in (xs[6:end-5] .+ 0.05), y in ys
+        if y in (1,2)
+            @test_broken h = @inferred(Interpolations.hessian(sitp, x, y))
+            h = Interpolations.hessian(sitp, x, y)
+            @test ≈(h[1], -f(x,y), atol=0.05)
```
timholy: This is fine as-is, but just FYI you can also write `@test h[1] ≈ -f(x,y) atol=0.05` if you like that better. (This file probably uses the "call" form for historical reasons, from before that syntax was possible, and was updated by search/replace.)

src/extrapolation/extrapolation.jl
```diff
@@ -67,6 +67,25 @@ end
     end
 end
 
+@inline function hessian(etp::AbstractExtrapolation{T,N}, x::Vararg{Number,N}) where {T,N}
```
timholy: I like this, thanks for adding it.

src/scaling/scaling.jl
timholy: Ref https://github.com/JuliaArrays/StaticArrays.jl/issues/546

test/scaling/nointerp.jl
dkarrasch: Yes, I'm familiar with this syntax, but didn't want to change style halfway. I have changed it now to this "modern" form for better readability in the test files related to this PR.

dkarrasch:
Many thanks, @timholy. Regarding the boundary accuracy, what confused me is the fact that the true functions do match the boundary conditions of the interpolant:
```
xs = -pi:2pi/10:pi
f1(x) = sin(x)
f2(x) = cos(x)
f3(x) = sin(x) .* cos(x)
f(x,y) = y == 1 ? f1(x) : (y == 2 ? f2(x) : (y == 3 ? f3(x) : error("invalid value for y (must be 1, 2 or 3, you used $y)")))
ys = 1:3
A = hcat(map(f1, xs), map(f2, xs), map(f3, xs))
itp = interpolate(A, (BSpline(Cubic(Periodic(OnGrid()))), NoInterp()))
sitp = scale(itp, xs, ys)
```
I realized that this is a boundary-only issue after plotting the interpolation-hessian versus the analytical one. In the interior, they match really well, but close to the periodic boundary, there is some ridiculous zigzag-pattern, way off of what it should be.

dkarrasch:
Inference in the NoInterp case still fails... :-( I had a look at code_warntype, but couldn't make much sense out of it. It seems, as if the type ambiguity occurs in the `hessian(::BSpline)` call, which can't be because tests in the normal `hessian` test file confirm that the type is correctly inferred.

dkarrasch:
Hallelujah, tests pass!!! Thanks for your patience and support, @timholy.

timholy:
Thanks for a first-rate contribution, @dkarrasch!

PR #283, opened 2018-12-14
link to documentation in readme

jpfairbanks:
The developer docs link was dead so I fixed it. I also added the RTD badge to the top of the page.


timholy:
Those docs are no longer relevant, unfortunately---the internals changed completely in release 0.9.

I guess we need some Documenter docs.

timholy:
(In the meantime, the key concept behind the package is now [described here](https://github.com/JuliaMath/Interpolations.jl/blob/35de5f9bb6ef2c7516ff385402d225918eabc72c/src/Interpolations.jl#L145-L187).)

tomasaschan:
@timholy Started reading those docs out of curiosity, and stumbled a little on [this line](https://github.com/JuliaMath/Interpolations.jl/blob/35de5f9bb6ef2c7516ff385402d225918eabc72c/src/Interpolations.jl#L155). Are you sure it's not supposed to say `ret = (1-f)*a[i] + f*a[i+1]` (note the `+1` on the second index)? Also, it might be nice to explain that `f` is in `[0,1]`, and that, if the interpolated function is `g`, then `g(x) \approx ret` when `x = i + f`.

timholy:
Yeah, that's a typo, good catch.

Issue #273, opened 2018-11-22
Add documentation (e.g. docstrings)


dkarrasch:
I realized that currently there is no documentation of any interpolation options. There is only the README on the github page, which refers to the PDF containing some of the derivations, but this seems to be incomplete in terms of documenting options, and pointing out some specifics of the options if applicable. So we could expand on some options that require additional information (like `Periodic()` BCs) in the README, or start writing docstrings. Any opinions?

This came up in #272, and it may after all affect only that `Periodic()` case, I don't have an overview if it's worth it writing docstrings for other options.

tomasaschan:
Hm. I know I wrote lots of docstrings at one point - but it looks like most of them were dropped in #224... 🤔

Re-adding them (well, the relevant parts) would of course be a welcome contribution, but I don't blame you if it feels overwhelming 😄

timholy:
Despite heroic efforts in the past, documentation is still a weak point of the package. I think the docstrings deleted in #226 were no longer relevant (e.g., they described the generated-function based API), and on balance I think it added quite a few missing ones: as just one example, before #226 `interpolate` itself didn't have a docstring. But more would be better.

timholy:
While on topic of the boundary-condition options, another thing that I think is largely missing (?) is tests that show the boundary conditions actually do what we claim they to. I thought about adding those in #226, but it was such a big change that at a certain point it became all I could do just to get that over the finish line and then clean up the (many) messes it caused.

Crown421:
I would like to follow up here, and mention that this package could use a "Getting started" section. 
The General Usage section did not give me enough information to apply the package, and the "Interpolation Algorithms" section starts with some code containing an `a` and an `A` that I can't see defined. 
From what I understand so far, the array being given is assumed to be the function outputs on a uniformly spaced [0,1] grid, but I can't find an explicit mention of this. 

It also possible that I am wrong, and just haven't found the right passage, but I just wanted to give some feedback as someone who just wanted to start interpolating. 
Maybe once I have gotten it to work and understand why, I can submit a PR for the documentation, but until then I wanted to leave it here.

timholy:
It would be a huge contribution to give this package some serious documentation.

PR #285, opened 2018-12-17
Add Documenter built docs

felixcremer:
This PR sets up Documenter built docs. These are at the moment visible at 
https://felixcremer.github.io/Interpolations.jl/dev/
I split the README into small .md files in the docs/ folder and added these into the documentation. 
I also included the docstrings at the end under Library splitted into Public and Internal API. 
See also #273 for a discussion on the docstrings. 

I think the docs need further work, and this could be a good starting point to see what docstrings are already there. 
Somebody with a better understanding of the internals would need to check which parts of the doc/ folder is still relevant and need to be included.


docs/make.jl
```diff
@@ -0,0 +1,13 @@
+using Documenter, Interpolations
+makedocs(
+sitename="Interpolations.jl",
+modules=[Interpolations],
```
timholy: Also add

```julia
format=Documenter.HTML(prettyurls = get(ENV, "CI", nothing)=="true"),
```

so that you can build and inspect documentation locally.

docs/make.jl
```diff
@@ -0,0 +1,13 @@
...
+pages=["Home" => "index.md",
+        "General usage" => "interpolations.md",
+        "Interpolation algorithms" => "control.md",
+        "Extrapolation" => "extrapolation.md",
+        "Convenience Construcors" => "convenience-construction.md",
+        "Library" => "api.md"]
+)
+
+deploydocs(repo="github.com/felixcremer/Interpolations.jl")
```
timholy: I've done the necessary key setup for JuliaMath/Interpolations.jl, so you should be able to change this to point to this repo.

docs/src/interpolations.md
```diff
@@ -0,0 +1,65 @@
...
+    end
+end
+```
+would store the on-grid value at each grid point of `itp` in the output `dest`.
+Finally, courtesy of Julia's indexing rules, you can also use
+```julia
+fine = itp(range(1,stop=10,length=1001), range(1,stop=15,length=201))
+```
+
+There is also an abbreviated notion [described below](convenience-notation.md).
```
timholy: Does this actually work? Maybe get rid of the `.md`?

felixcremer: This didn't work good catch. I had to use the @ref machinery of Documenter. It is now working locally.

docs/make.jl
felixcremer: Changed. I had to change the Documenter version to 0.21 to use this syntax.

docs/make.jl
felixcremer: Changed. The documentation will only be built when the changes are merged, because the docs deployment is pointing on the master branch.

felixcremer:
Thanks for the comments, I think this is now good to merge.

Issue #284, opened 2018-12-16
Building from coefficients


nlw0:
Probably just an issue with my own understanding of things, but I would like to build a `BSplineInterpolation` object straight from coefficients, and I don't know what I'm missing. Is it even possible? What's the story with this first nameless parameter?

This is what I'm trying:

```
itp = interpolate(randn(5,5), BSpline(Cubic(Line(OnGrid()))))
itq = Interpolations.BSplineInterpolation(zeros(7,7), itp.parentaxes, itp.it)
```

Result:

> julia> itp = interpolate(randn(5,5), BSpline(Cubic(Line(OnGrid()))))
> 5×5 interpolate(OffsetArray(::Array{Float64,2}, 0:6, 0:6), BSpline(Cubic(Line(OnGrid())))) with element type Float64:
>   0.566049   1.95704   -2.06191     -1.08552    0.854445
>   0.243544   0.296544   0.00928176  -0.332793  -1.18356 
>   0.289149  -1.45364   -1.0586       1.81307    0.360731
>   0.858396  -0.58452   -0.0484449   -1.18305    1.4658  
>  -0.548378   0.898356   0.101132     0.226077   0.213026
> 
> julia> itq = Interpolations.BSplineInterpolation(zeros(7,7), itp.parentaxes, itp.it)
> ERROR: MethodError: no method matching Interpolations.BSplineInterpolation(::Array{Float64,2}, ::Tuple{Base.OneTo{Int64},Base.OneTo{Int64}}, ::BSpline{Cubic{Line{OnGrid}}})
> Closest candidates are:
>   Interpolations.BSplineInterpolation(::Type{TWeights<:Real}, ::AbstractArray{Tel,N}, ::IT<:Union{NoInterp, Tuple{Vararg{Union{NoInterp, BSpline},N} where N}, BSpline}, ::Any) where {N, Tel, TWeights<:Real, IT<:Union{NoInterp, Tuple{Vararg{Union{NoInterp, BSpline},N} where N}, BSpline}} at /home/user/.julia/packages/Interpolations/1CM1Q/src/b-splines/b-splines.jl:48
> Stacktrace:
>  [1] top-level scope at none:0
>

timholy:
Looks like we're missing a constructor method that fills in all the type parameters. Care to submit a PR?

For the record the following works:

```julia
julia> BS = typeof(itp)
[...]

PR #286, opened 2018-12-17
BSplineInterpolation constructors that take the coefficient matrix

nlw0:
Patch for my own feature request #284

### DIFF
--- a/src/b-splines/b-splines.jl
+++ b/src/b-splines/b-splines.jl
@@ -62,6 +62,16 @@ function BSplineInterpolation(::Type{TWeights}, A::AbstractArray{Tel,N}, it::IT,
     BSplineInterpolation{T,N,typeof(A),IT,typeof(axs)}(A, fix_axis.(axs), it)
 end
 
+function BSplineInterpolation(A::AbstractArray{Tel,N}, it::IT, axs) where {N,Tel,IT<:DimSpec{BSpline}}
+    axspad = padded_axes(axs, it)
+    oa = OffsetArray(A, axspad)
+    BSplineInterpolation(tweight(A), oa, it, axs)
+end
+
+function BSplineInterpolation(A::OffsetArray{Tel,N}, it::IT, axs) where {N,Tel,IT<:DimSpec{BSpline}}
+    BSplineInterpolation(tweight(A), A, it, axs)
+end
+
 iscomplete(its::Tuple) = all(iscomplete, its)
 
 coefficients(itp::BSplineInterpolation) = itp.coefs
--- a/src/b-splines/b-splines.jl
+++ b/src/b-splines/b-splines.jl
@@ -62,14 +62,14 @@ function BSplineInterpolation(::Type{TWeights}, A::AbstractArray{Tel,N}, it::IT,
     BSplineInterpolation{T,N,typeof(A),IT,typeof(axs)}(A, fix_axis.(axs), it)
 end
 
-function BSplineInterpolation(A::AbstractArray{Tel,N}, it::IT, axs) where {N,Tel,IT<:DimSpec{BSpline}}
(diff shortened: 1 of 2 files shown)


timholy:
Looks good. Can you add a test? As future developers update the package, having a test will ensure this won't silently break---if it causes your test(s) to fail, it will force him/her to figure out how to fix the problem before continuing.

nlw0:
I can sure try, but I'm a total Julia newbie and appreciate any guidance.

Does the code look good? I'm not even familiar with using OffsetArrays, for instance. Could it be better if I find a way to define another method instead of having these multiple class constructors?

test/b-splines/cubic.jl
```diff
@@ -68,4 +68,8 @@
     itp_flat_c = interpolate(A, BSpline(Cubic(Flat(OnCell()))))
     @test Interpolations.gradient(itp_flat_c,0.5)[1] ≈ 0 atol=eps()
     @test Interpolations.gradient(itp_flat_c,ix[end] + 0.5)[1] ≈ 0 atol=eps()
+
+    # Can construct from coefficients
+    itp = interpolate(rand(5, 9), BSpline(Cubic(Flat(OnGrid()))))
+    @test itp == Interpolations.BSplineInterpolation(itp.coefs, itp.it, itp.parentaxes)
```
timholy: Need to test the non-OffsetArrays constructor too. (If it doesn't work as intended it's better not to have it at all.)

timholy:
> Does the code look really good? I'm not even familiar with using OffsetArrays, for instance. Could it be better if I find a way to define another method instead of having these multiple class constructors?

For an intro to OffsetArrays, see [this blog post](https://julialang.org/blog/2017/04/offset-arrays). It was written for an older version of Julia, so some info might be a bit outdated but it should give you the main points.

nlw0:
I think the second constructor might be handy... I included a test for it, and also refactored the code a little bit

nlw0:
I take that back, let's keep it simple

nlw0:
@timholy Maybe the function could actually even take an AbstractArray instead? It's really just about avoiding the type argument, and any errors related to non-matching array ranges are just the user's fault here...

src/b-splines/b-splines.jl
```diff
@@ -62,6 +62,10 @@ function BSplineInterpolation(::Type{TWeights}, A::AbstractArray{Tel,N}, it::IT,
     BSplineInterpolation{T,N,typeof(A),IT,typeof(axs)}(A, fix_axis.(axs), it)
 end
 
+function BSplineInterpolation(A::AbstractArray{Tel,N}, it::IT, axs) where {N,Tel,IT<:DimSpec{BSpline}}
```
timholy: We need to do some sanity-checking:
```suggestion
function BSplineInterpolation(A::AbstractArray{Tel,N}, it::IT, axs) where {N,Tel,IT<:DimSpec{BSpline}}
    @noinline err_axes(A, it, axs) = throw(ArgumentError("parentaxes $axs are inconsistent with coefficient array axes $(axes(A)) and interpolation type $it"))

    paxs = padded_axes(axs, it)
    all(map(⊆, paxs, axes(A))) || err_axes(A, it, axs)
```

test/b-splines/cubic.jl
timholy:
```suggestion
    @test itp == Interpolations.BSplineInterpolation(itp.coefs, itp.it, itp.parentaxes)
    @test_throws ArgumentError Interpolations.BSplineInterpolation(itp.coefs, itp.it, (2:3, 2:10))
```

nlw0:
Great, thanks!

timholy:
Thanks @nlw0!

PR #290, opened 2018-12-21
Require that singleton dimensions be NoInterp. Fixes #289.

### LINKED ISSUE #289: Potential memory corruption: `extrapolate`  produces `NaN` at random
The `extrapolate` function produces `NaN` at random. In the example below, I occasionally get `NaN` which evidently I should not. If you have trouble reproducing this then increase the upper bound of the loop. However, I obtain the `NaN` relatively frequently. 

```julia
Using Interpolations, Test
```
```julia
# This test fails for at least one i.
for i = 1:10000
    block_cdfs = ones(1,3)
    itp = interpolate(view(block_cdfs,:,:), BSpline(Linear()))
    etp = extrapolate(itp, Flat())
    @test sum(isnan.(etp)) == 0
end
```
```julia
# This test fails for at least one i.
for i = 1:10000
    block_cdfs = ones(4,1)
    itp = interpolate(view(block_cdfs,:,:), BSpline(Linear()))
    etp = extrapolate(itp, Flat())
    @test sum(isnan.(etp)) == 0
end
```

```julia
# This test fails as well.
for i = 1:10000
    for p = 1:10
        for q = 1:10
            block_cdfs = ones(p,q)
            itp = interpolate(view(block_cdfs,:,:), BSpline(Linear()))
            etp = extrapolate(itp, Flat())
            @test sum(isnan.(etp)) == 0
        end
    end
end
```

```julia
Julia Version 0.7.0
Commit a4cb80f3ed (2018-08-08 06:46 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: AMD Ryzen 5 1600 Six-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.0 (ORCJIT, znver1)
Environment:
  JULIA_EDITOR = atom -a
  JULIA_NUM_THREADS = 6
```

timholy:
CC @zygmuntszpak. I wonder if this could be breaking, so I'll leave it open a while for comment.

### DIFF
--- a/src/b-splines/b-splines.jl
+++ b/src/b-splines/b-splines.jl
@@ -48,10 +48,14 @@ function BSplineInterpolation(::Type{TWeights}, A::AbstractArray{Tel,N}, it::IT,
     @noinline err_concrete(IT) = error("The b-spline type must be a concrete type (was $IT)")
     @noinline warn_concrete(A) = @warn("For performance reasons, consider using an array of a concrete type (typeof(A) == $(typeof(A)))")
     @noinline err_incomplete(it) = error("OnGrid/OnCell is not supplied for some of the interpolation modes in $it")
+    @noinline err_singleton(A, it) = throw(ArgumentError("size $(size(A)) is inconsistent with $it, use NoInterp along singleton dimensions"))
 
     isconcretetype(IT) || err_concrete(IT)
     isconcretetype(typeof(A)) || warn_concrete(A)
     iscomplete(it) || err_incomplete(it)
+
+    # Check that dimensions of size 1 are NoInterp
+    is_singleton_ok(A, it) || err_singleton(A, it)
 
     # Compute the output element type when positions have type TWeights
     if isempty(A)
@@ -71,6 +75,16 @@ function BSplineInterpolation(A::AbstractArray{Tel,N}, it::IT, axs) where {N,Tel
 end
 
 iscomplete(its::Tuple) = all(iscomplete, its)
+
+is_singleton_ok(A, it) = is_singleton_ok(size(A), it)
+@inline function is_singleton_ok(sz::Tuple, it)
+    if sz[1] == 1
+        it1 = getfirst(it)
+        it1 isa NoInterp || return false
+    end
+    return is_singleton_ok(Base.tail(sz), getrest(it))
+end
+is_singleton_ok(::Tuple{}, it) = true
 
 coefficients(itp::BSplineInterpolation) = itp.coefs
 interpdegree(itp::BSplineInterpolation) = interpdegree(itpflag(itp))
--- a/test/b-splines/linear.jl
+++ b/test/b-splines/linear.jl
@@ -52,4 +52,13 @@
     itp = interpolate(x, BSpline(Linear()))
     @test itp(1.5, CartesianIndex((2, 3))) === itp(1.5, 2, 3)
     @test itp(CartesianIndex((1, 2)), 1.5) === itp(1, 2, 1.5)
+
+    # Issue #289
+    for (x, it) in ((ones(1, 3), (NoInterp(), BSpline(Linear()))),
+                    (ones(3, 1), (BSpline(Linear()), NoInterp())))
+        @test_throws ArgumentError interpolate(x, BSpline(Linear()))
+        itp = interpolate(x, it)
+        etp = extrapolate(itp, Flat())
+        @test sum(isnan.(etp)) == 0
+    end
 end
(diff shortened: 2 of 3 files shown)


timholy:
It's worth discussing some of the subtleties here. The way this typically gets evaluated can be revealed with Rebugger (note I'm running Julia with `--check-bounds=yes`, and I'm using `master`, not this branch):

```julia
julia> a = [i for i = 1:3, j = 1:1]
3×1 Array{Int64,2}:
 1
 2
 3

julia> itp = interpolate(a, BSpline(Linear()))
3×1 interpolate(::Array{Float64,2}, BSpline(Linear())) with element type Float64:
 #undef
 #undef
 #undef

# Note: the next comes from stepping into `itp(1.1,1)`:
(itp::Interpolations.BSplineInterpolation{T,N,TCoefs,IT,Axs} where Axs<:Tuple{Vararg{AbstractUnitRange,N}} where IT<:Union{NoInterp, Tuple{Vararg{Union{NoInterp, BSpline},N} where N}, BSpline} where TCoefs<:AbstractArray)(x::Vararg{Number,N}) where {T, N} in Interpolations at /home/tim/.julia/dev/Interpolations/src/b-splines/indexing.jl:6
  itp = [#undef; #undef; #undef]
  x = (1.1, 1)
  T = Float64
  N = 2
rebug> @eval Interpolations let (itp, x, T, N) = Main.Rebugger.getstored("0a91928a-06bf-11e9-2d75-97032c03971d")
       begin
           @boundscheck checkbounds(Bool, itp, x...) || Base.throw_boundserror(itp, x)
           wis = weightedindexes((value_weights,), itpinfo(itp)..., x)
           @show wis   # this line inserted during Rebugging
           itp.coefs[wis...]
       end
       end
wis = (Interpolations.WeightedAdjIndex{2,Float64}(1, (0.8999999999999999, 0.10000000000000009)), Interpolations.WeightedAdjIndex{2,Int64}(0, (0, 1)))
ERROR: BoundsError: attempt to access 3×1 Array{Float64,2} at index [2, 0]
Stacktrace:
 [1] getindex at ./array.jl:732 [inlined]
 [2] interp_getindex at /home/tim/.julia/dev/Interpolations/src/Interpolations.jl:298 [inlined]
 [3] macro expansion at /home/tim/.julia/dev/Interpolations/src/Interpolations.jl:282 [inlined]
 [4] interp_getindex at /home/tim/.julia/dev/Interpolations/src/Interpolations.jl:273 [inlined]
 [5] macro expansion at /home/tim/.julia/dev/Interpolations/src/Interpolations.jl:282 [inlined]
 [6] interp_getindex at /home/tim/.julia/dev/Interpolations/src/Interpolations.jl:273 [inlined]
 [7] _getindex at /home/tim/.julia/dev/Interpolations/src/Interpolations.jl:238 [inlined]
 [8] getindex(::Array{Float64,2}, ::Interpolations.WeightedAdjIndex{2,Float64}, ::Interpolations.WeightedAdjIndex{2,Int64}) at ./abstractarray.jl:905
 [9] top-level scope at REBUG:6
 [10] eval(::Module, ::Any) at ./boot.jl:319
 [11] top-level scope at none:0
```

Note the index for dimension 2 is

```julia
Interpolations.WeightedAdjIndex{2,Int64}(0, (0, 1))
```
which means it will use `0*a[i,0] + 1*a[i,1]`. (In case you haven't seen it, explanation is [here](https://github.com/JuliaMath/Interpolations.jl/blob/36bd7a5ceeed95367e6d20ac5e2547c940faa189/src/Interpolations.jl#L145-L200).)
You get the NaN in #289 when whatever `a[i,0]` points to happens to be `NaN`, otherwise the value at `a[i,0]` is irrelevant since it gets multiplied by 0.

Options for fixing it:
1. check this case at runtime and return valid index objects of different type (`WeightedAdjIndex` always uses adjacent indices, and hence can't represent a single index for orders higher than `Contant`). Ideally for a singleton dimension one would return an `Int` but that would not be inferrable, and since it could happen in any dimension union-splitting would be unlikely to come to the rescue in multidimensional problems.
2. eliminate `WeightedAdjIndex` and always use `WeightedArbIndex`, in this case checking at runtime and returning `WeightedArbIndex((1,1),(0,1))` for that second dimension and thus computing `0*a[i,1] + 1*a[i,1]`. I imagine this would prevent SIMD vectorization, though I can't guarantee it's being used now everywhere it could/should be.
3. automatically filling in `Contant()` or `NoInterp()` for singleton dimensions even when the user specifies a single interpolation order for the whole array. The downside is that `interpolate` would no longer be inferrable.
4. throwing an error, like in this PR. This preserves inferrability at the cost of forcing users to jump through more hoops.

These are the only options I can think of. I suspect 2 and 4 are the most viable, but thoughts from others would be welcome.

zygmuntszpak:
Thank you for the detailed clarification and explaining the trade-off. I think option 4 is reasonable.

PR #294, opened 2019-01-13
Linrange scaling

### LINKED ISSUE #293: scaling using a LinRange unsupported
Two nearly-equivalent ranges used for scaling produce different behavior:
The one using StepRangeLen works fine.
```julia
> using Interpolations
> s= [x^2 for x in 1:5]
> ts = StepRangeLen(0, 0.5, 5)
> itp = interpolate(s, (BSpline(Cubic(Line(OnGrid())))))
> sitp = scale(itp, ts)
5-element scale(interpolate(OffsetArray(::Array{Float64,1}, 0:6), BSpline(Cubic(Line(OnGrid())))), (0.0:0.5:2.0,)) with element type Float64:
  0.9999999999999999
  4.0               
  8.999999999999998 
 15.999999999999998 
 24.999999999999996 
```

However, using LinRange produces an error

```julia
> using Interpolations
> s= [x^2 for x in 1:5]
> ts = LinRange(0, 2.0, 5)
> itp = interpolate(s, (BSpline(Cubic(Line(OnGrid())))))
> sitp = scale(itp, ts)
MethodError: no method matching coordlookup(::LinRange{Float64}, ::Float64)
Closest candidates are:
  coordlookup(!Matched::NoInterp, ::Any, !Matched::Any) at /home/danmackinlay/.julia/dev/Interpolations/src/scaling/scaling.jl:88
  coordlookup(!Matched::Interpolations.Flag, ::Any, !Matched::Any) at /home/danmackinlay/.julia/dev/Interpolations/src/scaling/scaling.jl:89
  coordlookup(!Matched::UnitRange, ::Any) at /home/danmackinlay/.julia/dev/Interpolations/src/scaling/scaling.jl:91
  ...
```

I wish to avoid StepRangeLen, as it uses TwicePrecision number internally, and these have irritating properties when using ForwardDiff. 

Possibly related: #165 

This looks fixable; in fact I'm stuck on understanding the test suite which stubbornly refuses to fail for me when I try to insert a test for this, which is probably due to a case of foolishness on my part

danmackinlay:
For this to be a sensible commit to fix #293 , my changes to the test suite should actually cause a test failure. However, they don't; I'm doing it wrong. Hints gratefully accepted.

### DIFF
--- a/src/scaling/scaling.jl
+++ b/src/scaling/scaling.jl
@@ -92,9 +92,9 @@ coordlookup(r::UnitRange, x) = x - r.start + oneunit(eltype(r))
 # coordlookup(i::Bool, r::AbstractRange, x) = i ? coordlookup(r, x) : convert(typeof(coordlookup(r,x)), x)
 coordlookup(r::StepRange, x) = (x - r.start) / r.step + oneunit(eltype(r))
 
-coordlookup(r::StepRangeLen, x) = (x - first(r)) / step(r) + oneunit(eltype(r))
-boundstep(r::StepRangeLen) = 0.5*step(r)
-rescale_gradient(r::StepRangeLen, g) = g / step(r)
+coordlookup(r::AbstractRange, x) = (x - first(r)) / step(r) + oneunit(eltype(r))
+boundstep(r::AbstractRange) = 0.5*step(r)
+rescale_gradient(r::AbstractRange, g) = g / step(r)
 
 basetype(::Type{ScaledInterpolation{T,N,ITPT,IT,RT}}) where {T,N,ITPT,IT,RT} = ITPT
 basetype(sitp::ScaledInterpolation) = basetype(typeof(sitp))
--- a/test/scaling/scaling.jl
+++ b/test/scaling/scaling.jl
@@ -11,11 +11,24 @@ using Test, LinearAlgebra, StaticArrays
     @test typeof(sitp) <: Interpolations.ScaledInterpolation
     @test parent(sitp) === itp
 
-    for (x,y) in zip(-3:.05:1.5, 1:.1:10)
+    for (x,y) in zip(-3:.05:1.5, 1:.1:10,)
         @test sitp(x) ≈ y
     end
 
     @test_throws ArgumentError scale(itp, reverse(-3:.5:1.5))
+
+    # Model linear interpolation of y = -3 + .5x by interpolating y=x
+    # and then scaling to the new x range, this time using a LinRange,
+    # which should give the same values as with the StepRangeLen used previously
+
+    itp = interpolate(1:1.0:10, BSpline(Linear()))
+
+    sitp = @inferred(scale(itp, LinRange(-3,1.5,10)))
+    @test typeof(sitp) <: Interpolations.ScaledInterpolation
+
+    for (x,y) in zip(-3:.05:1.5, 1:.1:10,)
+        @test sitp(x) ≈ y
+    end
 
     # Verify that it works in >1D, with different types of ranges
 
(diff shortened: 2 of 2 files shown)


dkarrasch:
I think tests don't fail because your very own fix from within this PR is applied. You should run your tests against master to make it fail. But I think it's a good sign that tests don't fail. That means any previously intended functionality is preserved by your modification, *and* your fix works.

howthebodyworks:
@dkarrasch that would be great - but the first commit in the PR that introduces the test but *not* the fix still causes no failures, so I haven't really covered it :disappointed:

dkarrasch:
I didn't read carefully enough. It should be
```
zip(LinRange(-3,1.5,91), 1:.1:10)
```
in the test. You missed to replace the `StepLenRange` `-3:.05:1.5` with your `LinRange`.

test/scaling/scaling.jl
```diff
@@ -11,7 +11,7 @@ using Test, LinearAlgebra, StaticArrays
     @test typeof(sitp) <: Interpolations.ScaledInterpolation
     @test parent(sitp) === itp
 
-    for (x,y) in zip(-3:.05:1.5, 1:.1:10)
+    for (x,y) in zip(-3:.05:1.5, 1:.1:10, LinRange(-3,1.5,91))
```
dkarrasch: I should use the code review functionality. 😃 This should be
```
for (x,y) in zip(LinRange(-3,1.5,91), 1:.1:10)
```

src/scaling/scaling.jl
```diff
@@ -92,9 +92,9 @@ coordlookup(r::UnitRange, x) = x - r.start + oneunit(eltype(r))
 # coordlookup(i::Bool, r::AbstractRange, x) = i ? coordlookup(r, x) : convert(typeof(coordlookup(r,x)), x)
 coordlookup(r::StepRange, x) = (x - r.start) / r.step + oneunit(eltype(r))
 
-coordlookup(r::StepRangeLen, x) = (x - first(r)) / step(r) + oneunit(eltype(r))
-boundstep(r::StepRangeLen) = 0.5*step(r)
-rescale_gradient(r::StepRangeLen, g) = g / step(r)
+coordlookup(r::AbstractRange, x) = (x - first(r)) / step(r) + oneunit(eltype(r))
+boundstep(r::AbstractRange) = 0.5*step(r)
+rescale_gradient(r::AbstractRange, g) = g / step(r)
```
dkarrasch: Are there potentially more instances that need a consistent change? Something like `rescale_hessian` for instance?

danmackinlay: Hmm. AFAICT no, `rescale_hessian_components`, for example, uses `steps`, which accepts any `AbstractRange` subtype. We should not have changed any behaviour. It presumably can't hurt to test the hessian scaling also though.

test/scaling/scaling.jl
danmackinlay: Ah. Yes, Clearly I did not read that. But we are still not there, right, because we don't want to lose the original test that works with StepRangeLen. I will need to add another one.

test/scaling/scaling.jl
```diff
@@ -11,12 +11,25 @@ using Test, LinearAlgebra, StaticArrays
...
 
     @test_throws ArgumentError scale(itp, reverse(-3:.5:1.5))
 
+    # Model linear interpolation of y = -3 + .5x by interpolating y=x
+    # and then scaling to the new x range, this time using a LinRange,
+    # which should give the same values as with the StepRangeLen used previously
+
+    itp = interpolate(1:1.0:10, BSpline(Linear()))
+
+    sitp = @inferred(scale(itp, LinRange(-3,1.5,91)))
```
danmackinlay: This duplicates most of the previous test but with a different LinRange.
possibly this is wasteful? But it does fail without the fix as desired.

src/scaling/scaling.jl
dkarrasch: Generally, I think you have found an oversight. #165 was marked as resolved by #226. So it indeed could be that this had been fixed in some places but not all.

danmackinlay: There are now tests for hessians specifically.. I'm not sure if they are useful tests.

tests/scaling.jl also tests for iteration behaviour and other float types when interpolating; but I don't think we especially want to clone those also for LinRange

danmackinlay:
I'll rebase and squash these if they make sense.

Issue #282, opened 2018-12-07
fft-based interpolations 


babaq:
matlab have `interpft` which is a fft-based interpolation, or generally speaking, a smoothing method.

hope this can be added to the package.

timholy:
Contributions are always welcome!

gdestouet:
very easy to implement for 1D real signal ( most cases ) : 
What I am using :
```
using FFTW 
function FFTW_based_interpolation(sig::AbstractArray{<:Real,1},n::Integer)
    n_orig = div(length(sig),2)
    return irfft( vcat( rfft(sig), zeros( div(n,2) - n_orig ) ) , n )
```

tomasaschan:
Is `FFTW` a heavy package? If so, it might be a bad idea to pull it in as a dependency for everyone who uses Interpolations.jl just to support this case.

What should be possible, though, is to extend `Interpolations.jl` in a separate package; if you build something that has a similar API (e.g. extending the `interpolate` function with methods that take your own signalling types, and that return your own type deriving from `AbstractInterpolation`) you can probably piggy-back on many of the constructs in the package (like extrapolation and/or scaling) without having to have the implementation inside the package. (This is analogous to how Julia code outside of Base is not second-class to Julia code inside Base, so there is no real need to have everything in Base.)

gdestouet:
Well that's just for 1D real signals, Discrete Fourier Transform based interpolation can be generalized to any class of discrete and multidimensionnal signals. 

Intermediate points are evaluated by summing cardinal sinuses weighted by the initial and given samples. The DFT interpolation just make it easy to calculate this by padding the frequency domain of the signal and taking the inverse DFT.

It is really useful in image and audio processing when you know that your original signal ( or function ) is band limited and that the sampled version respect the Nyquist-Shanon theorem of sampling ( i.e that your original function has been sampled properly ). See [this paper for example](https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=143310)

Concerning the FFTW package, it is supported by Julia-Math, weigths 11 Mo when builded, and needs the [FFTW library](http://fftw.org/). In comparison with `Interpolations.jl` it weights **a lot** for just this basic application and I understand that it is not useful right now.

I made a [PR](https://github.com/JuliaMath/Interpolations.jl/pull/300) and tested the implementation, people can check it if they want to use DFT interpolation.

gdestouet:
[...]

PR #300, opened 2019-02-07
 add fft-based interpolation for 1D real signals

gdestouet:
Simple test that I didn't included yet 

``` 
using Interpolations
N=512
new_size = 1024

sig = sin.( 2 * pi * 0.01 * N .* range(0,1,length=N) )
new_sig = interpolate(sig,FFT(new_size))
sig2 = sin.( 2 * pi * 0.01 * N .* range(0,1,length=new_size) )
```
You can check using Plots that `sig2` and `new_sig` have the same frequency

### DIFF
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -21,6 +21,7 @@ export
     InPlace,
     InPlaceQ,
     Throw,
+    FFT,
 
     LinearInterpolation,
     CubicSplineInterpolation
@@ -433,5 +434,6 @@ include("utils.jl")
 include("io.jl")
 include("convenience-constructors.jl")
 include("deprecations.jl")
+include("fftbased/with_ffts.jl")
 
 end # module
--- a/src/fftbased/with_ffts.jl
+++ b/src/fftbased/with_ffts.jl
@@ -1,0 +1,12 @@
+import FFTW: irfft,rfft
+
+struct FFT <: InterpolationType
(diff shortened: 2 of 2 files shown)


tomasaschan:
For reasons outlined in https://github.com/JuliaMath/Interpolations.jl/issues/282#issuecomment-461761255, I don't think this is a good fit inside Interpolations.jl. It's a good candidate to start building a separate package (FFTInterpolations.jl?), though!

timholy:
If Requires is acceptable, should we reopen?

timholy:
This can't be quite right, however:
```julia
julia> using Interpolations

julia> x = rand(5)
5-element Array{Float64,1}:
 0.2475749273753287 
 0.5177360746376121 
 0.4359109842436517 
 0.2480371524923537 
 0.09161945602953914

julia> y = interpolate(x, FFT(100));

julia> using Statistics

julia> mean(y)
0.015408785947784854

julia> mean(x)
0.30817571895569706
```
interpolate

gdestouet:
Yup, with this implementation you need to renormalize by N_new / N_old after the interpolation ( forgot to add this ).

Another thing : if you take random sample it will have an infinite bandwith thus you cant really interpolate with an FFT since it doesnt respect the Nyquist Shannon theroem.

For an example you can check the comment I made for the PR. You better try with a sinusoid with a frequency below the nyquist rate.

PR #324, opened 2019-05-27
Use more efficient searchsorted when inputs are sorted for Gridded

aaowens:
See https://discourse.julialang.org/t/using-interpolations/23933/31?u=aaowens

For a sorted input vector, this updates the lower bound as it moves through the list. Since the next index is likely very close to the previous one, it uses a linear search of the first 4 elements, then switches to exponential search from there.

```
using Interpolations, Dierckx, BenchmarkTools

# Generate random, non-gridded data.
# Ensure the data goes all the way to the ends of the interval, chosen as [-1000, 1000]
x = sort(rand(1000)*2000 .- 1000); x[1] = -1000; x[end] = 1000
y = rand(length(x))

# Random evaluation points
x2 = rand(1000)*2000 .- 1000
sx2 = sort(x2)
# Interpolants
fit1 = Spline1D(x,y,w=ones(length(x)), k=1, bc="nearest", s=0.0)
```

Before
```
julia> @btime $itp($x2);
  87.715 μs (9 allocations: 31.56 KiB)

julia> @btime $fit1($x2);
  254.596 μs (1 allocation: 7.94 KiB)

julia> @btime $itp($sx2);
  66.936 μs (9 allocations: 31.56 KiB)

julia> @btime $fit1($sx2);
  22.002 μs (1 allocation: 7.94 KiB)

julia> all(itp(sx2) .≈ fit1(sx2))
true
```
After
```
julia> @btime $itp($x2);
  88.359 μs (11 allocations: 47.44 KiB)

julia> @btime $fit1($x2);
  252.432 μs (1 allocation: 7.94 KiB)

julia> @btime $itp($sx2);
  15.019 μs (11 allocations: 47.44 KiB)

julia> @btime $fit1($sx2);
  21.973 μs (1 allocation: 7.94 KiB)

julia> all(itp(sx2) .≈ fit1(sx2))
true
```


src/gridded/indexing.jl
```diff
@@ -41,6 +41,7 @@ function gridded_floorbounds(x, knotvec::AbstractVector)
 end
 
 @inline find_knot_index(knotv, x) = searchsortedfirst(knotv, x, Base.Order.ForwardOrdering()) - 1
+@inline find_knot_index(knotv, x::Array) = searchsortedfirst_vec(knotv, x) .- 1
```
timholy:
```suggestion
@inline find_knot_index(knotv, x::AbstractVector) = searchsortedfirst_vec(knotv, x) .- 1
```

src/gridded/indexing.jl
```diff
@@ -49,6 +50,12 @@ end
     weightedindex(fs, degree(mode), knotvec, x, iclamp)
 end
 
+@inline function weightedindex_parts2(fs::F, mode::Gridded, knotvec::AbstractVector, x, i) where F
```
timholy: Looks like this is essentially all of the previous method except the lookup of `i`. Rather than duplicate code, I'd rather see the previous one just call this function, i.e., previous method becomes

```julia
    i = find_knot_index(knotvec, x)
    return weightedindex_parts2(fs, mode, knotvec, x, i)
```

src/gridded/indexing.jl
```diff
@@ -69,7 +76,11 @@ rescale_gridded(::typeof(hessian_weights), coefs, Δx) = coefs./Δx.^2
 @inline function (itp::GriddedInterpolation{T,N})(x::Vararg{Union{Number,AbstractVector},N}) where {T,N}
     @boundscheck (checkbounds(Bool, itp, x...) || Base.throw_boundserror(itp, x))
     itps = tcollect(itpflag, itp)
-    wis = dimension_wis(value_weights, itps, itp.knots, x)
+    if x[1] isa Vector
```
timholy:
```suggestion
    if x[1] isa AbstractVector
```

src/gridded/indexing.jl
```diff
@@ -96,3 +118,39 @@ function getindex_return_type(::Type{GriddedInterpolation{T,N,TCoefs,IT,K}}, arg
...
+        tn2m1 *= 2
+        ind = lo + tn2
+    end
+    return searchsortedfirst(v, xx, lo + tn2 - tn2m1, hi, Base.Order.Forward)
+end
+
+function searchsortedfirst_vec(v::AbstractVector, x::AbstractVector)
+    issorted(x) || return searchsortedfirst.(Ref(v), x)
+    out = zeros(Int, length(x))
+    firstlo = 1
```
timholy: Not sure what the `firstlo` and `firsthi` are doing. Could these just be `lo` and `hi`?

src/gridded/indexing.jl
```diff
@@ -96,3 +118,39 @@ function getindex_return_type(::Type{GriddedInterpolation{T,N,TCoefs,IT,K}}, arg
...
+    lo = firstlo
+    firsthi = length(v)
+    hi = firsthi
+    it = Base.OneTo(length(x))
+    @inbounds for i in it
```
timholy: Could these two lines just be `for i = 1:length(x)`?

timholy:
Thanks again for a very nice contribution!

PR #334, opened 2019-10-15
fix ==(::AbstractInterpolation, ::AbstractInterpolation)

### LINKED ISSUE #333: `==` broken
Asking whether two interpolations are equal is broken:
```julia
julia> using Interpolations
^[[A
julia> itp = interpolate(([1,1.2],), randn(2), Gridded(Linear()))
2-element interpolate((::Array{Float64,1},), ::Array{Float64,1}, Gridded(Linear())) with element type Float64:
 -1.267482014558195 
 -0.5952689972849705

julia> @which itp == itp
==(A::AbstractArray, B::AbstractArray) in Base at abstractarray.jl:1771

julia> itp == itp
ERROR: BoundsError: attempt to access 2-element interpolate((::Array{Float64,1},), ::Array{Float64,1}, Gridded(Linear())) with element type Float64 at index [2]
Stacktrace:
 [1] throw_boundserror(::Interpolations.GriddedInterpolation{Float64,1,Float64,Gridded{Linear},Tuple{Array{Float64,1}}}, ::Tuple{Int64}) at ./abstractarray.jl:538
 [2] GriddedInterpolation at /home/ptwlinuxvm/.julia/packages/Interpolations/jXoPW/src/gridded/indexing.jl:3 [inlined]
 [3] getindex at /home/ptwlinuxvm/.julia/packages/Interpolations/jXoPW/src/Interpolations.jl:403 [inlined]
 [4] iterate at ./abstractarray.jl:914 [inlined]
 [5] _zip_iterate_some at ./iterators.jl:350 [inlined]
 [6] _zip_iterate_all at ./iterators.jl:342 [inlined]
 [7] iterate at ./iterators.jl:333 [inlined]
 [8] ==(::Interpolations.GriddedInterpolation{Float64,1,Float64,Gridded{Linear},Tuple{Array{Float64,1}}}, ::Interpolations.GriddedInterpolation{Float64,1,Float64,Gridded{Linear},Tuple{Array{Float64,1}}}) at ./abstractarray.jl:1780
 [9] top-level scope at REPL[13]:1
```

jw3126:
Close #333

### DIFF
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -50,6 +50,14 @@ abstract type AbstractInterpolation{T,N,IT<:DimSpec{InterpolationType}} <: Abstr
 abstract type AbstractInterpolationWrapper{T,N,ITPT,IT} <: AbstractInterpolation{T,N,IT} end
 abstract type AbstractExtrapolation{T,N,ITPT,IT} <: AbstractInterpolationWrapper{T,N,ITPT,IT} end
 
+function Base.:(==)(itp1::AbstractInterpolation, itp2::AbstractInterpolation)
+    propertynames(itp1) == propertynames(itp2) || return false
+    for pname in propertynames(itp1)
+        getproperty(itp1, pname) == getproperty(itp2, pname) || return false
+    end
+    return true
+end
+
 """
     BoundaryCondition
 
--- a/src/b-splines/b-splines.jl
+++ b/src/b-splines/b-splines.jl
@@ -43,6 +43,13 @@ struct BSplineInterpolation{T,N,TCoefs<:AbstractArray,IT<:DimSpec{BSpline},Axs<:
     parentaxes::Axs
     it::IT
 end
+
+function Base.:(==)(o1::BSplineInterpolation, o2::BSplineInterpolation)
+    o1.it == o2.it &&
+    o1.parentaxes == o2.parentaxes &&
+    o1.coefs == o2.coefs
+end
+
 function BSplineInterpolation(::Type{TWeights}, A::AbstractArray{Tel,N}, it::IT, axs) where {N,Tel,TWeights<:Real,IT<:DimSpec{BSpline}}
     # String interpolation causes allocation, noinline avoids that unless they get called
     @noinline err_concrete(IT) = error("The b-spline type must be a concrete type (was $IT)")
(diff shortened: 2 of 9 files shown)


timholy:
That's a pretty slow way to implement `==`. Wouldn't matter for big arrays, but I worry about small ones.
```julia
itp = interpolate(rand(5, 5), BSpline(Quadratic(Flat(OnGrid()))))
coefs = copy(itp.coefs)

using BenchmarkTools

julia> @btime $itp.coefs == $coefs
  52.584 ns (0 allocations: 0 bytes)
true

julia> @btime propertynames($itp)
  183.799 ns (1 allocation: 32 bytes)
(:coefs, :parentaxes, :it)
```
It could be a fallback, of course, but it wouldn't hurt to have faster implementations for concrete types.

jw3126:
Wow I new that this implementation was slow, but I did not expect it to be that slow:
```julia
@btime $itp == $itp #   1.193 μs (5 allocations: 160 bytes)
```
Would you like the following better?
```julia
@generated function eq_gen(o1, o2)
    fieldnames(o1) == fieldnames(o1) || return :(false)
    ex = quote 
        o1 === o2 && return true
    end
    for fn in fieldnames(o1)
        push!(ex.args, :(o1.$fn == o2.$fn || return false))
    end
    push!(ex.args, :(return true))
    ex
end

using BenchmarkTools
using Interpolations
itp = interpolate(rand(5, 5), BSpline(Quadratic(Flat(OnGrid()))))
itp2 = deepcopy(itp)
@btime eq_gen($itp, $itp2) #141.398 ns
```

timholy:
Anything wrong with
```julia
==(itp1::BSplineInterpolation, itp2::BSplineInterpolation) = itp1.parentaxes == itp2.parentaxes && itp1.it == itp2.it && itp1.coefs == itp2.coefs
```
and similarly for a couple of other types? Some advantages with this explicit version:
- it allows you to short-circuit the slow comparison (check `parentaxes` and `it` first, then check `coefs` only if both of those pass)
- it plays nicely with the debugger & Revise
- it's friendlier for people who don't really grok `@generated` and metaprogramming (this would not be enough on its own, but nice nonetheless)

jw3126:
The only downside of the explicit version is, that it is more error-prone. It's easy to forget a type or worse a field. Especially if both can vary over time.

timholy:
That is an interesting point I hadn't considered. Still, one would hope that tests would catch this kind of problem. (That's been my experience thus far.)

Thanks for also fixing CI!

PR #336, opened 2019-10-22
make compatible with ForwardDiff at lower bound of extrapolation object

### LINKED ISSUE #335: Extrapolation with ForwardDiff yields incorrect gradients at lower bound
Gradients are incorrect returned from ForwardDiff when at the lower bound of an interpolation object if an extrapolation boundary condition is used. The same issue does not exist if an extrapolation boundary condition is not used. Here's my MWE:

```julia
using ForwardDiff, Interpolations

function itp_extrapolate(x)
    A_x = 1:5
    A = A_x
    itp = LinearInterpolation(A_x, A, extrapolation_bc=Line())
    return itp(x)
end

function itp_noextrapolate(x)
    A_x = 1:5
    A = A_x
    itp = LinearInterpolation(A_x, A)
    return itp(x)
end

ForwardDiff.jacobian(itp_noextrapolate, [1.0])
# 1×1 Array{Float64,2}:
# 1.0
ForwardDiff.jacobian(itp_extrapolate, [1.0])
# 1×1 Array{Float64,2}: 
# 0.0
```

Interestingly this issue only appears at the lower boundary of the interpolation object.

```julia
ForwardDiff.jacobian(itp_extrapolate, [0.0])
# 1×1 Array{Float64,2}: 
# 1.0

ForwardDiff.jacobian(itp_extrapolate, [5.0])
# 1×1 Array{Float64,2}:
# 1.0
```

This can cause issues for gradient based optimization since often design variables may be at the lower bound of their allowable range.

taylormcd:
fixes #335

### DIFF
--- a/src/extrapolation/extrapolation.jl
+++ b/src/extrapolation/extrapolation.jl
@@ -123,7 +123,7 @@ function inbounds_index((flagl,flagu)::Tuple{Nothing,Throw}, (l,u), x, etp, xN)
 end
 
 function inbounds_index((flagl,flagu)::Tuple{Union{Flat,Line},Flag}, (l,u), x, etp, xN)
-    inbounds_index((nothing,flagu), (l,u), maxp(x,l), etp, xN)
+    inbounds_index((nothing,flagu), (l,u), maxp(l,x), etp, xN)
 end
 function inbounds_index((flagl,flagu)::Tuple{Nothing,Union{Flat,Line}}, (l,u), x, etp, xN)
     minp(x,u)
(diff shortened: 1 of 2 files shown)


timholy:
Can you add some tests? And make sure both edges work?

taylormcd:
I added tests for both edges using the example I put in the issue, so I think everything should be good to go.  Let me know if I need to do anything else.  Also, thanks for your work with Julia. I see your name all over the place on packages that I use.

timholy:
This will be released as 0.12.5, https://github.com/JuliaRegistries/General/pull/4969. Thanks so much, @taylormcd!

PR #347, opened 2020-01-02
Fixed printing of monotonic interpolation objects

mateuszbaran:
I've fixed printing of monotonic interpolation objects. If they are wrapped in an extrapolation object, printing them on the master branch even causes an error.


timholy:
Thanks!

PR #348, opened 2020-01-10
Add regularization for BSpline

### LINKED ISSUE #254: [Feature request] Smoothing factor
This is mentioned in https://github.com/JuliaMath/Interpolations.jl/issues/5#issuecomment-352149843.
The feature is provided by the Dierckx.jl package and in R as mentioned in the comment, but also by scipy https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.UnivariateSpline.html
It is a very useful parameter to avoid overfitting.

getzze:
Closes #254 

Implements a regularized BSpline.
To implement a regularized spline on unequal-grid, the x-axis should be provided from the beginning, it cannot work with defining a ScaledSpline afterwards.

No tests for now and minimal documentation.


src/b-splines/prefiltering.jl
```diff
@@ -53,6 +62,53 @@ function prefilter!(
...
+#end
+
+function prefilter!(
+    ::Type{TWeights}, ret::TCoefs, it::BSpline, λ::Real, k::Int
+    ) where {TWeights,TCoefs<:AbstractArray}
+    local buf, shape, retrs
+    sz = size(ret)
+    first = true
+    if ndims(ret) > 1
+        @warn "Smooth BSpline only available for Vectors, fallback to non-smooth"
```
timholy: How about restricting `interpolate` to `AbstractVector` when `λ` and `k` are supplied?

src/b-splines/prefiltering.jl
```diff
@@ -53,6 +62,53 @@ function prefilter!(
...
+##        lmul!(D1[1+i:end, 1+i:end], D)
+#    end
+#
+#    D' * D
+#end
+
+function prefilter!(
+    ::Type{TWeights}, ret::TCoefs, it::BSpline, λ::Real, k::Int
+    ) where {TWeights,TCoefs<:AbstractArray}
+    local buf, shape, retrs
```
timholy: These don't seem to be used

src/b-splines/prefiltering.jl
```diff
@@ -53,6 +62,53 @@ function prefilter!(
...
+    sz = size(ret)
+    first = true
+    if ndims(ret) > 1
+        @warn "Smooth BSpline only available for Vectors, fallback to non-smooth"
+        prefilter!(TWeights, ret, it)
+    end
+    if λ <= 0
+        prefilter!(TWeights, ret, it)
```
timholy:
```suggestion
        return prefilter!(TWeights, ret, it)
```

src/b-splines/prefiltering.jl
```diff
@@ -53,6 +62,53 @@ function prefilter!(
...
+    end
+    M, b = prefiltering_system(TWeights, eltype(TCoefs), sz[1], degree(it))
+    ### TEST REGULARIZATION
+    n = sz[1]
+    Q = Matrix(diffop(TWeights, n, k))
```
timholy: Do you have to enforce dense here? I'm wondering what happens on a large array.

getzze: I had to transform to Dense matrix because I was getting an error running Cholesky with sparse matrices (see next line). I don't know enough about this type of optimization to solve it. But yes, there is no reason for it to be dense.

src/b-splines/prefiltering.jl
getzze: It should work for higher dimensions too, it is just not implemented yet.
When implemented, it will be easier to change then.

getzze:
I added tests, but they are maybe too minimal. I have no idea what is important to check. If somebody can comment on it that would be great.

mkitti:
This looks good to merge this now. Could we update the base branch so this can automatically merged?

getzze:
It would be good to try optimizing the code before (but I don't know enough to do it), like for instance @timholy 's comment about transforming to a dense Matrix, I think it could be avoided.

mkitti:
Tests are passing here and this is mergeable, so I'm merging for v0.13.1 prospectively.

Feel free to submit another PR for further optimization slated for v0.13.2 or earlier.

PR #353, opened 2020-01-30
document bounds

jw3126:
I think `bounds` is quite useful and should be part of the public API.

### DIFF
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -5,6 +5,7 @@ export
     interpolate!,
     extrapolate,
     scale,
+    bounds,
 
     AbstractInterpolation,
     AbstractExtrapolation,
@@ -116,7 +117,7 @@ Return the `bounds` of the domain of `itp`. This is best explained by example:
 ```jldoctest
 julia> itp = interpolate([1 2 3; 4 5 6], BSpline(Linear()));
 
-julia> Interpolations.bounds(itp)
+julia> bounds(itp)
 ((1, 2), (1, 3))
 
 julia> data = 1:3;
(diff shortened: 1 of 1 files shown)


jw3126:
Would it be okay to also export bounds?

timholy:
This seems reasonable to me, with perhaps one caveat. Suppose someday we define a `DelaunyInterpolation` type; in that case `bounds` should describe a pretty complicated region. Would we be signing ourselves up for trouble?

jw3126:
Good point. We could define bounds only in easy cases? Where the bounds can be described as a rectangular region?

src/Interpolations.jl
```diff
@@ -107,6 +107,28 @@ axes(exp::AbstractExtrapolation) = axes(exp.itp)
 
 twotuple(r::AbstractUnitRange) = (first(r), last(r))
 twotuple(x, y) = (x, y)
+
+"""
+    bounds(itp::AbstractInterpolation)
+
+Return the `bounds` of the domain of `itp`. This is best explained by example:
```
timholy:
```suggestion
Return the `bounds` of the domain of `itp` as a tuple of `(min, max)` pairs for each coordinate. This is best explained by example:
```

timholy:
In theory, adding a new export could be deemed breaking (if it conflicts with an export from another package), but I am inclined to just bump the patch version. I'll hold off for a day or so for feedback before tagging a new release.

jw3126:
Yeah, the name `bounds` probably already exists at some places already.
But one can argue its the consumers responsibility to either avoid "wildcard imports" `using SomePackage` altogether, or deal with surprising new exports from time to time.

PR #361, opened 2020-03-26
Add support for Unitful.jl  numbers to MonotonicInterpolations

MatFi:
Now one can do:
```julia
using Unitful
using Interpolations

x=-1u"s":0.25u"s":1u"s" ;
y=sign.(0.1u"s".+x)u"m";
itp = interpolate(x, y, FiniteDifferenceMonotonicInterpolation());
julia> itp(-0.125*u"s")
0.0 m
```

The reason why Unitful.jl did not work here before is that all polynomial coefficients were of the same type `TCoeffs`. When using Unitful numbers, the different dimensions of the coefficients become relevant. Therefore every coefficient now has its own type.
This pull request additionally contains a fix of a erroneous tangent calculation in FritschCarlsonMonotonicInterpolation. 
_(Be warned, I have been using julia only since version 1.0, so this may not be implemented optimally)_


timholy:
Sorry for the lack of review. If you're interested in a bigger role (see #365) I'll be happy to review this.

MatFi:
I'm honored and I'd love to, but giving a hack like me even a shred of responsibility for this package would be the end of it. I think the best way to give something back to the community is to stay on the ground where I'm reasonably proficient in what I do. Of course I like to help out where I can.

timholy:
Sounds good. I'll leave this for my successor to review as good practice! Sorry for the likely delay.

MatFi:
No problem, I'm glad to hear that this PR becomes useful even before the merge

mkitti:
Successor in training here. Could we develop some tests and add Unitful as a test dependency to ensure integration going forward? Also see #367

timholy:
Note that this does add some Unitful tests (see `xa` and `y`). Unitful is already in the test deps, see https://github.com/JuliaMath/Interpolations.jl/blob/master/Project.toml

timmyfaraday:
Is this feature available in the latest release, as I'm not able to recreate this behavior.
If not, any indication on when would it be available?

mkitti:
> Is this feature available in the latest release, as I'm not able to recreate this behavior.
> If not, any indication on when would it be available?

This branch as not been merged yet. It's a high priority at the moment since @timholy already looked at it. I'll take a look this weekend and hopefully merge within two weeks.

mkitti:
I intend to merge this and eventually release as v0.13.1

PR #315, opened 2019-05-08
A few performance improvements. Fixes #311

timholy:
See #311, https://discourse.julialang.org/t/using-interpolations/23933, and https://discourse.julialang.org/t/help-with-poor-linearinterpolation-performance/22325.

There is still some ways to go, particularly for `Gridded` (Julia's `searchsortedfirst` is not speedy), but this helps.

Issue #311, opened 2019-03-26
Performance issues with linear interpolations


angel-devicente:
Hi,

as suggested by Tim Holy in the Discourse page, I send here the details of the issue I'm having with Interpolations.jl.

I have four versions of a test function to do some interpolations:

```
function interp_test1(size)
    xs = range(1,size,length=size)
    ys = range(1,size,length=size)
    zs = range(1,size,length=size)

    xsi = xs .+ 0.5 ; ysi = ys .+ 0.5 ; zsi = zs .+ 0.5

    dat = randn(size,size,size)

    dat_out = zeros(size,size,size)
    dat_int = LinearInterpolation((xs,ys,zs),dat,extrapolation_bc = Line())  # ~23s 124M allocs 4.65 GiB                                                                                                             
    for C in CartesianIndices((length(xsi),length(ysi),length(zsi)))
        dat_out[C] = dat_int(xsi[C[1]],ysi[C[2]],zsi[C[3]])
    end

    return sum(dat_out)
end


function interp_test2(size)
    xs = range(1,size,length=size)
    ys = range(1,size,length=size)
    zs = range(1,size,length=size)

    xsi = xs .+ 0.5 ; ysi = ys .+ 0.5 ; zsi = zs .+ 0.5

    dat = randn(size,size,size)

    dat_out = zeros(size,size,size)
    dat_int = extrapolate(interpolate(dat, BSpline(Linear())), Line())        # ~16s 125M allocs 4.65 GiB                                                                                                            
    for C in CartesianIndices((length(xsi),length(ysi),length(zsi)))
        dat_out[C] = dat_int(xsi[C[1]],ysi[C[2]],zsi[C[3]])
    end

    return sum(dat_out)
end

function interp_test3(size)
    xs = range(1,size,length=size)
    ys = range(1,size,length=size)
    zs = range(1,size,length=size)

    xsi = xs .+ 0.5 ; ysi = ys .+ 0.5 ; zsi = zs .+ 0.5

    dat = randn(size,size,size)

    dat_out = zeros(size-1,size-1,size-1)
    dat_int = LinearInterpolation((xs,ys,zs),dat)     # ~20s 124M allocs 4.64 GiB                                                                                                                                    

    for C in CartesianIndices((length(xsi)-1,length(ysi)-1,length(zsi)-1))
        dat_out[C] = dat_int(xsi[C[1]],ysi[C[2]],zsi[C[3]])
    end

    return sum(dat_out)
end


[...]

PR #362, opened 2020-03-28
Performance improvements in gradient computation

timholy:
From the benchmarks in #311, this is a further improvement compared to #315:
```julia
julia> @time interp_test1(500)
  8.222822 seconds (10 allocations: 2.794 GiB, 1.44% gc time)
-5168.12456559581

julia> @time interp_test2(500)
  5.556805 seconds (10 allocations: 2.794 GiB, 2.03% gc time)
-1605.1519426096165

julia> @time interp_test3(500)
  6.571917 seconds (11 allocations: 2.788 GiB, 1.69% gc time)
-6341.612041378765

julia> @time interp_test4(500)
  4.652481 seconds (9 allocations: 2.788 GiB, 2.53% gc time)
-24121.63637598708
```
The large number of allocations has been eliminated.


timholy:
CC @ChantalJuntao

PR #367, opened 2020-05-21
Add support for Unitful.jl numbers to FilledExtrapolations

npbarnes:
I think this part was just trying to promote the type of the fillvalue to something that matches both the coordinates, x, and the coefs. It was doing it in a very strange way. Why not use the standard type promotion machinery?

### DIFF
--- a/src/extrapolation/filled.jl
+++ b/src/extrapolation/filled.jl
@@ -30,7 +30,7 @@ extrapolate(itp::AbstractInterpolation{T,N,IT}, fillvalue) where {T,N,IT} = Fill
 @inline function (etp::FilledExtrapolation{T,N})(x::Vararg{Number,N}) where {T,N}
     itp = parent(etp)
     coefs = coefficients(itp)
-    Tret = promote_type(eltype(coefs), typeof.(x)...,T)
+    Tret = promote_type(eltype(coefs), typeof.(x)...)
     if checkbounds(Bool, itp, x...)
         @inbounds itp(x...)
     else
--- a/src/extrapolation/filled.jl
+++ b/src/extrapolation/filled.jl
@@ -30,7 +30,7 @@ extrapolate(itp::AbstractInterpolation{T,N,IT}, fillvalue) where {T,N,IT} = Fill
 @inline function (etp::FilledExtrapolation{T,N})(x::Vararg{Number,N}) where {T,N}
     itp = parent(etp)
     coefs = coefficients(itp)
-    Tret = typeof(prod(x) * zero(eltype(coefs)))
+    Tret = promote_type(eltype(coefs), typeof.(x)...,T)
     if checkbounds(Bool, itp, x...)
         @inbounds itp(x...)
     else
--- a/perf/benchmarks.jl
+++ b/perf/benchmarks.jl
@@ -30,14 +30,14 @@ end
 strip_prefix(str::AbstractString) = replace(str, "Interpolations."=>"")
 benchstr(::Type{T}) where {T<:Interpolations.GridType} = strip_prefix(string(T))
 
-benchstr(::Type{Constant}) = "Constant()"
+benchstr(::Type{<:Constant}) = "Constant()"
 benchstr(::Type{Linear}) = "Linear()"
 benchstr(::Type{Quadratic{BC}}, ::Type{GT}) where {BC<:Interpolations.BoundaryCondition,GT<:Interpolations.GridType} =
     string("Quadratic(", strip_prefix(string(BC)), "(", strip_prefix(string(GT)), "()))")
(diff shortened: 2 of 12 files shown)


npbarnes:
I changed the title because I realized that I didn't explain why I made this change. I think using promote_type is the correct abstraction here. Obviously, using prod(x) made it dimensionally unsound, but I also think `prod(x)*zero` is a confusing line of code that distracts from the intent. On the other hand, promote_type is the standard way to find a minimally widened type.

I also suspect that this will improve compatibility with other exotic Number types, although I don't have an example handy.

mkitti:
Thanks for the explanation. Sorry for the delay. Could we add Unitful.jl as an extra dependency and do a few tests to ensure this functions as expected and continues to do so?

src/extrapolation/filled.jl
```diff
@@ -30,7 +30,7 @@ extrapolate(itp::AbstractInterpolation{T,N,IT}, fillvalue) where {T,N,IT} = Fill
 @inline function (etp::FilledExtrapolation{T,N})(x::Vararg{Number,N}) where {T,N}
     itp = parent(etp)
     coefs = coefficients(itp)
-    Tret = typeof(prod(x) * zero(eltype(coefs)))
+    Tret = promote_type(eltype(coefs), typeof.(x)...)
```
MatFi, line 33: `Tret = promote_type(eltype(coefs), typeof.(x)...,T)` would fix #370

mkitti:
I would like to add a test here for the issue in #370

mkitti:
Looks like the last change I made broke some tests:
```julia
ExtrapTypeStability: Error During Test at /home/travis/build/JuliaMath/Interpolations.jl/test/extrapolation/type-stability.jl:70
  Test threw exception
  Expression: #= /home/travis/build/JuliaMath/Interpolations.jl/test/extrapolation/type-stability.jl:70 =# @inferred(etp(2018)) === 2.0f0u
  return type Quantity{Float32,𝐋^2 𝐌 𝐈^-1 𝐓^-3,Unitful.FreeUnits{(V,),𝐋^2 𝐌 𝐈^-1 𝐓^-3,nothing}} does not match inferred return type Union{Quantity{Float32,𝐋^2 𝐌 𝐈^-1 𝐓^-3,FreeUnits{(V,),𝐋^2 𝐌 𝐈^-1 𝐓^-3,nothing}}, Quantity{Float64,𝐋^2 𝐌 𝐈^-1 𝐓^-3,FreeUnits{(V,),𝐋^2 𝐌 𝐈^-1 𝐓^-3,nothing}}}
  Stacktrace:
   [1] error(::String) at ./error.jl:33
   [2] macro expansion at /home/travis/build/JuliaMath/Interpolations.jl/test/extrapolation/type-stability.jl:70 [inlined]
   [3] macro expansion at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/Test/src/Test.jl:1083 [inlined]
   [4] top-level scope at /home/travis/build/JuliaMath/Interpolations.jl/test/extrapolation/type-stability.jl:6
ExtrapTypeStability: Error During Test at /home/travis/build/JuliaMath/Interpolations.jl/test/extrapolation/type-stability.jl:71
  Test threw exception
  Expression: #= /home/travis/build/JuliaMath/Interpolations.jl/test/extrapolation/type-stability.jl:71 =# @inferred(etp(2019)) === 1.0f0u
  return type Quantity{Float64,𝐋^2 𝐌 𝐈^-1 𝐓^-3,Unitful.FreeUnits{(V,),𝐋^2 𝐌 𝐈^-1 𝐓^-3,nothing}} does not match inferred return type Union{Quantity{Float32,𝐋^2 𝐌 𝐈^-1 𝐓^-3,FreeUnits{(V,),𝐋^2 𝐌 𝐈^-1 𝐓^-3,nothing}}, Quantity{Float64,𝐋^2 𝐌 𝐈^-1 𝐓^-3,FreeUnits{(V,),𝐋^2 𝐌 𝐈^-1 𝐓^-3,nothing}}}
  Stacktrace:
   [1] error(::String) at ./error.jl:33
   [2] macro expansion at /home/travis/build/JuliaMath/Interpolations.jl/test/extrapolation/type-stability.jl:71 [inlined]
   [3] macro expansion at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/Test/src/Test.jl:1083 [inlined]
   [4] top-level scope at /home/travis/build/JuliaMath/Interpolations.jl/test/extrapolation/type-stability.jl:6
```

Perhaps having a single return type is better. I'll look into further, but I'm currently inclined to change the test.

mkitti:
We'll revisit #370 at a later time.

PR #369, opened 2020-06-15
Lanczos resampling

### LINKED ISSUE #366: Lanczos Resampling
I'm interested in implementing a Lanczos resampling algorithm here. This appears to be simply the convolution of the [Lanczos Kernel](https://en.wikipedia.org/wiki/Lanczos_resampling) by the samples.

I created a fork and started developing, but I have become completely overwhelmed by the complexity inside the library. I have no idea if such an interpolator should be its own interpolation type, or if it should be a `Gridded` type. I also have no clue how to interface with, well, anything here.

Can i get some advice on how/where to implement this algorithm (and whether you want it at all!)?

mileslucas:
This PR tries to add Lanczos resampling as an interpolation method. Currently I've added some code that appears to do the interpolation correctly, but I'm failing at extrapolation and I'm honestly not too sure I've got the interface set up correctly.

Here's a MWE
```julia
a = randn(10)
itp = interpolate(a, Lanczos()) # default degree is 4
itp2 = interpolate(a, Lanczos(2))
itp(2) == itp2(2) == a[2]

X = randn(10, 10)
itp = interpolate(X, Lanczos())
itp(2, 3) == X[2, 3]
```

I would love some feedback and help, and would happily accept direct edits.

---
fixes #366


mileslucas:
So after re-reading the developer documentation I think this can be easily rewritten to use the WeightedIndex struct. Unfortunately, I'm _extremely_ lost in the actual implementation. I've been trying to read through the code to understand the control flow and figure what actual functions I need to write. This has lead me down a rabbit hole where functions start in places like `gridded/indexing.jl` which uses functions from `b-splines.jl` which calls stuff from `utils.jl`, and so on.

If there is anybody here who can explain what I actually need to implement, I'll happily write the code and docs. Everything is just so tightly coupled to the b-splines code that I'm never sure whats supposed to be generic and what is specialized (for a concrete example: should I be fully implementing my own `weightedindexes` function, or should I be implementing the `weightedindex_parts` function?).

timholy:
I'm trying to avoid being the maintainer here, but teaching/answering questions is fine! I'll need a bit of time to refresh my own memory and then address your specific questions. Feel free to post more things that confuse you in the meantime.

CC @mkitti

mkitti:
I'll start taking a look as well. Thanks for the ping.

mileslucas:
Okay, thanks both for the help. Here's what I've gathered about the method and how it can be used within this framework.

Overall the Lanczos method is _about the same_ as using BSplines. There is a kernel function that can be used to create a discrete impulse using a certain number of local values. This is then discretely convolved with the original data (with padding at edges) to produce the interpolated value.

So the overall workflow should look pretty similar to BSplines- get the positions, generate the kernel impulse, and then use that to create a WeightedIndex. There are some options we can consider for the boundary behavior, but I think just using 0s padding is good for now to focus on the interface. I'm not entirely certain about whether or not there could be a prefilter step for the kernel (everytime I read the literature I can't quite grasp how it would apply to the Lanczos compared to the examples they give for linear interp.).

mkitti:
Hi @mileslucas , both B-splines and Lancsoz could be thought as approximations of sinc interpolation. Higher order B-splines approach the sinc kernel. Sinc can be problematic because it requires infinite support. Both are attempts to effectively limit the support needed. Both are great for bandlimited signals that should be smooth.

src/convolutions/convolutions.jl
```diff
@@ -0,0 +1,67 @@
...
+
+"""
+    Lanczos(a=4)
+
+Lanczos resampling via a kernel with size `a`.
+
+This form of interpolation is merely the discrete convolution of the samples with a Lanczos kernel of size `a`. The size is directly related to how "far" the interpolation will reach for information, and has `O(a^2)` impact on runtime. A default value of 4 matches the OpenCV implementation `lanczos4`.
+"""
+struct Lanczos <: InterpolationType
+    a::Integer
```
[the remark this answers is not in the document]
mkitti: There may also be an opportunity to memoize or otherwise cache the kernel here.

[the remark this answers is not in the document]
mileslucas: According to a few sources I've came across the Lanczos window is defined explicitly with a single parameter ([here is the paper I have been referencing the most, see section 11](http://www.ipol.im/pub/art/2011/g_lmii/article.pdf)). None of the BSplines separate their degree and support, either, so this seems divergent from the current style.

Not saying we couldn't, but it seems like unnecessary work (especially if there's no proper use-case for it).

src/convolutions/convolutions.jl
```diff
@@ -0,0 +1,67 @@
...
+
+Lanczos() = Lanczos(4)
+
+"""
+    lanczos(x, a)
+
+Implementation of the [Lanczos kernel](https://en.wikipedia.org/wiki/Lanczos_resampling)
+"""
+lanczos(x::T, a::Integer) where{T} = lanczos(T, x, a)
+lanczos(T, x, a::Integer) = abs(x) < a ? T(sinc(x) * sinc(x / a)) : zero(T)
```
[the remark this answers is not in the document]
mileslucas: worth taking a look as well as using some pre-computed values for certain common widths (this is what opencv does for lanczos4, for instance).

src/convolutions/convolutions.jl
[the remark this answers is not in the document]
mkitti: Mathematically lanczos definitely has a single parameter. The practical computation requires a cutoff somewhere though, which is a practical approximation to when the kernel is "almost" zero.  If you plot out the kernel you will notice that it actually is not zero beyond the default limits.

The frequency domain implications of such a cutoff are actually quite profound. The zero cutoff is equivalent to multiplying the real domain by a `rect` which in terms equates to convolution with a sinc.

The BSplines have an explicit cutoff mathematically because they actually are zero beyond a certain point. This is not necessarily the case for lanczos. `sinc(x) * sinc(x/a)` is not zero for `abs(x) >= a`.

If one parameter is given, we can practically revert to the current case. The only real cost here is storing an extra integer. Being able to change the support though will allow one to compare two kernels with distinct scaling factors but the same support.

mileslucas:
okay, just pushed some stuff I've tried wrestling with. Right now you can instantiate an interpolator, but I'm not sure about the accuracy.

Can either of you explain the `positions` function? I don't understand what is supposed to be returned, and it appears to have a profound effect on the interpolator output.

src/convolutions/convolutions.jl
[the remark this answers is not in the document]
mileslucas: Thanks for the further explanation, I'll add it in!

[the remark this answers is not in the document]
mileslucas: Actually, upon further thinking, I want to be careful about how this is declared.

Since you've written this comment, I've switched to using `Lanczos{a}()` to represent the type, with the expectation that we could use dispatch like `f(::Lanczos{4})` when applicable for specializing fucntions using look up tables, for instance.

If we separate the parameters, we either have to use both as type parameters for allowing specialization, or we commit to not specializing at all (especially because the LUT will be for the default case of same scale and support).

For now, I'll write up an impl that doesn't use any special type parameters or generated functions.

mileslucas:
I've put some LUT for lanczos4 and it seems to be significant- using `Lanczos(3)` is about 1.5 times slower than using `Lanczos(4)` for non-integral `δx`s. For integral `δx` I've decided rather than specializing the `lanczos` function, I've added some `ntuple` calls if `isinteger(δx)`, which is what we'll have to do because we can't specialize off the `Integer` type.

For some timing reference, currently the cubic bspline takes ~20ns to do a single interpolation while the lanczos interpolator takes ~500ns for integral values, ~900ns for Lanczos(4) non-integral values, and ~1.5 for Lanczos(3).

mkitti:
Thanks, Miles. By "non-integral" do you mean non-integer?

mileslucas:
Yes, so I would test say `0.0, 1.0` for "integral" values and `0.2, 0.5` for "non-integral" values. Within the code, it's anything that returns true for `isinteger`

timholy:
> Can either of you explain the positions function?

It's been a while, but it seems that it's designed to split a continuous location into two pieces: a "center index" and an offset. For instance,

```julia
julia> Interpolations.positions(Linear(), 1:3, 1.2)
(1, 0.19999999999999996)

julia> Interpolations.positions(Linear(), 1:3, 2.2)
(2, 0.20000000000000018)

julia> Interpolations.positions(Linear(), 1:3, 2.8)
(2, 0.7999999999999998)
```

For `Linear()` this mostly splits integral and fractional pieces. However, in one case it does something sneaky to avoid a `BoundsError` from interpolation at the right edge:
```julia
julia> Interpolations.positions(Linear(), 1:3, 1.0)
(1, 0.0)

julia> Interpolations.positions(Linear(), 1:3, 2.0)
(2, 0.0)

julia> Interpolations.positions(Linear(), 1:3, 3.0)
(2, 1.0)
```

I think the goal is to prevent needing special checking for the right edge in the core interpolant-evaluation. In multidimensional interpolation, this approach means you have one branch (or select) statement per dimension, whereas checking in the core interpolation routine would give `2^n` branches per evaluation.

mileslucas:
Wonderful, that seems to clear things up a little along with some of the testing I had done. My understanding of it is that it is the left-most index from which the `value_weights` will start at.

Another piece that has me curious: `value_weights` for, say, Cubic B-Spline only returns two weights. It's possible I don't understand how B-Spline interpolation works, but shouldn't Cubic splines include weights from further points?

This seems to be part of the considerable speed differences between the two, since B-Splines uniformly seem to be doing discrete convolution over 4 pixels whereas Lanczos is occurring over `4n^2` ~ 64 for  Lanczos4.

timholy:
> Another piece that has me curious: value_weights for, say, Cubic B-Spline only returns two weights. It's possible I don't understand how B-Spline interpolation works, but shouldn't Cubic splines include weights from further points?

Your understanding seems spot on, the misunderstanding must be elsewhere? I get this:
```julia
julia> Interpolations.value_weights(Cubic(Line()), 0.2)
(0.08533333333333334, 0.6306666666666666, 0.28266666666666657, 0.0013333333333333335)
```

mileslucas:
well, it seems I got that wrong (probably just mis-remembering).

So right now I'm feeling confident-ish in the accuracy of my implementation, but I'm running into a speed barrier. I've been constantly testing timings for the Lanczos4 vs. the cubic b-spline. 

During my tests, I've noticed that `value_weights` is pretty similar for both of them, but somehow calling `weightedindexes` ends up being much faster. I have to believe this ends up because of a compiler-trick. Right now my `value_weights` function returns `Tuple{Vararg{Float64,N}}` which is slightly unstable.

Any clues on how to elude that?

timholy:
I might be able to figure that out, but @mkitti want to take a shot at it as part of your educating on maintainership? This is more of a generic Julia question, I imagine (I haven't even looked at the code though) than something specific about knowing Interpolations well.

mkitti:
My guess at the moment is that for `lanczos` `value_weights` still requires calling trigonometric functions which is harder to vectorize at the compiler level than normal polynomials.

Earlier I recommended memoization so that we did not necessarily need to call trig functions repeatedly. Rather than calling lanczos directly, perhaps you could you use spline interpolation to interpolate then lanczos kernel after sampling it. In subsequent calls you would just use the interpolated lanczos rather than recalculating it. Perhaps FFT based calculation might be useful here.

On the type stability issue, it may make sense to parameterize `Lanczos` on `n` and also explicitly expand out the cases for low or common n.

mkitti:
@mileslucas Could we also add some tests here? Some ideas would be to interpolate some constant functions or delta distributions.

src/lanczos/lanczos.jl
```diff
@@ -0,0 +1,105 @@
...
+function positions(it::Lanczos, ax, x)
+    xf = floorbounds(x, ax)
+    δx = x - xf
+    fast_trunc(Int, xf) - it.n + 1, δx
+end
+
+function value_weights(it::Lanczos, δx::S) where S
+    T = float(S)
+    # short-circuit if integral
+    isinteger(δx) && return ntuple(i -> i == it.n - δx ? one(T) : zero(T), 2it.n)
```
timholy, line 68: @mkitti, not sure where you are currently on your Julia knowledge. Let's proceed by hints to give you a chance to figure out as much as you can on your own. This first hint is that the performance issue mentioned in the discussion is on this line. First task is to describe what about this line contributes to poor performance. Second task is to come up with an understanding of what might be done about it, along with the likely tradeoffs of that solution. (Do your best! I'm here to help, but you'll get there faster if you put effort into solving this yourself.)

mkitti: My understanding is there are several specializations issues here.

1. One issue is that given type `Lanczos` and `S` the size of the ntuple that is variable and cannot be predicted. My suggestion above is to parameterize `Lanczos` on n. That is given input types to value_weights of  `Lanczos{N}` and `S` the output of this function is a tuple of length 2N. The risk is over-specialization. There would be a distinct version of `value_weights` for each N. I think this is the main issue.

2. Another issue is there could be some optimization of the anonymous function and use of ntuple. `it.n -δx` is a constant and indicates the exact index where the tuple should be 1. `dx` could be converted to an integer. The last argument could be wrapped by `Val` to specialize `ntuple` and stabilize the return type.

timholy: Very good. Your first point hits all the main points, both in terms of the solution and its downsides. @mileslucas might be able to tell us if the diversity of lengths would be a major problem.

For the second one, wrapping an ordinary number in `Val` doesn't help, you need to be able to predict it from the beginning. See https://docs.julialang.org/en/latest/manual/performance-tips/#man-performance-value-type-1.

mkitti: Regarding `Val`, I think that depends on `N`:

https://github.com/JuliaLang/julia/blob/master/base/ntuple.jl#L17-L73

```
function ntuple(f::F, n::Integer) where F
    t = n == 0  ? () :
        n == 1  ? (f(1),) :
        n == 2  ? (f(1), f(2)) :
        n == 3  ? (f(1), f(2), f(3)) :
        n == 4  ? (f(1), f(2), f(3), f(4)) :
        n == 5  ? (f(1), f(2), f(3), f(4), f(5)) :
        n == 6  ? (f(1), f(2), f(3), f(4), f(5), f(6)) :
        n == 7  ? (f(1), f(2), f(3), f(4), f(5), f(6), f(7)) :
        n == 8  ? (f(1), f(2), f(3), f(4), f(5), f(6), f(7), f(8)) :
        n == 9  ? (f(1), f(2), f(3), f(4), f(5), f(6), f(7), f(8), f(9)) :
        n == 10 ? (f(1), f(2), f(3), f(4), f(5), f(6), f(7), f(8), f(9), f(10)) :
        _ntuple(f, n)
    return t
end

function _ntuple(f, n)
    @_noinline_meta
    (n >= 0) || throw(ArgumentError(string("tuple length should be ≥ 0, got ", n)))
    ([f(i) for i = 1:n]...,)
end

# inferrable ntuple (enough for bootstrapping)
ntuple(f, ::Val{0}) = ()
ntuple(f, ::Val{1}) = (@_inline_meta; (f(1),))
ntuple(f, ::Val{2}) = (@_inline_meta; (f(1), f(2)))
ntuple(f, ::Val{3}) = (@_inline_meta; (f(1), f(2), f(3)))


"""
    ntuple(f, ::Val{N})
Create a tuple of length `N`, computing each element as `f(i)`,
where `i` is the index of the element. By taking a `Val(N)`
argument, it is possible that this version of ntuple may
generate more efficient code than the version taking the
length as an integer. But `ntuple(f, N)` is preferable to
`ntuple(f, Val(N))` in cases where `N` cannot be determined
at compile time.
# Examples
```jldoctest
julia> ntuple(i -> 2*i, Val(4))
(2, 4, 6, 8)
"""
@inline function ntuple(f::F, ::Val{N}) where {F,N}
    N::Int
    (N >= 0) || throw(ArgumentError(string("tuple length should be ≥ 0, got ", N)))
    if @generated
        quote
            @nexprs $N i -> t_i = f(i)
            @ncall $N tuple t
        end
    else
        Tuple(f(i) for i = 1:N)
    end
end
```

I think @mileslucas is mainly concerned up to Lanczos{4} up to , which would be mean N = 8 for `ntuple`

mkitti: julia> using BenchmarkTools

julia> test_ntuple(N,x) = ntuple(i -> i == N - x, 2N)
test_ntuple (generic function with 2 methods)

julia> test_ntuple_pass_N(v::Val{N},x) where N = ntuple(i -> i == N/2 - x, N)
test_ntuple_pass_N (generic function with 1 method)

julia> test_ntuple_val(v::Val{N},x) where N = ntuple(i -> i == N/2 - x, v)
test_ntuple_val (generic function with 2 methods)

julia> let N = Ref(2), x = Ref(1.0), v = Ref(Val(N[]*2))
           @btime $test_ntuple($N[], $x[])
           @btime $test_ntuple_pass_N($v[], $x[])
           @btime $test_ntuple_val($v[], $x[])
       end
  19.670 ns (1 allocation: 16 bytes)
  19.102 ns (1 allocation: 16 bytes)
  5.727 ns (0 allocations: 0 bytes)
(true, false, false, false)

julia> let N = Ref(3), x = Ref(1.0), v = Ref(Val(N[]*2))
           @btime $test_ntuple($N[], $x[])
           @btime $test_ntuple_pass_N($v[], $x[])
           @btime $test_ntuple_val($v[], $x[])
       end
  21.712 ns (1 allocation: 16 bytes)
  21.182 ns (1 allocation: 16 bytes)
  8.882 ns (0 allocations: 0 bytes)
(false, true, false, false, false, false)


julia> let N = Ref(4), x = Ref(1.0), v = Ref(Val(N[]*2))
           @btime $test_ntuple($N[], $x[])
           @btime $test_ntuple_pass_N($v[], $x[])
           @btime $test_ntuple_val($v[], $x[])
       end
  23.937 ns (1 allocation: 16 bytes)
  23.129 ns (1 allocation: 16 bytes)
  11.213 ns (0 allocations: 0 bytes)
(false, false, true, false, false, false, false, false)

timholy: It's not just the construction of the tuple; it's also the *consumer* of the tuple. Consider:
```julia
julia> struct HideInt
           n::Int
       end

julia> buildt(::Val{N}) where N = ntuple(identity, Val(N))
buildt (generic function with 1 method)

julia> buildt(h::HideInt) = ntuple(identity, h.n)
buildt (generic function with 2 methods)

julia> f(x) = sum(buildt(x))
f (generic function with 1 method)

julia> using BenchmarkTools
[ Info: Precompiling BenchmarkTools [6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf]

julia> @btime f(x) setup=(x=Val(6))    # this result is "fake"
  0.016 ns (0 allocations: 0 bytes)
21

julia> @btime f(x) setup=(x=Val(rand(1:6)))
  7.640 ns (0 allocations: 0 bytes)
1

julia> @btime f(x) setup=(x=HideInt(rand(1:6)))
  14.472 ns (0 allocations: 0 bytes)
```

and

```julia
julia> @code_warntype f(Val(5))
Variables
  #self#::Core.Compiler.Const(f, false)
  x::Core.Compiler.Const(Val{5}(), false)

Body::Int64
1 ─ %1 = Main.buildt(x)::Core.Compiler.Const((1, 2, 3, 4, 5), false)
│   %2 = Main.sum(%1)::Core.Compiler.Const(15, false)
└──      return %2

julia> @code_warntype f(HideInt(5))
Variables
  #self#::Core.Compiler.Const(f, false)
  x::HideInt

Body::Any
1 ─ %1 = Main.buildt(x)::Tuple{Vararg{Int64,N} where N}
│   %2 = Main.sum(%1)::Any
└──      return %2
```

timholy: Also, use backticks to avoid pinging any user named `@Btime`

mkitti: This is getting a bit long, but parameterizing `Lanczos` alone does not appear to create a specific return type. I just get `Tuple`:

```
julia> import Interpolations: InterpolationType, _lanczos4, lanczos

julia> struct Lanczos{N} <: InterpolationType where {N <: Int}
           a::Int

           function Lanczos{N}(a) where {N}
               N < a && @warn "Using a smaller support than scale for Lanczos window. Proceed with caution."
               new(a)
           end
       end

julia> Lanczos(a=4) = Lanczos{a}(a)
Lanczos

julia> function value_weights(it::Lanczos{N}, δx::S) where {N,S}
           T = float(S)
           # short-circuit if integral
           isinteger(δx) && return ntuple(i -> i == N - δx ? one(T) : zero(T), 2N)
           # LUTs
           it.a == N == 4 && return _lanczos4(δx)

           cs = ntuple(i -> lanczos(N - i + δx, it.a, N), 2N)
           normed_cs = ntuple(i -> cs[i] / sum(cs), length(cs))
           return normed_cs
       end
value_weights (generic function with 1 method)

julia> `@code_warntype` value_weights(Lanczos(2),1.0)
Variables
  #self#::Core.Compiler.Const(value_weights, false)
  it::Lanczos{2}
  δx::Float64
  #25::var"#25#28"{2,Float64,DataType}
  #26::var"#26#29"{2,Lanczos{2},Float64}
  #27::var"#27#30"{_A} where _A
  T::Type{Float64}
  cs::Tuple{Vararg{Float64,N} where N}
  normed_cs::Tuple
  @_10::Bool

Body::Tuple
1 ─       Core.NewvarNode(:(#25))
│         Core.NewvarNode(:(#26))
│         Core.NewvarNode(:(#27))
│         Core.NewvarNode(:(cs))
│         Core.NewvarNode(:(normed_cs))
│         (T = Main.float($(Expr(:static_parameter, 2))))
│   %7  = Main.isinteger(δx)::Bool
└──       goto #3 if not %7
2 ─ %9  = Main.:(var"#25#28")::Core.Compiler.Const(var"#25#28", false)
│   %10 = $(Expr(:static_parameter, 1))::Core.Compiler.Const(2, false)
│   %11 = Core.typeof(δx)::Core.Compiler.Const(Float64, false)
│   %12 = Core.typeof(T::Core.Compiler.Const(Float64, false))::Core.Compiler.Const(DataType, false)
│   %13 = Core.apply_type(%9, %10, %11, %12)::Core.Compiler.Const(var"#25#28"{2,Float64,DataType}, false)
│         (#25 = %new(%13, δx, T::Core.Compiler.Const(Float64, false)))
│   %15 = #25::var"#25#28"{2,Float64,DataType}
│   %16 = (2 * $(Expr(:static_parameter, 1)))::Core.Compiler.Const(4, false)
│   %17 = Main.ntuple(%15, %16)::Tuple
└──       return %17
3 ─ %19 = Base.getproperty(it, :a)::Int64
│   %20 = (%19 == $(Expr(:static_parameter, 1)))::Bool
└──       goto #5 if not %20
4 ─       (@_10 = $(Expr(:static_parameter, 1)) == 4)
└──       goto #6
5 ─       (@_10 = false)
6 ┄       goto #8 if not @_10::Core.Compiler.Const(false, false)
7 ─       Core.Compiler.Const(:(Main._lanczos4(δx)), false)
└──       Core.Compiler.Const(:(return %26), false)
8 ┄ %28 = Main.:(var"#26#29")::Core.Compiler.Const(var"#26#29", false)
│   %29 = $(Expr(:static_parameter, 1))::Core.Compiler.Const(2, false)
│   %30 = Core.typeof(it)::Core.Compiler.Const(Lanczos{2}, false)
│   %31 = Core.typeof(δx)::Core.Compiler.Const(Float64, false)
│   %32 = Core.apply_type(%28, %29, %30, %31)::Core.Compiler.Const(var"#26#29"{2,Lanczos{2},Float64}, false)
│         (#26 = %new(%32, it, δx))
│   %34 = #26::var"#26#29"{2,Lanczos{2},Float64}
│   %35 = (2 * $(Expr(:static_parameter, 1)))::Core.Compiler.Const(4, false)
│         (cs = Main.ntuple(%34, %35))
│   %37 = Main.:(var"#27#30")::Core.Compiler.Const(var"#27#30", false)
│   %38 = Core.typeof(cs)::Type{Tuple{Vararg{Float64,N}}} where N
│   %39 = Core.apply_type(%37, %38)::Type{var"#27#30"{_A}} where _A
│         (#27 = %new(%39, cs))
│   %41 = #27::var"#27#30"{_A} where _A
│   %42 = Main.length(cs)::Int64
│         (normed_cs = Main.ntuple(%41, %42))
└──       return normed_cs
```

mkitti: Adding in `Val` seems more stable. I get `NTuple{4,Float64}`:

```
julia> function value_weights(it::Lanczos{N}, δx::S) where {S,N}
           # short-circuit if integral
           isinteger(δx) && return ntuple(i->convert(float(S),i == N - δx), Val(2N))

           # LUTs
           it.a == N == 4 && return _lanczos4(δx)

           cs = ntuple(i -> lanczos(N - i + δx, it.a, N), Val(2N))
           normed_cs = ntuple(i -> cs[i] / sum(cs), Val(length(cs)))
           return normed_cs
       end
value_weights (generic function with 1 method)

julia> @code_warntype value_weights(Lanczos(2),1.0)
Variables
  #self#::Core.Compiler.Const(value_weights, false)
  it::Lanczos{2}
  δx::Float64
  #43::var"#43#46"{2,Float64,Float64}
  #44::var"#44#47"{2,Lanczos{2},Float64}
  #45::var"#45#48"{NTuple{4,Float64}}
  cs::NTuple{4,Float64}
  normed_cs::NTuple{4,Float64}
  @_9::Bool

Body::NTuple{4,Float64}
1 ─       Core.NewvarNode(:(#43))
│         Core.NewvarNode(:(#44))
│         Core.NewvarNode(:(#45))
│         Core.NewvarNode(:(cs))
│         Core.NewvarNode(:(normed_cs))
│   %6  = Main.isinteger(δx)::Bool
└──       goto #3 if not %6
2 ─ %8  = Main.:(var"#43#46")::Core.Compiler.Const(var"#43#46", false)
│   %9  = $(Expr(:static_parameter, 2))::Core.Compiler.Const(2, false)
│   %10 = $(Expr(:static_parameter, 1))::Core.Compiler.Const(Float64, false)
│   %11 = Core.typeof(δx)::Core.Compiler.Const(Float64, false)
│   %12 = Core.apply_type(%8, %9, %10, %11)::Core.Compiler.Const(var"#43#46"{2,Float64,Float64}, false)
│         (#43 = %new(%12, δx))
│   %14 = #43::var"#43#46"{2,Float64,Float64}
│   %15 = (2 * $(Expr(:static_parameter, 2)))::Core.Compiler.Const(4, false)
│   %16 = Main.Val(%15)::Core.Compiler.Const(Val{4}(), true)
│   %17 = Main.ntuple(%14, %16)::NTuple{4,Float64}
└──       return %17
3 ─ %19 = Base.getproperty(it, :a)::Int64
│   %20 = (%19 == $(Expr(:static_parameter, 2)))::Bool
└──       goto #5 if not %20
4 ─       (@_9 = $(Expr(:static_parameter, 2)) == 4)
└──       goto #6
5 ─       (@_9 = false)
6 ┄       goto #8 if not @_9::Core.Compiler.Const(false, false)
7 ─       Core.Compiler.Const(:(Main._lanczos4(δx)), false)
└──       Core.Compiler.Const(:(return %26), false)
8 ┄ %28 = Main.:(var"#44#47")::Core.Compiler.Const(var"#44#47", false)
│   %29 = $(Expr(:static_parameter, 2))::Core.Compiler.Const(2, false)
│   %30 = Core.typeof(it)::Core.Compiler.Const(Lanczos{2}, false)
│   %31 = Core.typeof(δx)::Core.Compiler.Const(Float64, false)
│   %32 = Core.apply_type(%28, %29, %30, %31)::Core.Compiler.Const(var"#44#47"{2,Lanczos{2},Float64}, false)
│         (#44 = %new(%32, it, δx))
│   %34 = #44::var"#44#47"{2,Lanczos{2},Float64}
│   %35 = (2 * $(Expr(:static_parameter, 2)))::Core.Compiler.Const(4, false)
│   %36 = Main.Val(%35)::Core.Compiler.Const(Val{4}(), true)
│         (cs = Main.ntuple(%34, %36))
│   %38 = Main.:(var"#45#48")::Core.Compiler.Const(var"#45#48", false)
│   %39 = Core.typeof(cs)::Core.Compiler.Const(NTuple{4,Float64}, false)
│   %40 = Core.apply_type(%38, %39)::Core.Compiler.Const(var"#45#48"{NTuple{4,Float64}}, false)
│         (#45 = %new(%40, cs))
│   %42 = #45::var"#45#48"{NTuple{4,Float64}}
│   %43 = Main.length(cs)::Core.Compiler.Const(4, false)
│   %44 = Main.Val(%43)::Core.Compiler.Const(Val{4}(), true)
│         (normed_cs = Main.ntuple(%42, %44))
└──       return normed_cs
```

mkitti:
@mileslucas I just pushed a commit on my fork since I've been working on it anyways. https://github.com/mkitti/Interpolations.jl/commit/9fc2303bb5ea0e610a2cf98cb0ad3dfb2cfb27f0

```
julia> @code_warntype Interpolations.value_weights(Lanczos(2),1.2)
Variables
  #self#::Core.Compiler.Const(Interpolations.value_weights, false)
  it::Lanczos{2}
  δx::Float64
  #60::Interpolations.var"#60#63"{2,Float64,Float64}
  #61::Interpolations.var"#61#64"{2,Lanczos{2},Float64}
  #62::Interpolations.var"#62#65"{NTuple{4,Float64}}
  cs::NTuple{4,Float64}
  normed_cs::NTuple{4,Float64}
  @_9::Bool

Body::NTuple{4,Float64}
...
```

mileslucas:
@mkitti I'll take a look- thanks for all the help; I've been busy the last couple days. Some tests and some more docs will bring this out of the draft stage.

mkitti:
There asymmetry in the interpolation that bothers me. The kernel itself is symmetric, so I would expect the interpolation to be also asymmetric. I'll post a concrete example using a delta distribution input later.

mkitti:
I'm going to turn my focus back to this in August.

src/lanczos/lanczos.jl
timholy: Sorry for my delay here. In your first post, `where {N <: Int}` doesn't do what you are intending (really it should be `N::Int`, definitely not the same thing as `N <: Int`, but that's not supported).

mileslucas:
So I'm kind of torn because the wikipedia entry for lanczos resampling doesn't seem to consistently bound `a` is inclusive or exclusive. However, one of the properties is listed that it has `2a-1` nodes, so for 4 that would be 7 nodes, so a range of `-3:3`, instead of `-3:4` as the wikipedia formula states. On the other hand, the Lanczos window could be asymmetric given the weird sinc windowing bounds shown on that same page.

mkitti:
It should be symmetric around zero, `-a:δx:a`. In the common case, the value of the kernel at `-a` or `a` is zero anyways, so perhaps we could omit both endpoints.

mileslucas:
Yeah, that makes sense. So the lanczos{4} kernel should be `-3:3`

mkitti:
My concern at the moment is there are small differences depending on whether you interpolate to the left or right.

```julia
julia> using Interpolations

julia> A = zeros(9); A[5] = 1
1

julia> A
9-element Array{Float64,1}:
 0.0
 0.0
 0.0
 0.0
 1.0
 0.0
 0.0
 0.0
 0.0

julia> itp = interpolate(A, Lanczos(4))
9-element Interpolations.LanczosInterpolation{Float64,1,Lanczos{4},OffsetArrays.OffsetArray{Float64,1,Array{Float64,1}},Tuple{Base.OneTo{Int64}}}:
 0.0
 0.0
 0.0
 0.0
 1.0
 0.0
 0.0
 0.0
 0.0

julia> itp.(5:0.1:6) - itp.(5:-0.1:4)
11-element Array{Float64,1}:
  0.0
 -3.3306690738754696e-16
 -1.1102230246251565e-16
 -3.3306690738754696e-16
 -4.440892098500626e-16
 -4.440892098500626e-16
 -1.1102230246251565e-16
 -1.1102230246251565e-16
 -4.163336342344337e-16
 -3.0531133177191805e-16
  0.0
```

The problem is that depending on whether you go left or right you get a different set of weights and slightly different normalization:

```julia
julia> Interpolations.value_weights(Lanczos(4),0.1)
(-0.008461469489771193, 0.02830383750635515, -0.07868372457374333, 0.9823541017536349, 0.10038854542802601, -0.034576156067320264, 0.01132072612773238, -0.0006458606849136277)

julia> Interpolations.value_weights(Lanczos(4),-0.1)
(0.011320030240761888, -0.03457403066509556, 0.10038237452697592, 0.9822937162790474, -0.07867888787174168, 0.02830209766462471, -0.008460949361795055, -0.0005843508127775427)
```

The above tuples should be flipped versions of each other. However, the first element corresponds to the second to the last element in the other tuple. The first element should correspond to the last element in the other tuple.

src/lanczos/lanczos.jl
```diff
@@ -0,0 +1,108 @@
...
+function value_weights(it::Lanczos, δx::S) where S
+    N = degree(it)
+    # short-circuit if integral
+    isinteger(δx) && return ntuple(i->convert(float(S), i == N - δx), Val(2N))
+
+    # LUTs
+    it.a === N === 4 && return _lanczos4(δx)
+
+    cs = ntuple(i -> lanczos(N - i + δx, it.a, N), Val(2N))
+    normed_cs = ntuple(i -> cs[i] / sum(cs), Val(length(cs)))
```
mkitti, line 75: You only need to calculate `sum(cs)` once here rather than `length(cs)` times.

mkitti:
Maybe using rationals is the better test:

```julia
julia> A = zeros(9); A[5] = 1
1

julia> itp = interpolate(A, Lanczos(4))
9-element Interpolations.LanczosInterpolation{Float64,1,Lanczos{4},OffsetArrays.OffsetArray{Float64,1,Array{Float64,1}},Tuple{Base.OneTo{Int64}}}:
 0.0
 0.0
 0.0
 0.0
 1.0
 0.0
 0.0
 0.0
 0.0

julia> itp.(5:1//10:6) - itp.(5:-1//10:4)
11-element Array{Float64,1}:
  0.0
  0.0
 -1.1102230246251565e-16
 -4.440892098500626e-16
 -4.440892098500626e-16
 -4.440892098500626e-16
 -2.220446049250313e-16
 -1.1102230246251565e-16
 -3.885780586188048e-16
 -3.3306690738754696e-16
  0.0
```

mkitti:
Is `_lanczos4` derived directly from opencv code? https://github.com/alalek/opencv/blob/de15636724967faf62c2d1bce26f4335e4b359e5/modules/imgproc/src/resize.cpp#L917-L946

mileslucas:
Yes, except the ordering is reversed according to my testing.

mkitti:
I found a reference for Lanczos here:
https://journals.ametsoc.org/jamc/article/18/8/1016/350573/Lanczos-Filtering-in-One-and-Two-Dimensions

I'm trying to justify the 8-tap filter but cannot find it in that reference. Also, perhaps the OpenCV compatibility should be a special case rather than the default for lanczos4.

mkitti:
Noting some issues to avoid in downstream applications:
https://twitter.com/ChangLeeTW/status/1258411911764475905

mkitti:
Here is the current state of affairs after my edits. I think the best is to actually offer two distinct implementations of `lanczos4` and allow the user to choose.

`Lanczos(4)` now offers the direct implementation using `sinc(x)*sinc(x/4)`

`Lanczos4OpenCV()` offers the accelerated version using the lookup table that uses the the sin addition identity at the cost of some floating point numerical accuracy.

They are both subtypes of `AbstractLanczos` which implements the common details. The deviate mainly in how `value_weights` is implemented.

```julia
julia> using Interpolations
[ Info: Precompiling Interpolations [a98d9a8b-a2ab-59e6-89dd-64a1c18fca59]

julia> A = zeros(9); A[5] = 1
1

julia> itp = interpolate(A, Lanczos(4));

julia> itp.(5:0.1:6) - itp.(5:-0.1:4)
11-element Array{Float64,1}:
  0.0
  0.0
  0.0
 -2.220446049250313e-16
  0.0
  0.0
  0.0
  1.1102230246251565e-16
  0.0
  0.0
  0.0

julia> itp.(5:1//10:6) - itp.(5:-1//10:4)
11-element Array{Float64,1}:
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0

julia> itp_ocv = interpolate(A, Lanczos4OpenCV());

julia> itp_ocv.(5:1//10:6) - itp_ocv.(5:-1//10:4)
11-element Array{Float64,1}:
  0.0
  0.0
 -1.1102230246251565e-16
 -4.440892098500626e-16
 -4.440892098500626e-16
 -4.440892098500626e-16
 -2.220446049250313e-16
 -1.1102230246251565e-16
 -3.885780586188048e-16
 -3.3306690738754696e-16
  0.0

julia> @btime itp.(5:0.1:6) - itp.(5:-0.1:4)
  8.597 μs (7 allocations: 848 bytes)
11-element Array{Float64,1}:
  0.0
  0.0
  0.0
 -2.220446049250313e-16
  0.0
  0.0
  0.0
  1.1102230246251565e-16
  0.0
  0.0
  0.0

julia> @btime itp_ocv.(5:0.1:6) - itp_ocv.(5:-0.1:4)
  3.292 μs (29 allocations: 2.52 KiB)
11-element Array{Float64,1}:
  0.0
 -3.3306690738754696e-16
 -1.1102230246251565e-16
 -3.3306690738754696e-16
 -4.440892098500626e-16
 -4.440892098500626e-16
 -1.1102230246251565e-16
 -1.1102230246251565e-16
 -4.163336342344337e-16
 -3.0531133177191805e-16
  0.0
```

mileslucas:
That seems like a smart path forward! I appreciate the digging you did into the kernels and the comments you've added, they help me understand how the OpenCV implementation works more than OpenCV did.

We're still an order of magnitude slower than B-splines, which makes this (in my opinion) unusable for image processing. Here's my quick benchmarks of a cubic B-Spline vs. Lanczos(2)
```julia
julia> @btime itp1(2.5) # cubic b-spline
  31.907 ns (1 allocation: 16 bytes)
2.5424682452694514

julia> @btime itp2(2.5) # lanczos(2)
  165.646 ns (1 allocation: 16 bytes)
2.5
```

mkitti:
I would like to have a few basic tests for this and then merge this PR.

The `AbstractLanczos` type provides some opportunities to try different performant implementations. The most obvious ones to me are to calculate the kernel by interpolating precomputed samples. That will be a classic memory cache for performance trade.

mileslucas:
I've added some basic tests. I think this symmetry issue is having a bigger problem: in the tests I've marked some broken tests like `interpolate(1:100, Lanczos(4))(99.5) # evaluates 100.5`.

mkitti:
> I've added some basic tests. I think this symmetry issue is having a bigger problem: in the tests I've marked some broken tests like `interpolate(1:100, Lanczos(4))(99.5) # evaluates 100.5`.

There is no general guarantee that interpolated values will between the values of their grid points. Since we don't have boundary conditions, the default is zero padding. Thus at the end of this range there is a huge jump from 100 back to 0. The Lanczos kernel is responding to this sharp edge with ringing. This is a known issue with Lanczos.

```julia
julia> itp = interpolate(1:100, Lanczos(4));

julia> itp(99.5)
111.47211604579358

julia> itp(99.9)
105.82152068677874
```

If you remove the sharp drop and ramp back down, you can get something more reasonable:

```julia
julia> itp = interpolate([1:100; 99:-1:1], Lanczos(4));

julia> itp.(99:0.1:101)
21-element Array{Float64,1}:
  99.0
  99.13781999870045
  99.2767429906275
  99.41385748861181
  99.54579762869791
  99.66874727258286
  99.778501188242
  99.87062159615076
  99.9407171907457
  99.9848454890934
 100.0
  99.9848454890934
  99.9407171907457
  99.87062159615077
  99.77850118824198
  99.66874727258286
  99.54579762869794
  99.41385748861181
  99.27674299062751
  99.13781999870045
  99.0
```

mkitti:
@timholy @mileslucas I intend to merge sometime on Monday (next 24 hours)

I expect to revisit Lanczos to work on boundary conditions and optimizations later on.

I also set Travis to allow failures on Julia nightly.

Issue #359, opened 2020-02-25
Type promotion


haampie:
It seems interpolating static arrays of type Float32 causes a type promotion to Float64:

```
julia> interpolate([@SVector(rand(Float32, 3)) for i=1:5], BSpline(Linear()))
5-element interpolate(::Array{SArray{Tuple{3},Float32,1,3},1}, BSpline(Linear())) with element type SArray{Tuple{3},Float64,1,3}:
 Float32[0.40708745, 0.08114183, 0.64380157] 
 Float32[0.889678, 0.40728843, 0.3275386]    
 Float32[0.064697385, 0.27480924, 0.08783579]
 Float32[0.21843159, 0.546651, 0.7467506]    
 Float32[0.6662456, 0.8931141, 0.17646503]
```

Is that expected?

haampie:
Similar issue with `scale` and without static arrays:

```
julia> itp = scale(interpolate([1f0, 2f0, 3f0], BSpline(Quadratic(Natural(OnCell())))), range(0.0f0, 10.0f0, length=3));

julia> itp(5.0f0)
1.9999999701976776
```

haampie:
There's multiple issues. Regarding `scale`:

```diff
diff --git a/src/scaling/scaling.jl b/src/scaling/scaling.jl
index b700824..5457814 100644
--- a/src/scaling/scaling.jl
+++ b/src/scaling/scaling.jl
@@ -104,7 +104,7 @@ coordlookup(r::AbstractUnitRange, x) = x - first(r) + oneunit(eltype(r))
 coordlookup(r::StepRange, x) = (x - r.start) / r.step + oneunit(eltype(r))
 
 coordlookup(r::AbstractRange, x) = (x - first(r)) / step(r) + oneunit(eltype(r))
-boundstep(r::AbstractRange) = 0.5*step(r)
+boundstep(r::AbstractRange) = step(r) / 2
 rescale_gradient(r::AbstractRange, g) = g / step(r)
 
 basetype(::Type{ScaledInterpolation{T,N,ITPT,IT,RT}}) where {T,N,ITPT,IT,RT} = ITPT
```

and 

```diff
diff --git a/src/b-splines/b-splines.jl b/src/b-splines/b-splines.jl
index e066cce..4b88b54 100644
--- a/src/b-splines/b-splines.jl
+++ b/src/b-splines/b-splines.jl
[...]

PR #373, opened 2020-06-28
Improve type promotion for SVectors in tweight()

maxfreu:
This fixes the first post of #359.


mkitti:
Thanks. I think we may still want to constrain `T`. If it is an `AbstractFloat` then we can use the type. Otherwise, we still need need to use `float` to get the corresponding float type for `T`. @timholy, do you concur?

maxfreu:
You mean like this?
```julia
tweight(A::AbstractArray{<:AbstractVector{T}}) where {T<:AbstractFloat} = T
```
To be honest, I'm not very familiar with the matter, so I rely on your judgement.
However interpolating between Ints seems to work and produce meaningful results, it just doesn't preserve the type.
```julia
itp = interpolate([@SVector(Int16[0,0]), @SVector(Int16[1,1])], BSpline(Linear())) # ... with eltype SArray{..., Int16,...}
typeof(itp(1f0))  # Float32
typeof(itp(Int16(1)))  # Int64
```

mkitti:
I think we might also want this:
```julia
tweight(A::AbstractArray{<:AbstractVector{T}}) where {T<:Integer} = typeof(float(zero(T)))
```

I'm not clear that this accomplishes what you want, however.

maxfreu:
Maybe I have to clarify what I want giving an example:
I want the element type of
```julia
itp = interpolate([@SVector(Float32[0,0]), @SVector(Float32[1,1])], BSpline(Linear()))
```
to be an `SArray{Tuple{2}, Float32, 1, 2}` instead of `SArray{Tuple{2}, Float64, 1, 2}` because this fixes the first post of #359 and (my) [downstream issue](https://github.com/Evizero/Augmentor.jl/issues/62) with Float32 interpolations in Augmentor; this would reduce memory allocation and improve speed. Having Floats with the correct bit depth would do the job; interpolating ints is a different story.

Note that unlike I first thought this is not related to calling `itp(...)`, which doesn't promote the type of its input. But code which uses the element type of the interpolation might return unexpected results.

mkitti:
That is clearer. I will examine the issue further.

mkitti:
This my is my current focus for this package.

maxfreu:
Any progress on this? Making it work for 32bit floats would already be good.

mkitti:
I am going to merge this for now, but may narrow this later if needed.

PR #377, opened 2020-08-14
ReadMe sugestion

briochemc:
Remove old (obsolete?) badges, and show the maintenance status in a quote box.

PR #384, opened 2020-10-13
patch_UniftulVector

briochemc:
Currently (v0.13.0), this throws:

```julia
using Unitful, Interpolations
x = sort(rand(10))*u"m"
y = rand(10)u"s"
itp = LinearInterpolation(x, y)
itp(0.5u"m") # <- this works
itp([0.5, 0.55]*u"m") # <- this errors
```

because (I think) of the return type being assigned via
https://github.com/JuliaMath/Interpolations.jl/blob/master/src/extrapolation/extrapolation.jl#L56
and ends up being "meters seconds" in the MWE above, which is incorrect.

This PR fixes that by instead assigning the same type as `x` for the return type. Tests passed locally so I assume this is OK :) 
(Although maybe there was a reason for this line before?)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Interpolations"
 uuid = "a98d9a8b-a2ab-59e6-89dd-64a1c18fca59"
-version = "0.13.1"
+version = "0.13.0"
 
 [deps]
 AxisAlgorithms = "13072b0f-2c55-5437-9ae7-d433b7a33950"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Interpolations"
 uuid = "a98d9a8b-a2ab-59e6-89dd-64a1c18fca59"
-version = "0.13.0"
+version = "0.13.1"
 
 [deps]
 AxisAlgorithms = "13072b0f-2c55-5437-9ae7-d433b7a33950"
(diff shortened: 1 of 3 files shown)


briochemc:
Side note: I think you forgot to register v0.13.0! (Unless it's on purpose)

mkitti:
Hi! I'm the new maintainer now since @timholy is getting busy with other packages.

The PR looks good and thanks for the test. I would like to understand why `Tret = typeof(lispyprod(zero(T), x...))` was needed in the first place before merging though. Do you have any insight in that?

I am still in the process of testing v0.13.0, but it might be time to make the new release anyways since it's been a while. We should keep this as v0.13.0. Could you change the version back?

briochemc:
Hi! Yes, I knew you were the new maintainer since https://github.com/JuliaMath/Interpolations.jl/pull/377 :) 

I have no idea what this line was for, this is why I asked
> (Although maybe there was a reason for this line before?)

Maybe @timholy can chime in? For convenience, it's this line:
https://github.com/JuliaMath/Interpolations.jl/blob/4f16e8dcde3437aff407dbbda3cc9bc6b4cd7f95/src/extrapolation/extrapolation.jl#L56

And yes, just commited with the version number down to v0.13.0.

timholy:
The idea is that the computation being performed is `coef*fx*fy` where `fx` and `fy` are fractional steps. Suppose the coefficients were `Float32` but you supplied the evaluation position with `Float64`, and since `1.0f0 * 1.0 * 1.0 === 1.0` I think you should return a `Float64` rather than a `Float32`.

That said, that implementation clearly blew it. Here you might consider returning the same type as used in the scalar case.

mkitti:
Thanks @timholy . I'm going to merge this for now and look into the implementation details later. I'll also setup for registering 0.13.0 this weekend.

PR #394, opened 2020-12-24
Fix parametric spline image URL

squaregoldfish:
I'm not 100% sure that this will work, because the image file is outside the `docs` directory. It works in the editing preview but I don't know what will happen when the final renderer runs. If it doesn't work I can copy the image across.

### DIFF
--- a/docs/src/control.md
+++ b/docs/src/control.md
@@ -149,4 +149,4 @@ using Plots
 scatter(x, y, label="knots")
 plot!(xs, ys, label="spline")
 ```
-![parametric spline](doc/images/parametric_spline.png)
+![parametric spline](/doc/images/parametric_spline.png)


mkitti:
Thanks for bringing up the issue and suggesting a fix. I am quite sure this fix will _not_ work.

The relevant branch is the `gh-pages` branch:
https://github.com/JuliaMath/Interpolations.jl/tree/gh-pages/dev
It appears that the image is simply not there.

Documenter.jl is used to create the docs which are then pushed to the `gh-pages` branch. What is needed to be done is to copy the image into assets so the HTML writer can find it?
https://juliadocs.github.io/Documenter.jl/stable/lib/internals/writers/#Documenter.Writers.HTMLWriter.HTML

The docs are generated with this script:
https://github.com/JuliaMath/Interpolations.jl/blob/master/docs/make.jl

Would you like to take a shot at trying to fix this?

squaregoldfish:
I'll give it a go - see if I can figure out how the scripts work.

PR #399, opened 2021-01-21
Add `knotsbetween` for iterating over interpolation knots

awadell1:
This pull request adds a `knotsbetween(itp; start, stop)` method for iterating over knots between `start` and `stop`

The existing `KnotIterator` was reworked to be indexable instead of using a tuple of `(index, offset)` when iterating
- This gives a large speed bump (Like ~4x based on one-off tests)
- Eliminated type instability issues caused by using both `Int` and `(Int, Any)` for the iteration state

## Fixes
Bug Fix: Prior `KnotIterator` would yield incorrect results for `Reflect` boundary condition
- Fixed by e7d9d58
- Error occurs after iterating over 2+ cycles -> Expanded test to check the first 100 knots (~16 cycles)
- Added @test_knots macro to simplify testing (And drive up the number of tests)

Fix: Corrected `IteratorEltype` to return `UnknownEltype` if Type parameter not include  (ie. `KnotIterator` vs `KnotIterator{T}`
- Fixed by 39a7923
- Impact is minimal as `eltype` would return `Any` and the `Any` is used if the eltype is unknown

## Other New Things
`KnotIterator` is now indexable, I've added an example to the Developer's documentation, but kept it out of the rest of the docs as support is limited

- Indexing doesn't work for Multidimensional Interpolations (As `KnotIterators` get wrapped in a `ProductIterator`)
- Indexing is only defined for `Int`, so slice indexing doesn't work
- It does get exported because `Base.getindex` is imported by `filter1d.jl`. _Should this be removed now that it's in `Interpolatons.jl`?_

### DIFF
--- a/docs/src/iterate.md
+++ b/docs/src/iterate.md
@@ -254,3 +254,80 @@ julia> size(kiter)
 (4,)
 
 ```
+
+## `knotsbetween`
+
+Given an `AbstractInterpolation` `itp` or results of `knots(itp)` get an iterator
+over knots between `start` and `stop`.
+
+```julia
+using Interpolations
+itp = interpolate(rand(4), options...)
+krange = knotsbetween(itp; start=1.2, stop=3.0)
+collect(kiter) # Array of knots between 1.2 and 3.0
+
+```
+
+We can iterate over all knots greater than `start` by omitting `stop`
+
+```jldoctest knotsbetween-usage; setup = :(using Interpolations)
+julia> x = [1.0, 1.5, 1.75, 2.0];
+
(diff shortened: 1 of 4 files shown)


awadell1:
@mkitti Looks like all checks are passing (It's crazy how much has been added since v1.0)
Any thoughts on the PR?

Best,
Alex

mkitti:
Looks good. I see there are some modifications to your previous pull request. Since we have not released yet, this is fine. Once we do release though, future modifications like that would require more scrutiny.

PR #400, opened 2021-01-31
Narrow compatibility requirements for StaticArrays

rdeits:
Currently, `using Interpolations` fails with StaticArrays 0.10 and 0.11
due to the missing `StaticArrays.SA` value. This PR removes the claimed
compatibility to avoid users accidentally getting a broken installation.

We'll also need to edit the General registry to apply this compatibility
fix to Interpolations.jl v0.13.0 and v0.13.1 otherwise the package
resolver will incorrectly want to downgrade to those versions if a user
has an older StaticArrays version.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,7 +17,7 @@ WoodburyMatrices = "efce3f68-66dc-5838-9240-27a6d6f5f9b6"
 AxisAlgorithms = "0.3, 1"
 OffsetArrays = "0.10, 0.11, 1.0.1"
 Ratios = "0.3, 0.4"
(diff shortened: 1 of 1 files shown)


mkitti:
`SA` was introduced in Interpolations.jl on a single line:
https://github.com/JuliaMath/Interpolations.jl/blob/4913d3bd0e95fc0845ffc5cc461e98aee2a96efb/src/lanczos/lanczos.jl#L115

Perhaps we should fix that to restore compatibility?

mkitti:
I am going to merge this for the moment and work on the general registry, although I am still considering using StaticArrays older API.

Issue #405, opened 2021-02-27
Document `interpolate!`


nandoconde:
Hello!

I tried the so-called "in-place" interpolation, `interpolate!`, which supposedly used less memory, and got an error when I tried later to scale back the results with `scale` I got an error. From what I assume, the "in-place" version decreases the size of the used nodes by two in each dimension e.g.:

When trying to interpolate using a matrix `A` as values, with size 5×11, I got error later when I tried to scale back with ranges of length 5 and 11 respectively, because it was expecting ranges of 3 and 9. If this is expected behavior, I guess it should be in the `docs`.

Thanks!

mkitti:
This makes sense. I wonder if there is a bug there somewhere. Pull requests always welcome, especially ones that just improve the documentation.

timholy:
It probably depends on the order of the interpolation. For quadratic & cubic, you have to solve a tridiagonal system. When it's not in place, the padding framework adds an "edge" around it so that you can both match the values and have the chosen boundary conditions; when you're in place you can't add an edge, so you have to sacrifice something.

nandoconde:
Great to know! 

This would be my first time contributing to something in Julia, so I am kind of lost. I guess that the steps are:

1. Branch `master` into my computer
2. Add a docstring to `interpolate!`
3. Push upstream 

And I wouldn’t know how to proceed 😅.

I could try it nonetheless if you kindly tell me how!

mkitti:
The easiest way in Julia is `]dev Interpolations` in the REPL or "using Pkg; Pkg.dev("Interpolations")". This will clone master into your `~/.julia/dev` directory. You can then edit the code there.

When you are ready to submit, you could fork this project, and then push your changes into your fork. You can then create a pull request from your fork. I would be more than happy to help walk you through that in more detail if needed.

timholy:
You can also click on the pencil icon for any of the files in the package.

nandoconde:
I have created the PR! #410

mkitti:
Thanks. Looks there's a start there. I'll comment more on the PR.

mkitti:
I merged #410. We may need to enhance that note in the future, but some docstring is better than none.

PR #410, opened 2021-03-03
Update `interpolate!` docstring

nandoconde:
As per #405


mkitti:
Let's make the docstring look a bit more like the one for `interpolate` above. Use a triple quoted docstring, and give a synopsis of the method signature. It should be pretty similar to what we have for `interpolate`.

Perhaps we can give some more details about what happens to `A` and why this function is useful. Basically, we can say that `A` is being used as the `prefilter` buffer and this function is used to reduce allocations if the content of `A` is no longer needed.

Let me know if this is too much. Thanks.

PR #411, opened 2021-03-19
Convenience constructors for constant (nearest-neighbor) interpolation

eirikbrandsaas:
Unless there is some specific reason for this not be in the code, I don't see why this shouldn't be included.

If you want to merge the pull request, I'd be happy to update the docs and tests.

### DIFF
--- a/docs/src/convenience-construction.md
+++ b/docs/src/convenience-construction.md
@@ -1,8 +1,9 @@
 
 ## Convenience notation
 
-For linear and cubic spline interpolations, `LinearInterpolation` and `CubicSplineInterpolation`
-can be used to create interpolating and extrapolating objects handily:
+For constant, linear, and cubic spline interpolations, `ConstantInterpolation`, `LinearInterpolation`, and `CubicSplineInterpolation`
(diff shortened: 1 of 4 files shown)


mkitti:
Seems reasonable. Add some tests please.

Also, perhaps we would like to support `Previous`, `Nearest`, and `Next` in these convenience constructors.
https://github.com/JuliaMath/Interpolations.jl/blob/master/src/b-splines/constant.jl#L4-L6

eirikbrandsaas:
I added tests for regular and irregular grids for 1d and 2d constant interpolation.

I also updated the docs (in two separate commits).

The first 08b0ee5 just adds constant interpolations to the docs and lists it as working with irregular grids.
The second f454c7d adds a motivating example on the use of the convenience constructor (because it wasn't that obvious to me at first).

I didn't add the previous or next, because a) I don't need them (sorry!) and b) maybe they should be added in a more clever way (e.g. with an optional argument for using the different constant methods).

mkitti:
Wonderful, thank you.

PR #417, opened 2021-04-18
Make documentation look prettier

rick2047:
I just added some syntax highlighting to some files. I recently learned how to do syntax highlighting, so ... you know, highlight everything!!!!

In some places we use `jldoctest derivs` which dont know how to handle. Should I just add highlighting like normal?


mkitti:
Thanks

PR #418, opened 2021-04-28
Docs for monotonic interpolation

mateuszbaran:
I've written some documentation for monotonic interpolation following this thread on discourse: https://discourse.julialang.org/t/interpolations-jl-discrete-cdf-to-pdf/60124 .


mkitti:
Looks good to me. Thank you.

Issue #401, opened 2021-02-08
Add non-parenthetical BSpline constructor?


stillyslalom:
This is mostly a matter of taste, but the current `BSpline` constructor has too damn many parentheses. 

`interpolate(A, BSpline(Cubic(Line(OnGrid()))))` - heaven help you if you don't have bracket-matching turned on!

`∘(args...)` allows a more-legible syntax:

```
julia> Interpolations.BSpline(args...) = BSpline(∘(args...)())

julia> BSpline(Cubic, Line, OnGrid)
BSpline(Cubic(Line(OnGrid())))
```

If it's a good idea, I can put together a PR.

mkitti:
I agree. The parentheses can be confusing. Do note for this specific case, there is `CubicSplineInterpolation`:

```julia
julia> using Interpolations

help?> CubicSplineInterpolation
search: CubicSplineInterpolation

  etp = CubicSplineInterpolation(knots, A; bc=Line(OnGrid()), extrapolation_bc=Throw())

  A shorthand for extrapolate(interpolate(knots, A, BSpline(Cubic(bc))), extrapolation_bc).
```

One could also apply the alternative pipe syntax:
```julia
julia> OnGrid() |> Line |> Cubic |> BSpline
BSpline(Cubic(Line(OnGrid())))
```

I'm also trying to find a good reason that `OnGrid` needs to be constructed rather than being a singleton constant. The above line shouldn't need any parentheses:

```julia
julia> OnGridSingleton = OnGrid()
OnGrid()

julia> OnGridSingleton |> Line |> Cubic |> BSpline
BSpline(Cubic(Line(OnGrid())))
```

Perhaps this is better since it does not introduce new names with the current API:
```julia
julia> using Interpolations

julia> Interpolations.Line(::Type{T}) where T <: Interpolations.GridType = Line(T())

julia> OnGrid |> Line |> Cubic |> BSpline
BSpline(Cubic(Line(OnGrid())))
```

Rather than `Interpolations.BSpline(args...) = BSpline(∘(args...)())` I would like a method with a more specific type signature. For your example,

```julia
julia> import Interpolations: BSpline, DegreeBC, BoundaryCondition, GridType

julia> BSpline(D::Type{ <:DegreeBC }, BC::Type{ <: BoundaryCondition }, GT::Type{ <: GridType }) = BSpline( D( BC( GT() ) ) )
BSpline

julia> BSpline(Cubic, Line, OnGrid)
[...]

PR #428, opened 2021-05-27
Periodic boundary conditions for Linear and Constant

### LINKED ISSUE #326: Issues with periodic boundaries
It seems like there is an issue somewhere with the periodic boundary conditions. If I approximate e.g. the sin-function in 1D, the result has errors near the boundary. 

The following is a minimal (non) working example:
```
using Interpolations
xs = range(0,stop=2π,length=100)
sinx = sin.(xs)
itp = interpolate(sinx,BSpline(Cubic(Periodic(OnGrid()))))
itp2 = scale(itp,xs)
itp3 = extrapolate(itp2,Periodic())
plot(range(2π-0.1,stop=2π+.1,length=500),x->itp3(x),label="Interpolated")
plot!(range(2π-0.1,stop=2π+.1,length=500),x->sin(x),label="Exact")
```

jaemolihm:
As discussed in #326, this PR enables periodic linear and constant interpolation without adding additional points.

The major change is that `Linear` and `Constant` are now subtypes of DegreeBC. The boundary conditions can only be `Throw(OnGrid())` or `Periodic(OnCell())` because others does not make sense or are redundant.

Existing constructors keeps working. Periodic constant and linear interpolations can be constructed similarly to cubic ones, e.g. `Linear(Periodic(OnCell()))`.

# Examples

### Example 1 (taken from https://github.com/JuliaMath/Interpolations.jl/issues/326#issuecomment-505220728)
```Julia
using Interpolations
using PyPlot
fig, axes = subplots(1, 2, figsize=(8, 4))
f(x) = sin(x)
xs = range(0, stop=2pi, length=10)[1:end-1] # note that we don't want the overlapping point at 2pi
A = f.(xs);
for i in 1:2
    sca(axes[i])
    if i == 1
        title("old")
        itp = interpolate(A, BSpline(Linear()), OnGrid()) # Old
    else
        title("new")
        itp = interpolate(A, BSpline(Linear(Periodic(OnCell())))) # New
    end
    stp = scale(itp, xs) # re-scale to the actual domain
    etp = extrapolate(stp, Periodic()) # extrapolate periodically

    xs_long = range(-2pi, stop=4pi, length=100)
    A_long = [etp[x] for x in xs_long]
    plot(xs, A, "o", label="Data")
    plot(xs_long, A_long, label="Extrapolated")
    plot(xs_long, sin.(xs_long), label="sin(x)")
    legend(loc="lower left")
end
display(gcf()); clf()
close(fig)
```

image

### Example 2

```Julia
using Interpolations
using PyPlot
f(x) = sin(2π * x + 0.5)
df(x) = 2π * cos(2π * x + 0.5)
N = 5
x = range(0, 1, length=N+1)[1:end-1]
y = f.(x)
degs = reshape([
    Constant(),
    Constant{Previous}(),
    Constant{Next}(),
    Linear(),
    Constant(Periodic(OnCell())),
    Constant{Previous}(Periodic(OnCell())),
    Constant{Next}(Periodic(OnCell())),
    Linear(Periodic(OnCell()))
    ], 4, 2)
fig, axes_ = subplots(size(degs)..., figsize=(12, 8), sharex=true, sharey=true)
for (i, deg) in enumerate(degs)
    sca(axes_[i])
    itp = interpolate(y, BSpline(deg));
    zero_to_one = range(0, 1, length=N+1)[1:end-1]
    stp = scale(itp, zero_to_one);
    etp = extrapolate(stp, Periodic());
    xx = range(-0.5, 2.5, step=0.001);
    plot(xx, etp.(xx), "C0-", label="Interpolated")
    plot(x, y, "ko", label="Data")
    plot(xx, f.(xx), "k--", label="Exact")
    # Plot derivatives
    if i in [4, 8]
        plot(xx, Interpolations.gradient.(Ref(etp), xx) ./ 5, "C1-")
        plot(xx, df.(xx) ./ 5, "y--")
    end
    title("$deg")
    if i == 1
        legend(loc="lower right")
    end
end
display(gcf()); clf()
close(fig)
```

Plots on the right side are the new features. The orange (yellow) lines are the interpolated (exact) derivatives.

image

# TODO

- [x] Documentation
- [x] Add tests
- [x] Add convenience constructor `Linear(Periodic()) = Linear(Periodic(OnCell()))`?

closes #326

### DIFF
--- a/src/b-splines/constant.jl
+++ b/src/b-splines/constant.jl
@@ -5,8 +5,24 @@ struct Nearest <: ConstantInterpType end
 struct Previous <: ConstantInterpType end
 struct Next <: ConstantInterpType end
 
-struct Constant{T<:ConstantInterpType} <: Degree{0} end
+# struct Constant{T<:ConstantInterpType} <: Degree{0} end
+# Constant() = Constant{Nearest}()
+
+struct Constant{T<:ConstantInterpType,BC<:Union{Throw{OnGrid},Periodic{OnCell}}} <: DegreeBC{0}
+    bc::BC
+end
+
+# Default to Nearest and Throw{OnGrid}
 Constant() = Constant{Nearest}()
+Constant(bc::BoundaryCondition) = Constant{Nearest}(bc)
+Constant{T}() where {T<:ConstantInterpType} = Constant{T,Throw{OnGrid}}(Throw(OnGrid()))
+Constant{T}(bc::BC) where {T<:ConstantInterpType,BC<:BoundaryCondition} = Constant{T,BC}(bc)
+
+function Base.show(io::IO, deg::Constant)
+    print(io, nameof(typeof(deg)), '{', typeof(deg).parameters[1], '}', '(')
+    show(io, deg.bc)
+    print(io, ')')
+end
 
 """
 Constant b-splines are *nearest-neighbor* interpolations, and effectively
@@ -28,6 +44,19 @@ function positions(c::Constant{Nearest}, ax, x)  # discontinuity occurs at half-
     xm = roundbounds(x, ax)
     δx = x - xm
     fast_trunc(Int, xm), δx
+end
+
+function positions(c::Constant{Previous,Periodic{OnCell}}, ax, x)
+    # We do not use floorbounds because we do not want to add a half at
+    # the lowerbound to round up.
+    xm = floor(x)
+    δx = x - xm
+    modrange(fast_trunc(Int, xm), ax), δx
+end
+function positions(c::Constant{Next,Periodic{OnCell}}, ax, x)  # discontinuity occurs at integer locations
+    xm = ceilbounds(x, ax)
+    δx = x - xm
+    modrange(fast_trunc(Int, xm), ax), δx
 end
 
 value_weights(::Constant, δx) = (1,)
--- a/src/b-splines/constant.jl
+++ b/src/b-splines/constant.jl
@@ -5,9 +5,6 @@ struct Nearest <: ConstantInterpType end
 struct Previous <: ConstantInterpType end
 struct Next <: ConstantInterpType end
 
-# struct Constant{T<:ConstantInterpType} <: Degree{0} end
-# Constant() = Constant{Nearest}()
-
 struct Constant{T<:ConstantInterpType,BC<:Union{Throw{OnGrid},Periodic{OnCell}}} <: DegreeBC{0}
     bc::BC
 end
--- a/src/b-splines/constant.jl
+++ b/src/b-splines/constant.jl
@@ -10,10 +10,10 @@ struct Constant{T<:ConstantInterpType,BC<:Union{Throw{OnGrid},Periodic{OnCell}}}
 end
 
 # Default to Nearest and Throw{OnGrid}
-Constant() = Constant{Nearest}()
-Constant(bc::BoundaryCondition) = Constant{Nearest}(bc)
+Constant(args...) = Constant{Nearest}(args...)
 Constant{T}() where {T<:ConstantInterpType} = Constant{T,Throw{OnGrid}}(Throw(OnGrid()))
 Constant{T}(bc::BC) where {T<:ConstantInterpType,BC<:BoundaryCondition} = Constant{T,BC}(bc)
+Constant{T}(::Periodic{Nothing}) where {T<:ConstantInterpType} = Constant{T,Periodic{OnCell}}(Periodic(OnCell()))
 
 function Base.show(io::IO, deg::Constant)
     print(io, nameof(typeof(deg)), '{', typeof(deg).parameters[1], '}', '(')
--- a/src/b-splines/constant.jl
+++ b/src/b-splines/constant.jl
@@ -21,6 +21,10 @@ function Base.show(io::IO, deg::Constant)
     print(io, ')')
 end
 
+function Base.show(io::IO, deg::Constant{T,Throw{OnGrid}}) where {T <: ConstantInterpType}
+    print(io, nameof(typeof(deg)), '{', typeof(deg).parameters[1], '}', '(', ')')
+end
+
 """
 Constant b-splines are *nearest-neighbor* interpolations, and effectively
 return `A[round(Int,x)]` when interpolating.
--- a/src/b-splines/linear.jl
+++ b/src/b-splines/linear.jl
@@ -4,6 +4,7 @@ end
 
 (deg::Linear)(gt::GridType) = Linear(deg.bc(gt))
 Linear() = Linear(Throw(OnGrid()))
+Linear(::Periodic{Nothing}) = Linear(Periodic(OnCell()))
 
 """
     Linear()
--- a/src/b-splines/linear.jl
+++ b/src/b-splines/linear.jl
@@ -1,4 +1,9 @@
-struct Linear <: Degree{1} end  # boundary conditions not supported
+struct Linear{BC<:Union{Throw{OnGrid},Periodic{OnCell}}} <: DegreeBC{1}
+    bc::BC
+end
+
+(deg::Linear)(gt::GridType) = Linear(deg.bc(gt))
+Linear() = Linear(Throw(OnGrid()))
 
 """
     Linear()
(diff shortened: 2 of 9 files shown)


mkitti:
@jaemolihm Sorry I'm a little confused. Is this ready for review or work-in-progress?

jaemolihm:
Sorry for the confusion. The code is ready for review. The WIP is for the documentation and tests about the new feature. Also sorry for the delay on adding the docs and tests. I'll try to do it soon.

src/b-splines/constant.jl
```diff
@@ -5,8 +5,24 @@ struct Nearest <: ConstantInterpType end
 struct Previous <: ConstantInterpType end
 struct Next <: ConstantInterpType end
 
-struct Constant{T<:ConstantInterpType} <: Degree{0} end
+# struct Constant{T<:ConstantInterpType} <: Degree{0} end
+# Constant() = Constant{Nearest}()
```
mkitti, line 9: You can remove this comment.

mkitti (changes requested):
Other than cleaning up some comments, the changes look fine to me. I will see if I can help you out with some tests and documentations.

jaemolihm:
I added tests on periodic interpolation (the `lbounds` and `ubounds` are extended by 0.5) as well as periodic extrapolation. I also added convenience constructors, e.g. `Linear(Periodic()) = Linear(Periodic(OnCell()))`

src/b-splines/constant.jl
jaemolihm: Done

jaemolihm:
I added a subsection explaining periodic extrapolation in the docs.

This PR is no more WIP and is fully ready for review. Sorry for the long delay.

jaemolihm:
Thanks or the comment. I changed the printing of `Linear()` and `Constant()` as you suggested.
``` julia
julia> Linear()
Linear()
julia> Linear(Throw(OnGrid()))
Linear()
julia> Linear(Periodic(OnCell()))
Linear(Periodic(OnCell()))
```

mkitti:
Could we add a test for `(deg::Linear)(gt::GridType)`?

mkitti:
We also need a test for `Base.show(io::IO, deg::Constant)`

test/b-splines/constant.jl
```diff
@@ -116,5 +116,53 @@
...
+        @test Constant() === Constant(Throw(OnGrid()))
+        @test Constant() isa Constant{Nearest,Throw{OnGrid}}
+        @test Constant(Periodic()) === Constant(Periodic(OnCell()))
+        @test Constant(Periodic()) isa Constant{Nearest,Periodic{OnCell}}
+        for T in (Nearest, Previous, Next)
+            it = Constant{T}()
+            @test it isa Constant{T, Throw{OnGrid}}
+            it = Constant{T}(Periodic())
+            @test it isa Constant{T, Periodic{OnCell}}
+        end
```
mkitti, line 131:
```suggestion
        for T in (Nearest, Previous, Next)
            it = Constant{T}()
            @test it isa Constant{T, Throw{OnGrid}}
            @test "$it" == "Constant{$T}()"
            it = Constant{T}(Periodic())
            @test it isa Constant{T, Periodic{OnCell}}
            @test "$it" == "Constant{$T}(Periodic(OnCell()))"
        end
```

src/b-splines/linear.jl
```diff
@@ -1,4 +1,14 @@
-struct Linear <: Degree{1} end  # boundary conditions not supported
+struct Linear{BC<:Union{Throw{OnGrid},Periodic{OnCell}}} <: DegreeBC{1}
+    bc::BC
+end
+
+(deg::Linear)(gt::GridType) = Linear(deg.bc(gt))
```
mkitti, line 5: Do we actually need this method?
```julia
julia> L = Linear()
Linear()

julia> L(OnGrid())
Linear()

julia> L(OnCell())
ERROR: MethodError: no method matching Linear(::Throw{OnCell})
...
```

test/b-splines/linear.jl
```diff
@@ -61,4 +61,39 @@
...
         @test sum(isnan.(etp)) == 0
     end
+
+    @testset "Linear periodic" begin
+        # Constructors
+        @test Linear() === Linear(Throw(OnGrid()))
+        @test Linear() isa Linear{Throw{OnGrid}}
+        @test Linear(Periodic()) === Linear(Periodic(OnCell()))
+        @test Linear(Periodic()) isa Linear{Periodic{OnCell}}
+
```
mkitti, line 71: If we need `(deg::Linear)(gt::GridType)`, test it here.

```suggestion
        @test Linear(Periodic()) isa Linear{Periodic{OnCell}}
        @test Linear() === Linear()(OnGrid())

```

mkitti (changes requested):
Here's a few suggestions for the tests. I'm not sure if we need `(deg::Linear)(gt::GridType)`.

mkitti:
What would you think of syntax that works as follows?

```julia
julia> import Interpolations: Periodic, GridType

julia> Periodic(::Type{GTT}) where GTT <: Union{GridType,Nothing} = Periodic(GTT())
Periodic

julia> itp = interpolate([1., 2., 3., 2.], OnCell |> Periodic |> Linear |> BSpline)
4-element interpolate(::Vector{Float64}, BSpline(Linear(Periodic(OnCell())))) with element type Float64:
 1.0
 2.0
 3.0
 2.0

```

test/b-splines/linear.jl
jaemolihm: I removed `(deg::Linear)(gt::GridType)`.

src/b-splines/linear.jl
jaemolihm: I agree, we do not need this method. (It was a leftover from copying from cubic.)

test/b-splines/constant.jl
jaemolihm: Done, thanks.

jaemolihm:
> ```julia
> Periodic(::Type{GTT}) where GTT <: Union{GridType,Nothing} = Periodic(GTT())
> ```

I guess this idea is from #401. I agree that piping constructor is more concise than parenthetical ones. Also, this issue is not limited to `Constant` or `Linear`. It applies to all `BSpline`.

As I understand, this new constructor is needed to make the first argument `OnGrid`, not `OnGrid()`. But I'm not sure that this is really desirable. All other constructors (like `Linear` or `BSpline`) accepts instances, not types, as arguments. So, I think `OnCell() |> Periodic |> Linear |> BSpline` is good enough.

For example, such a problem also occur when the starting point is `Periodic`, not `OnCell`. Then, one would also need to define `Cubic(::Type{BC}) where BC <: BoundaryCondition`, and also same thing for `Constant`, `Linear`, and `Quadratic`, which I think makes the code more complicated.
```julia
julia> Periodic() |> Cubic |> BSpline
BSpline(Cubic(Periodic()))

julia> Periodic |> Cubic |> BSpline
ERROR: MethodError: no method matching Cubic(::Type{Periodic})
Closest candidates are:
  Cubic(::BC) where BC<:Interpolations.BoundaryCondition at /home/jmlim/.julia/dev/Interpolations/src/b-splines/cubic.jl:2
Stacktrace:
 [1] |>(x::Type, f::Type{Cubic})
   @ Base ./operators.jl:858
 [2] top-level scope
   @ REPL[39]:1
```

(I do not know the Julia language design very well, so I may be wrong.)

src/b-splines/constant.jl
```diff
@@ -5,8 +5,25 @@ struct Nearest <: ConstantInterpType end
 struct Previous <: ConstantInterpType end
 struct Next <: ConstantInterpType end
 
-struct Constant{T<:ConstantInterpType} <: Degree{0} end
-Constant() = Constant{Nearest}()
+struct Constant{T<:ConstantInterpType,BC<:Union{Throw{OnGrid},Periodic{OnCell}}} <: DegreeBC{0}
+    bc::BC
```
johnnychen94, line 9: Is `bc` stateful here? Or could we always use `BC()` to construct it at runtime?

jaemolihm: https://github.com/JuliaMath/Interpolations.jl/blob/a8370bd2b4ef77cfe689c55c3ed217ab528f69c1/src/b-splines/b-splines.jl#L142
I am not sure I understand the question, but I think since the above line uses `deg.bc`, so `bc` should be a field.

Issue #438, opened 2021-07-02
support `gradient` for extrapolation with constant filling


johnnychen94:
Separated from the discussion in #434 and this is clearly an Interpolations issue.

When the extrapolation is constructed using a constant value, the out-of-range gradient is clearly `0`.

```julia
x = [1, 2, 3]
itp = interpolate(x, BSpline(Linear()))
etp = extrapolate(itp, 0)
etp(-1)
Interpolations.gradient(etp, -1) # not works
```

```julia
julia> Interpolations.gradient(etp, -1)
ERROR: MethodError: no method matching getfirst(::Int64)
Closest candidates are:
  getfirst(::Interpolations.Flag) at /Users/jc/Documents/Julia/Interpolations.jl/src/utils.jl:17
  getfirst(::Tuple) at /Users/jc/Documents/Julia/Interpolations.jl/src/utils.jl:18
Stacktrace:
 [1] _tcollect
   @ ~/Documents/Julia/Interpolations.jl/src/utils.jl:23 [inlined]
 [2] tcollect(f::typeof(Interpolations.etpflag), itp::Interpolations.FilledExtrapolation{Float64, 1, Interpolations.BSplineInterpolation{Float64, 1, Vector{Float64}, BSpline{Linear}, Tuple{Base.OneTo{Int64}}}, BSpline{Linear}, Int64})
   @ Interpolations ~/Documents/Julia/Interpolations.jl/src/utils.jl:22
 [3] gradient(etp::Interpolations.FilledExtrapolation{Float64, 1, Interpolations.BSplineInterpolation{Float64, 1, Vector{Float64}, BSpline{Linear}, Tuple{Base.OneTo{Int64}}}, BSpline{Linear}, Int64}, x::Int64)
   @ Interpolations ~/Documents/Julia/Interpolations.jl/src/extrapolation/extrapolation.jl:68
 [4] top-level scope
   @ REPL[67]:1
```

cc: @SomTambe @DhairyaLGandhi

SomTambe:
So I tackled it like this - 
```julia
struct ZeroRGB{GT <: Union{Nothing, Interpolations.GridType}} <: Interpolations.BoundaryCondition
    gt::GT
end

julia> function Interpolations.inbounds_index((flagl,flagu)::Tuple{Nothing,Union{Flat,Line,ZeroRGB}}, (l,u), x, etp, xN)
           Interpolations.minp(x,u)
       end

[...]

Issue #434, opened 2021-06-28
ImageTransformations issues


SomTambe:
I am writing backward passes for `ImageTransformations.warp!` and other Image agnostic functions. I have been solving most of the problems related to adjoints for Images, but after all I have gotten stuck on the Interpolations part.

I have arrived at this moment where I have this - 
```julia
julia> gradient((x,y)->rgb(sum(warp_me(x,y))), src2, h)
mn = [-348.0480263428803, -74.61757209029507]
mx = [-150.32291782157546, 116.97325016644439]
mn = [-348.0480263428803, -74.61757209029507]
mx = [-150.32291782157546, 116.97325016644439]
typeof(Δ) = FillArrays.Fill{NamedTuple{(:r, :g, :b), Tuple{Float64, Float64, Float64}}, 2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}
size(Δ) = (200, 193)
out_p.offsets = (-350, -76)
typeof(Δ) = OffsetMatrix{NamedTuple{(:r, :g, :b), Tuple{Float64, Float64, Float64}}, Matrix{NamedTuple{(:r, :g, :b), Tuple{Float64, Float64, Float64}}}}
ERROR: MethodError: no method matching getfirst(::RGB{N0f8})
```
I looked up the `getfirst` functions but they were literally giving out the first elements of a tuple. So I defined it accordingly for `RGB` -
```julia
julia> Interpolations.getfirst(f::RGB{T}) where T = f.r
```
Now it went ahead and gave me this -
```julia
julia> gradient((x,y)->rgb(sum(warp_me(x,y))), src2, h)
mn = [-348.0480263428803, -74.61757209029507]
mx = [-150.32291782157546, 116.97325016644439]
mn = [-348.0480263428803, -74.61757209029507]
mx = [-150.32291782157546, 116.97325016644439]
typeof(Δ) = FillArrays.Fill{NamedTuple{(:r, :g, :b), Tuple{Float64, Float64, Float64}}, 2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}
size(Δ) = (200, 193)
out_p.offsets = (-350, -76)
typeof(Δ) = OffsetMatrix{NamedTuple{(:r, :g, :b), Tuple{Float64, Float64, Float64}}, Matrix{NamedTuple{(:r, :g, :b), Tuple{Float64, Float64, Float64}}}}
[...]

PR #439, opened 2021-07-02
Add gradients for FilledExtrapolations

SomTambe:
Solves #438 and maybe #434. 
```julia
julia> x = rand(RGB,3,3)
3×3 Array{RGB{Float64},2} with eltype RGB{Float64}:
 RGB{Float64}(0.30544,0.136343,0.857564)    RGB{Float64}(0.34268,0.0164714,0.261636)  RGB{Float64}(0.0954148,0.197293,0.928313)
 RGB{Float64}(0.242728,0.960734,0.249229)   RGB{Float64}(0.414405,0.771848,0.286606)  RGB{Float64}(0.23209,0.623632,0.0502413)
 RGB{Float64}(0.921937,0.0466274,0.816069)  RGB{Float64}(0.781229,0.915429,0.926183)  RGB{Float64}(0.679568,0.737393,0.702759)

julia> itp2 = interpolate(x, BSpline(Linear()))
3×3 interpolate(::Array{RGB{Float64},2}, BSpline(Linear())) with element type RGB{Float64}:
 RGB{Float64}(0.30544,0.136343,0.857564)    RGB{Float64}(0.34268,0.0164714,0.261636)  RGB{Float64}(0.0954148,0.197293,0.928313)
 RGB{Float64}(0.242728,0.960734,0.249229)   RGB{Float64}(0.414405,0.771848,0.286606)  RGB{Float64}(0.23209,0.623632,0.0502413)
 RGB{Float64}(0.921937,0.0466274,0.816069)  RGB{Float64}(0.781229,0.915429,0.926183)  RGB{Float64}(0.679568,0.737393,0.702759)

julia> etp2 = extrapolate(itp2, RGB(.0,.0,.0))
3×3 extrapolate(interpolate(::Array{RGB{Float64},2}, BSpline(Linear())), RGB{Float64}(0.0,0.0,0.0)) with element type RGB{Float64}:
 RGB{Float64}(0.30544,0.136343,0.857564)    RGB{Float64}(0.34268,0.0164714,0.261636)  RGB{Float64}(0.0954148,0.197293,0.928313)
 RGB{Float64}(0.242728,0.960734,0.249229)   RGB{Float64}(0.414405,0.771848,0.286606)  RGB{Float64}(0.23209,0.623632,0.0502413)
 RGB{Float64}(0.921937,0.0466274,0.816069)  RGB{Float64}(0.781229,0.915429,0.926183)  RGB{Float64}(0.679568,0.737393,0.702759)

julia> Interpolations.gradient(etp2,3,3)
2-element StaticArrays.SVector{2, RGB{Float64}} with indices SOneTo(2):
 RGB{Float64}(0.4474772688335722,0.11376103826212525,0.6525178769801145)
 RGB{Float64}(-0.10166125750925725,-0.17803622073108794,-0.2234233554449263)

julia> Interpolations.gradient(etp2,4,4)
2-element StaticArrays.SVector{2, RGB{Float64}} with indices SOneTo(2):
 RGB{Float64}(0.0,0.0,0.0)
 RGB{Float64}(0.0,0.0,0.0)
```

### DIFF
--- a/src/extrapolation/filled.jl
+++ b/src/extrapolation/filled.jl
@@ -54,6 +54,16 @@ end
     return ret
 end
 
+@inline function Interpolations.gradient(etp::FilledExtrapolation{T}, x::Vararg{Number,N}) where {T,N}
+    if checkbounds(Bool, etp.itp, x...)
+        gradient(etp.itp, x...)
+    else
+        e = eltype(etp.itp)
+        ptype = promote_type(e, typeof(etp.fillvalue))
+        return SVector{ndims(etp), ptype}(zeros(ptype, ndims(etp)))
+    end
+end
+
 expand_index_resid_etp(deg, fillvalue, (l, u), x, etp::FilledExtrapolation, xN) =
     (l <= x <= u || Base.throw_boundserror(etp, xN))
 
--- a/src/extrapolation/filled.jl
+++ b/src/extrapolation/filled.jl
@@ -59,9 +59,8 @@ end
     if checkbounds(Bool, itp, x...)
         gradient(itp, x...)
     else
-        e = eltype(itp)
-        ptype = promote_type(e, mapreduce(typeof, promote_type, x))
-        return SVector{ndims(etp), ptype}(zeros(ptype, ndims(etp)))
+        ptype = promote_type(eltype(itp), map(typeof, x)...)
+        return zeros(SVector{ndims(etp), ptype})
     end
 end
 
--- a/src/extrapolation/filled.jl
+++ b/src/extrapolation/filled.jl
@@ -55,10 +55,11 @@ end
 end
 
 @inline function Interpolations.gradient(etp::FilledExtrapolation{T}, x::Vararg{Number,N}) where {T,N}
-    if checkbounds(Bool, etp.itp, x...)
-        gradient(etp.itp, x...)
+    itp = parent(etp)
+    if checkbounds(Bool, itp, x...)
+        gradient(itp, x...)
     else
-        e = eltype(etp.itp)
+        e = eltype(itp)
         ptype = promote_type(e, typeof(etp.fillvalue))
         return SVector{ndims(etp), ptype}(zeros(ptype, ndims(etp)))
     end
--- a/src/extrapolation/filled.jl
+++ b/src/extrapolation/filled.jl
@@ -60,7 +60,7 @@ end
         gradient(itp, x...)
     else
         e = eltype(itp)
-        ptype = promote_type(e, typeof(etp.fillvalue))
+        ptype = promote_type(e, mapreduce(typeof, promote_type, x))
         return SVector{ndims(etp), ptype}(zeros(ptype, ndims(etp)))
     end
 end
--- a/test/gradient.jl
+++ b/test/gradient.jl
@@ -202,16 +202,32 @@ using Test, Interpolations, DualNumbers, LinearAlgebra
 
     @testset "Extrapolations" begin
         # out of bounds extrapolation check where fillvalue=constant
-        dtypes = [Float32, Float64]
         dims = [1, 2, 3]
-        for dt in dtypes
+        @testset "Constant Fillvalue" begin
+            dtypes = [Float32, Float64]
+            for dt in dtypes
+                for dm in dims
+                    itp = interpolate(rand(dt, ntuple(_ -> 5, dm)...), BSpline(Linear()))
+                    etp = extrapolate(itp, rand(dt)) # TODO: Add tests for different boundary conditions in extrapolate
+                    cs = ntuple(_->3, dm)
+                    @test @inferred(Interpolations.gradient(etp, cs...)) == @inferred(Interpolations.gradient(itp, cs...))
+                    for ds in [ntuple(_->-0.5, dm), ntuple(_->5.5, dm), ntuple(_->6, dm)]
+                        @test all(iszero, @inferred(Interpolations.gradient(etp, ds...)))
+                    end
+                end
+            end
+        end
+        @testset "Boundary Conditions" begin
+            BC = (Flat, Line, Periodic, Reflect, Natural)
             for dm in dims
-                itp = interpolate(rand(dt, ntuple(_ -> 5, dm)...), BSpline(Linear()))
-                etp = extrapolate(itp, rand(dt))
-                cs = ntuple(_->3, dm)
-                @test @inferred(Interpolations.gradient(etp, cs...)) == @inferred(Interpolations.gradient(itp, cs...))
-                for ds in [ntuple(_->-0.5, dm), ntuple(_->5.5, dm), ntuple(_->6, dm)]
-                    @test all(iszero, @inferred(Interpolations.gradient(etp, ds...)))
+                for bc in BC
+                    itp = interpolate(rand(5, 5), BSpline(Linear()))
+                    etp = extrapolate(itp, bc()) # TODO: Add tests for different boundary conditions in extrapolate
+                    cs = ntuple(_->3, 2)
+                    @test @inferred(Interpolations.gradient(etp, cs...)) == @inferred(Interpolations.gradient(itp, cs...))
+                    for ds in [ntuple(_->-0.5, 2), ntuple(_->5.5, 2), ntuple(_->6, 2)]
+                        @test size(@inferred(Interpolations.gradient(etp, ds...))) == (2,)
+                    end
                 end
             end
         end
--- a/test/gradient.jl
+++ b/test/gradient.jl
@@ -206,25 +206,11 @@ using Test, Interpolations, DualNumbers, LinearAlgebra
         dims = [1, 2, 3]
         for dt in dtypes
             for dm in dims
-                itp_n = interpolate(rand(dt, ntuple(_ -> 5, dm)...), BSpline(Linear()))
-                etp_n = extrapolate(itp_n, rand(dt))
-                if dm==2
-                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, 6., 6.)))
-                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, -6., 6.)))
-                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, 6., -6.)))
-                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, -6., -6.)))
-                elseif dm==1
-                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, 6.)))
-                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, -6.)))
-                else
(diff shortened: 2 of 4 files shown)


SomTambe:
cc. @johnnychen94 @DhairyaLGandhi

johnnychen94 (commented):
this function is better moved to filled.jl

and most importantly, tests are needed.

mkitti:
Let me know when this is ready for review or needs a CI run

SomTambe:
For what types of constants should I test for here? The normal `FloatType` only right?

src/extrapolation/filled.jl
```diff
@@ -54,6 +54,16 @@ end
     return ret
 end
 
+@inline function Interpolations.gradient(etp::FilledExtrapolation{T}, x::Vararg{Number,N}) where {T,N}
+    if checkbounds(Bool, etp.itp, x...)
+        gradient(etp.itp, x...)
```
DhairyaLGandhi, line 59: `parent(etp)` instead of `etp.itp` to be symmetric with the other `gradient` function?

SomTambe: Okay that sounds fine.

johnnychen94 (commented):
> For what types of constants should I test for here? The normal FloatType only right?

FloatType should be sufficient enough. But since we're targeting DiffImages, we can add Colorant types as well (and also to `Flat()`/`Line` types). More test cases are always appreciated and help catches potential issues.

Please also add `@inferred` test to make sure type stability.

SomTambe:
I think we should do one thing - add the `FloatType` ones here, and the `Colorant` types in the DiffImages to ensure that any of these aren't breaking for our use. What do you think @johnnychen94 ?

johnnychen94:
> I think we should do one thing - add the FloatType ones here, and the Colorant types in the DiffImages to ensure that any of these aren't breaking for our use. What do you think @johnnychen94 ?

You won't test `Interpolations.gradient` in DIffImages, will you? The purpose of unit test is to break large things down into small parts, and test their basic functionality; only testing if the network forward/backward runs "normally" in its best scenario only makes sure things run correctly as an entire project. It doesn't tell you whether every single part of it runs normally, and it won't tell you which part should be checked when it goes wrong. In this sense, the unit test helps for future debugging. (You can't guarantee that you're still the maintainer three years later, and unit test helps future maintainers to understand things.)

It is often the case that people are too lazy to write tests, I understand it. And as I've said, more test cases are always appreciated and help catches potential issues.

SomTambe:
@mkitti I've added the tests, I think we are good to go for the CI.

mkitti:
Looks like we have some test issues based on type inference. The problem actually may not be in the new code, but older code. I'll have to take a closer look.

test/runtests.jl
```diff
@@ -1,3 +1,4 @@
+using Base: Float64, Float32
```
johnnychen94, line 1: This is unnecessary.

test/gradient.jl
```diff
@@ -199,4 +199,20 @@ using Test, Interpolations, DualNumbers, LinearAlgebra
...
     end
+
+    @testset "Extrapolations" begin
+        # out of bounds extrapolation check where fillvalue=constant
+        dtypes = [Float32, Float64]
+        dims = [1, 2, 3]
+        for dt in dtypes
+            for dm in dims
+                itp = interpolate(rand(dt, ntuple(_ -> 5, dm)...), BSpline(Linear()))
+                etp = extrapolate(itp, rand(dt))
```
johnnychen94, line 210: Let's also test in-range gradients are computed correctly.

```julia
@test @inferred(Interpolations.gradient(etp, 3, 3)) == @inferred(Interpolations.gradient(itp, 3, 3))
```

src/extrapolation/filled.jl
```diff
@@ -54,6 +54,17 @@ end
...
 
+@inline function Interpolations.gradient(etp::FilledExtrapolation{T}, x::Vararg{Number,N}) where {T,N}
+    itp = parent(etp)
+    if checkbounds(Bool, itp, x...)
+        gradient(itp, x...)
+    else
+        e = eltype(itp)
+        ptype = promote_type(e, typeof(etp.fillvalue))
+        return SVector{ndims(etp), ptype}(zeros(ptype, ndims(etp)))
+    end
```
johnnychen94, line 65: The type instability (`@inferred` failures) comes from the fact that this if-else block might return different types, hence @SomTambe you might need to double-check if `ptype` is appropriate. See https://docs.julialang.org/en/v1/manual/performance-tips/#Write-%22type-stable%22-functions for more explanation.

cc: @mkitti

johnnychen94 (commented):
It's not required for the scope of this PR but it is appreciated to also add the same test cases for extrapolation with other schemes; they're currently missing as far as I can tell.

Once the test passes I believe it is good to merge.

SomTambe:
Oh I just realized I made a really bad error while making the tests. Please wait until I update them.

test/gradient.jl
```diff
@@ -206,12 +206,26 @@ using Test, Interpolations, DualNumbers, LinearAlgebra
...
+                else
+                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, 6., 6., 6.)))
+                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, -6., 6., 6.)))
+                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, 6., -6., 6.)))
+                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, -6., -6., 6.)))
+                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, 6., 6., -6.)))
+                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, -6., 6., -6.)))
+                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, 6., -6., -6.)))
+                    @test all(iszero, @inferred(Interpolations.gradient(etp_n, -6., -6., -6.)))
+                end
```
johnnychen94, line 228: These two seem sufficient to me.

```julia
for x in [ntuple(_->0, dm), ntuple(_->6, dm)]
```

SomTambe: But this will fall short for the 2-dim and 3-dim cases where we have multiple quadrants right? It's better to get it checked for every quadrant right? That is why I hardcoded them.

johnnychen94: Yeah I get your points but it's not really necessary. We only need to choose _some_ out of range points and test it. This is a more programmable way and is more friendly for future maintainers.

SomTambe: Ok I have changed it.

src/extrapolation/filled.jl
```diff
@@ -60,7 +60,7 @@ end
         gradient(itp, x...)
     else
         e = eltype(itp)
-        ptype = promote_type(e, typeof(etp.fillvalue))
+        ptype = promote_type(e, mapreduce(typeof, promote_type, x))
```
DhairyaLGandhi, line 63: Can't we let Julia's type system figure out the types? Arebts things like `itp` already parameterized?

johnnychen94: Not really, this is because interpolations gradient use `x`(the indices) to generate `δx` and thus the gradient output, to keep type stability we need to keep consistent to interpolation gradient internal.

src/extrapolation/filled.jl
```diff
@@ -54,6 +54,17 @@ end
...
 end
 
+@inline function Interpolations.gradient(etp::FilledExtrapolation{T}, x::Vararg{Number,N}) where {T,N}
+    itp = parent(etp)
+    if checkbounds(Bool, itp, x...)
+        gradient(itp, x...)
+    else
+        e = eltype(itp)
+        ptype = promote_type(e, mapreduce(typeof, promote_type, x))
+        return SVector{ndims(etp), ptype}(zeros(ptype, ndims(etp)))
```
johnnychen94, line 64: `zeros(ptype, ndims(etp))` is unnecessary intermediate memory allocation, let's use tuple here.

```suggestion
        ptype = promote_type(eltype(itp), map(typeof, x)...)
        return SVector{ndims(etp), ptype}(ntuple(_->zero(ptype), ndims(etp)))
```

```julia
julia> etp = extrapolate(interpolate([1, 2, 3], BSpline(Linear())), 0);

# before
julia> @btime Interpolations.gradient($etp, -0.5f0);
  35.995 ns (1 allocation: 96 bytes)
# after
julia> @btime Interpolations.gradient($etp, -0.5f0);
  0.038 ns (0 allocations: 0 bytes)
```

SomTambe:
@mkitti Could you run CI on this? I have corrected type instability and also added more tests for the other boundary conditions as there weren't any written for them.

johnnychen94 (approved):

mkitti:
Looks good. Merging. Thank you.

SomTambe:
The code is breaking for these types - 
```julia
julia> x = rand(RGB,3,3);

julia> itp2 = interpolate(x, BSpline(Linear()));

julia> etp2 = extrapolate(itp2, RGB(.0,.0,.0));

julia> Interpolations.gradient(etp2,3,3)
2-element SVector{2, RGB{Float64}} with indices SOneTo(2):
 RGB{Float64}(0.43457141859445714,0.1577128433792203,0.02876792599552047)
 RGB{Float64}(0.19412021621096764,-0.5215101620517046,0.36728033563097706)

julia> Interpolations.gradient(etp2,4,4) # wrong, should be of types RGB{Float64}
2-element SVector{2, Any} with indices SOneTo(2):
 0.0
 0.0
```
This is coming from the lines [62](https://github.com/JuliaMath/Interpolations.jl/blob/master/src/extrapolation/filled.jl#L62) in `Filled.jl` - 
```julia
...
ptype = promote_type(eltype(itp), map(typeof, x)...)
...
```
This renders `ptype` to be `Any` instead of being `RGB{Float64}`, which is further causing problems in gradients to propogate in DiffImages.jl. 
I wonder how I missed this.
cc. @johnnychen94 @DhairyaLGandhi

johnnychen94:
> FloatType should be sufficient enough. But since we're targeting DiffImages, we can add Colorant types as well (and also to Flat()/Line types). More test cases are always appreciated and help catches potential issues.

A typical case that more test matters.

This is because Numbers can't be promoted to Colorants unless specified:

```julia
julia> promote_type(RGB{Float32}, Int)
Any # We would expect this to be RGB{Float32}
```

The reason we use `promote_type` is to mimic the output type of `gradient` for *Interpolation for type stability. Since this trick doesn't work, we might need to look deeper into the codebase to see how we can get that.

> Can't we let Julia's type system figure out the types?

Back to the comment, @DhairyaLGandhi do you mean to use `Base._return_type` to get the inferred output type?

DhairyaLGandhi:
> did you mean to use Base._return_type to get the inferred output type?

Yes, or change the method signature to use `Vararg{S} where S`. Are there common cases where indices would not be the same type? I would consider that a bug if anything.

PR #442, opened 2021-07-13
support zero-arg parametric constructor for Constant and Linear

johnnychen94:
ImageTransformations uses [`typeof(deg)()` internally](https://github.com/JuliaImages/ImageTransformations.jl/blob/c66b77a7e742ddd3f592395a2f983492a290e3ed/src/compat.jl#L27-L34) and it assumes the zero argument version constructor `Linear{Throw{OnGrid}}()` and `Constant{Nearest,Throw{OnGrid}}()` exist. So Interpolations v0.13.3 unfortunately breaks ImageTransformations 😢 

I'm not sure if the `bc` field is necessary. (See my comments in https://github.com/JuliaMath/Interpolations.jl/pull/428/files#r668457028). This patch assumes that `bc` is needed and I added these constructor methods.

---

https://github.com/JuliaImages/ImageTransformations.jl/pull/132 fixes the issue in ImageTransformations side, but I believe this is still useful.

### DIFF
--- a/src/b-splines/constant.jl
+++ b/src/b-splines/constant.jl
@@ -7,6 +7,9 @@ struct Next <: ConstantInterpType end
 
 struct Constant{T<:ConstantInterpType,BC<:Union{Throw{OnGrid},Periodic{OnCell}}} <: DegreeBC{0}
     bc::BC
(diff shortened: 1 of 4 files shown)


mkitti:
Do we need a new release soon to address this?

johnnychen94:
No need, I've fixed ImageTransformations by directly using the field value.

PR #444, opened 2021-07-15
Fix docstring issue in iterate docs

awadell1:
Fixes #443

Issue #443, opened 2021-07-15
Doctest failures


mkitti:
```
$ julia docs/make.jl
...
┌ Error: doctest failure in src/iterate.md:17-27
│ 
│ ```jldoctest iterate-interpolate; setup = :(using Interpolations)
│ julia> itp = interpolate(ones(3,3), BSpline(Linear()));
│ 
│ julia> kiter = knots(itp);
│ 
│ julia> collect(kiter)
│ 3×3 Array{Tuple{Int64,Int64},2}:
│  (1, 1)  (1, 2)  (1, 3)
│  (2, 1)  (2, 2)  (2, 3)
│  (3, 1)  (3, 2)  (3, 3)
│ ```
│ 
│ Subexpression:
│ 
│ collect(kiter)
│ 
│ Evaluated output:
│ 
│ 3×3 Matrix{Tuple{Int64, Int64}}:
│  (1, 1)  (1, 2)  (1, 3)
│  (2, 1)  (2, 2)  (2, 3)
│  (3, 1)  (3, 2)  (3, 3)
│ 
│ Expected output:
│ 
│ 3×3 Array{Tuple{Int64,Int64},2}:
│  (1, 1)  (1, 2)  (1, 3)
│  (2, 1)  (2, 2)  (2, 3)
│  (3, 1)  (3, 2)  (3, 3)
│ 
│   diff =
│    3×3 Array{Tuple{Int64,Int64},2}:
│     Matrix{Tuple{Int64, Int64}}:
│     (1, 1)  (1, 2)  (1, 3)
│     (2, 1)  (2, 2)  (2, 3)
│     (3, 1)  (3, 2)  (3, 3)
└ @ Documenter.DocTests ~/.julia/packages/Documenter/4JDQo/src/DocTests.jl:385
┌ Error: doctest failure in src/iterate.md:60-75
│ 
│ ```jldoctest periodic-demo; setup = :(using Interpolations)
│ julia> x = [1.0, 1.5, 1.75, 2.0];
│ 
│ julia> etp = LinearInterpolation(x, x.^2, extrapolation_bc=Periodic());
│ 
│ julia> kiter = knots(etp);
│ 
│ julia> k = Iterators.take(kiter, 6) |> collect
│ 6-element Array{Float64,1}:
│  1.0
│  1.5
│  1.75
│  2.0
│  2.5
│  2.75
│ 
│ ```
│ 
│ Subexpression:
│ 
[...]

PR #445, opened 2021-07-15
Enable Strict Doctests

awadell1:
Enable strict doctests so that failed doctests trigger a non-zero exit code. Currently, doctests errors (#443) do not cause the Documentation workflow to fail, which reduces its effectiveness as a check. 

To prevent non-zero exit codes for the missing `l4_2d_cs` docstring, I swapped the block comment for line comments.

Note, the Documentation workflow is expected to fail, until #444 or similar is merged in.

If it's easier, I can cherry-pick this over to #444, but wanted to give the opportunity to delay this changeset if desired.

Best,

Alex


johnnychen94 (approved):
IMO having CI complain about the documentation incorrectness is always better than having some real person to do so.

PR #446, opened 2021-07-16
Make gradient fixes

SomTambe:
This fixes the issue I mentioned in this [comment](https://github.com/JuliaMath/Interpolations.jl/pull/439#issuecomment-881361834).
I'll add the tests in a bit.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -39,6 +39,14 @@ jobs:
             ${{ runner.os }}-test-${{ env.cache-name }}-
             ${{ runner.os }}-test-
             ${{ runner.os }}-
+      - name: "Compat fix for Julia < v1.3.0"
+        if: ${{ matrix.julia-version == '1.0' }}
+        run: |
+          using Pkg
+          Pkg.add([
+            PackageSpec(name="AbstractFFTs", version="1.0"),
+          ])
+        shell: julia --project=. --startup=no --color=yes {0}
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
       - uses: julia-actions/julia-processcoverage@v1
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -39,14 +39,6 @@ jobs:
             ${{ runner.os }}-test-${{ env.cache-name }}-
             ${{ runner.os }}-test-
             ${{ runner.os }}-
-      - name: "Compat fix for Julia < v1.3.0"
-        if: ${{ matrix.julia-version == '1.0' }}
-        run: |
-          using Pkg
-          Pkg.add([
-            PackageSpec(name="AbstractFFTs", version="1.0"),
-          ])
-        shell: julia --project=. --startup=no --color=yes {0}
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
       - uses: julia-actions/julia-processcoverage@v1
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.13.3"
 [deps]
 AxisAlgorithms = "13072b0f-2c55-5437-9ae7-d433b7a33950"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+ImageCore = "a09fc81d-aa75-5fe9-8630-4744c3626534"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
@@ -26,6 +27,7 @@ WoodburyMatrices = "0.4, 0.5"
 julia = "1"
 
 [extras]
+ImageCore = "a09fc81d-aa75-5fe9-8630-4744c3626534"
 DualNumbers = "fa6b7ba4-c1ee-5f82-b5fc-ecf0adba8f74"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
@@ -37,4 +39,4 @@ Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [targets]
-test = ["OffsetArrays", "Unitful", "SharedArrays", "ForwardDiff", "LinearAlgebra", "DualNumbers", "Random", "Test", "Zygote"]
+test = ["OffsetArrays", "Unitful", "SharedArrays", "ForwardDiff", "LinearAlgebra", "DualNumbers", "Random", "Test", "Zygote", "ImageCore"]
--- a/Project.toml
+++ b/Project.toml
@@ -5,7 +5,6 @@ version = "0.13.3"
 [deps]
 AxisAlgorithms = "13072b0f-2c55-5437-9ae7-d433b7a33950"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
-ImageCore = "a09fc81d-aa75-5fe9-8630-4744c3626534"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
@@ -27,7 +26,6 @@ WoodburyMatrices = "0.4, 0.5"
 julia = "1"
 
 [extras]
-ImageCore = "a09fc81d-aa75-5fe9-8630-4744c3626534"
 DualNumbers = "fa6b7ba4-c1ee-5f82-b5fc-ecf0adba8f74"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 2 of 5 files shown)


SomTambe:
It is type unstable.
cc. @johnnychen94 @dhairyalgandhi

src/extrapolation/filled.jl
```diff
@@ -54,12 +54,12 @@ end
     return ret
 end
 
-@inline function Interpolations.gradient(etp::FilledExtrapolation{T}, x::Vararg{Number,N}) where {T,N}
-    itp = parent(etp)
+@inline function Interpolations.gradient(etp::FilledExtrapolation{T, N, ITP}, x::Vararg{Number, N}) where {T,N,ITP}
+    itp = parent(etp)::ITP
```
DhairyaLGandhi, line 58: We can forgo this since it's anyway defined in the method def

mkitti:
My sense is that a simple fix might not be possible. We might need a more sophisticated mechanism to decide when the arguments might affect the return type. It did not start with your changes

mkitti:
Huh. I'll merge it if this works for you. I'll have to study this more closely.

johnnychen94:
every non-trivial PR should have associated tests

mkitti:
That's a good point. You should add a test for the original issue you cited.

src/extrapolation/filled.jl
```diff
@@ -54,12 +54,13 @@ end
     return ret
 end
 
-@inline function Interpolations.gradient(etp::FilledExtrapolation{T}, x::Vararg{Number,N}) where {T,N}
+@inline function Interpolations.gradient(etp::FilledExtrapolation{T, N, ITP}, x::Vararg{Number, N}) where {T,N,ITP}
```
dkarrasch, line 57: Why do you need the `ITP` parameter here?
```suggestion
@inline function Interpolations.gradient(etp::FilledExtrapolation{T, N}, x::Vararg{Number, N}) where {T,N}
```

test/runtests.jl
```diff
@@ -3,7 +3,7 @@ if !isdefined(Main, :InterpolationTestUtils)
     @eval using Main.InterpolationTestUtils
 end
 
-using Test, SharedArrays, Random
+using Test, SharedArrays, Random, Images
```
johnnychen94, line 6: `Images` is too big a dependency, just `ImageCore` and/or `ColorVectorSpace` should be sufficient enough. And should also need to modify `Project.toml` to add it as test dependency.

src/extrapolation/filled.jl
```diff
@@ -54,12 +54,13 @@ end
...
     itp = parent(etp)
+    check_type() = Base._return_type(gradient, Tuple{typeof(itp), map(typeof, x)...})
     if checkbounds(Bool, itp, x...)
         gradient(itp, x...)
     else
-        ptype = promote_type(eltype(itp), map(typeof, x)...)
+        ptype = eltype(check_type())
         return zeros(SVector{ndims(etp), ptype})
```
johnnychen94, line 64: Actually, I don't know why wrapping as a function makes it type stable. Any ideas?

---

```suggestion
        return zeros(check_type())
```

src/extrapolation/filled.jl
johnnychen94: I guess it was meant to use `ITP` instead of `typeof(itp)`

mkitti: `ITP == typeof(itp)`. I'm guessing that affects the call `Base._return_type` which in turn affects the return type of the method. Therefore, one might need to parameterize on `ITP` for type stability of the overall method.

src/extrapolation/filled.jl
SomTambe: [This](https://discourse.julialang.org/t/could-base-return-type-ever-be-inferable/8057) thread did shed some insight as to how to make `Base._return_type` inferable, but I still don't quite firmly understand how did that make it stable.

src/extrapolation/filled.jl
```diff
@@ -54,12 +54,13 @@ end
     return ret
 end
 
-@inline function Interpolations.gradient(etp::FilledExtrapolation{T}, x::Vararg{Number,N}) where {T,N}
+@inline function Interpolations.gradient(etp::FilledExtrapolation{T, N, ITP}, x::Vararg{Number, N}) where {T,N,ITP}
     itp = parent(etp)
+    check_type() = Base._return_type(gradient, Tuple{typeof(itp), map(typeof, x)...})
```
DhairyaLGandhi, line 59: Why is this split off as a closure?

test/runtests.jl
SomTambe: Done.

SomTambe:
I have no idea what `ERROR: Unsatisfiable requirements detected for package FFTW [7a1cc6ca]:` means. I added a dependency on `ImageCore` as Johnny suggested Images would be a heavy/circular dependency. Is it creating the problem? I don't think so.

Project.toml
```diff
@@ -5,6 +5,7 @@ version = "0.13.3"
 [deps]
 AxisAlgorithms = "13072b0f-2c55-5437-9ae7-d433b7a33950"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+ImageCore = "a09fc81d-aa75-5fe9-8630-4744c3626534"
```
johnnychen94, line 8: This is test dependency, so should be listed in `[extras]` and `[targets.test]`

johnnychen94 (commented):
> I have no idea what ERROR: Unsatisfiable requirements detected for package FFTW [7a1cc6ca]: means.

This is an issue with the old Pkg version; it fails to resolve the possible version when there is one. You can assist the Pkg resolver by coping [this part](https://github.com/JuliaImages/ImageCore.jl/blob/7ab34caa31eb8d489659cb0e3c553e2403f2206a/.github/workflows/UnitTest.yml#L41-L48) to the `github/workflows/CI.yml` configuration file; that basically says: okay I need this version of FFTW, you Pkg do the rest and then it finds out one.

mkitti:
Are we picking up a dependency on FFTW?

johnnychen94:
> Are we picking up a dependency on FFTW?

No, but ImageCore and some other packages (e.g., Zygote) depend on it. We just need to manually assist the pkg resolving by explicitly adding them during CI initialization for old Julia versions.

DhairyaLGandhi:
That's a bit fiddly CI wise - would Interpolations.jl accept a lower bound to Julia 1.3 or 1.6?

What if 1.6 is LTS?

SomTambe:
Wow, changing it to `ColorVectorSpace` did the job! I think the PR is good to go!

johnnychen94:
CVS should be in [extras] and not in [deps]

Project.toml is designed in a way that we can manually edit it. You can read more on http://pkgdocs.julialang.org/v1/toml-files/

.github/workflows/CI.yml
```diff
@@ -39,6 +39,14 @@ jobs:
...
             ${{ runner.os }}-test-
             ${{ runner.os }}-
+      - name: "Compat fix for Julia < v1.3.0"
+        if: ${{ matrix.julia-version == '1.0' }}
+        run: |
+          using Pkg
+          Pkg.add([
+            PackageSpec(name="AbstractFFTs", version="0.5"),
+          ])
+        shell: julia --project=. --startup=no --color=yes {0}
```
johnnychen94, line 49: Actually, since this CI file uses `version` instead of `julia-version`, this stage will never be executed.

Project.toml
```diff
@@ -28,13 +28,15 @@ julia = "1"
...
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SharedArrays = "1a1011a3-84de-559e-8e89-a11a2f7dc383"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
+ColorVectorSpace = "c3611d14-8923-5661-9e6a-0046d554d3a4"
 
 [targets]
-test = ["OffsetArrays", "Unitful", "SharedArrays", "ForwardDiff", "LinearAlgebra", "DualNumbers", "Random", "Test", "Zygote"]
+test = ["OffsetArrays", "Unitful", "SharedArrays", "ForwardDiff", "LinearAlgebra", "DualNumbers", "Random", "Test", "Zygote", "ImageCore"]
```
johnnychen94, line 42: This is strange, are you sure `ColorVectorSpace` is used as a __direct__ dependency in the test?

johnnychen94 (commented):
I'm not convinced that adding CVS is the right fix. It's better to test these change locally first on Julia-1.0.

Project.toml
SomTambe: I overlooked this, should've removed the `ImageCore` from `[targets]` while trying with CVC.

.github/workflows/CI.yml
SomTambe: I did not know that. Should I remove CVC and try with ImageCore only with the changes you have suggested?

johnnychen94: I don't know, but the purpose of this stage is to assist the old Pkg version to figure out an existing solution. If we get a green check without this stage executed, then it is not needed.

johnnychen94 (approved):

SomTambe:
I will first squash all these previous commits, and then we shall merge.

test/gradient.jl
```diff
@@ -204,7 +205,7 @@ using Test, Interpolations, DualNumbers, LinearAlgebra
         # out of bounds extrapolation check where fillvalue=constant
         dims = [1, 2, 3]
         @testset "Constant Fillvalue" begin
-            dtypes = [Float32, Float64]
+            dtypes = [Float32, Float64, RGB{Float64}, RGB{Float32}, RGB{N0f8}]
```
DhairyaLGandhi, line 208: Add a Gray?

SomTambe:
I have messed up the rebase badly. I'll fix it asap :cry:

test/gradient.jl
SomTambe: Yeah that should be there. I'll add it.

SomTambe:
I think the PR is finally good to go!
Sorry for taking the time :sweat_smile:

johnnychen94 (approved):
Need squash and merge.

PR #451, opened 2021-07-27
Improve precision of Lanczos OpenCV4

timholy:
Formerly Lanczos OpenCV4 could return `NaN` when rounding
yielded truncation. This was hidden by a special rule for integer
shifts, but is revealed by choosing any shift that is within 4eps of
an integer.

### DIFF
--- a/src/lanczos/lanczos.jl
+++ b/src/lanczos/lanczos.jl
@@ -120,12 +120,13 @@ function _lanczos4_opencv(δx)
     s0, c0 = sincos(y0)
     cs = ntuple(8) do i
         y = (δx + 4 - i) * p_4
-        # Numerator is the sin subtraction identity
-        # It is equivalent to the following
-        # f(δx,i) = sin( π/4*( 5*(i-1)-δx-3 ) )
+        # Improve precision of Lanczos OpenCV4 #451, avoid NaN
         if iszero(y)
             y = eps(oneunit(y))/8
         end
+        # Numerator is the sin subtraction identity
+        # It is equivalent to the following
+        # f(δx,i) = sin( π/4*( 5*(i-1)-δx-3 ) )
         (l4_2d_cs[i, 1] * s0 + l4_2d_cs[i, 2] * c0) / y^2
     end
     sum_cs = sum(cs)
--- a/src/lanczos/lanczos.jl
+++ b/src/lanczos/lanczos.jl
@@ -14,7 +14,7 @@ This form of interpolation is merely the discrete convolution of the samples wit
 """
 struct Lanczos{N} <: AbstractLanczos
     a::Int
-    
+
     function Lanczos{N}(a) where N
         N < a && @warn "Using a smaller support than scale for Lanczos window. Proceed with caution."
         new{N}(a)
@@ -103,7 +103,7 @@ struct Lanczos4OpenCV <: AbstractLanczos end
 
 degree(::Lanczos4OpenCV) = 4
 
-value_weights(::Lanczos4OpenCV, δx::S) where S = ifelse(isinteger(δx),ntuple(i->convert(float(S), i == 4 - δx), Val(8)) ,_lanczos4_opencv(δx))
+value_weights(::Lanczos4OpenCV, δx) = _lanczos4_opencv(δx)
 
 # s45 = sqrt(2)/2
 const s45 = 0.70710678118654752440084436210485
@@ -123,6 +123,9 @@ function _lanczos4_opencv(δx)
         # Numerator is the sin subtraction identity
         # It is equivalent to the following
         # f(δx,i) = sin( π/4*( 5*(i-1)-δx-3 ) )
+        if iszero(y)
+            y = eps(oneunit(y))
+        end
         (l4_2d_cs[i, 1] * s0 + l4_2d_cs[i, 2] * c0) / y^2
     end
     sum_cs = sum(cs)
--- a/src/lanczos/lanczos.jl
+++ b/src/lanczos/lanczos.jl
@@ -124,7 +124,7 @@ function _lanczos4_opencv(δx)
         # It is equivalent to the following
         # f(δx,i) = sin( π/4*( 5*(i-1)-δx-3 ) )
         if iszero(y)
-            y = eps(oneunit(y))
+            y = eps(oneunit(y))/8
         end
         (l4_2d_cs[i, 1] * s0 + l4_2d_cs[i, 2] * c0) / y^2
     end
--- a/test/lanczos/runtests.jl
+++ b/test/lanczos/runtests.jl
@@ -53,6 +53,7 @@ end
 
     @test 1 < itp(1.5) < 2
     @test 99 < itp(99.5) < 100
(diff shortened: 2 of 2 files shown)


timholy:
CC @mileslucas

mkitti:
I'm curious if the original version has a similar issue:
https://github.com/opencv/opencv/blob/de15636724967faf62c2d1bce26f4335e4b359e5/modules/imgproc/src/resize.cpp#L917-L946

timholy:
Nope, due to https://github.com/opencv/opencv/blob/de15636724967faf62c2d1bce26f4335e4b359e5/modules/imgproc/src/resize.cpp#L928

mileslucas (approved):
Looks good to me!

src/lanczos/lanczos.jl
```diff
@@ -123,6 +123,9 @@ function _lanczos4_opencv(δx)
         # Numerator is the sin subtraction identity
         # It is equivalent to the following
         # f(δx,i) = sin( π/4*( 5*(i-1)-δx-3 ) )
+        if iszero(y)
+            y = eps(oneunit(y))/8
+        end
```
mkitti, line 128:
```suggestion
        # Improve precision of Lanczos OpenCV4 #451, avoid NaN
        if iszero(y)
            y = eps(oneunit(y))/8
        end
        # Numerator is the sin subtraction identity
        # It is equivalent to the following
        # f(δx,i) = sin( π/4*( 5*(i-1)-δx-3 ) )
```

test/lanczos/runtests.jl
```diff
@@ -53,6 +53,7 @@ end
 
     @test 1 < itp(1.5) < 2
     @test 99 < itp(99.5) < 100
+    @test itp(nextfloat(1.0)) ≈ itp(1.0)
```
mkitti, line 56:
```suggestion
    # Test precision of Lanczos OpenCV4 #451, test not NaN
    @test itp(nextfloat(1.0)) ≈ itp(1.0)
```

mkitti (changes requested):
I made some suggestions to add some comments and keep prior comments adjacent to the line of code they are commenting on.

mkitti:
@timholy , @mileslucas any objections to applying my comment suggestions and merging?

timholy:
Thanks for merging! Can we get a new release? https://github.com/JuliaImages/ImageTransformations.jl/pull/143 needs this (CC @johnnychen94).

timholy:
BTW, feel free to squash-merge PRs like this. Sometimes I make PRs that have been carefully crafted to have independent commits that I'd like to keep separated (it makes it easier to revert specific changes), but this was not one of them.

mkitti:
I tend to use squash sparingly because I like to see the original commit history, and I use blame a lot to figure out why things are the way they are. Let me clear the ChainRuleCore compat PRs, and then I'll release.

timholy:
It certainly is safer not to squash, esp if you aren't sure whether the author intended to keep things separate.

johnnychen94:
> I tend to use squash sparingly because I like to see the original commit history, and I use blame a lot to figure out why things are the way they are.

The master history really looks chaotic to me and doesn't have a clue with change belongs to which PR because a lot of commits are made very casually. For example, I have no idea what this part is about and I'm sure you'll have the same feeling, say, one year later.



A PR with long back-and-forth commits should just be squashed before merging.

The practice I used most often is to use PR for every functionality changes even if as a maintainer, there are two main advantages:

- By opening a PR there will be a notification to others' inboxes, so you get a chance to receive reviews from others. For example, I'll wait for one or more days for not-so-trivial changes when I know they're active people watching it. Thus you get more chances to fix things before merging.
- It's a good practice to make PR an atomic functionality change, which is why we squash commits before merging. By making PR more atomic you get more information by traveling back using git blame because you get pointed to one specific PR with a bunch of discussion history.

FWIW, julialang/julia also adopts the same squash-and-merge by default, see: https://github.com/JuliaLang/julia/blob/master/CONTRIBUTING.md#git-recommendations-for-pull-request-reviewers

---

I'd like also to share a well-written blog written by `@oxinabox` [Continuous Delivery For Julia Packages](https://www.oxinabox.net/2019/09/28/Continuous-Delivery-For-Julia-Packages.html) and I think it's very well-argued that it's a good practice to release more often.

> ### Why do this?
> In enterprise continuous delivery is a well-established Agile practice that you can read a bunch about. So I am not going to talk about that here, you can google for it. In open-source there are additional advantages: you avoid annoying your contributors.
> 
> There are few things as frustrating as the following tale. Someone opens an issue on one of my projects. I determine that it is caused by an unstream bug in a library I am using. I fix that bug, and have my patch merged. I think all is done and that it will no doubt solved for the end-user shortly. A month later someone else opens the same issue on my project. Because no release has been made on the library I fixed. I have to go back over to the library, open issues nagging them to release. Maybe even send a maintainer an direct message. Often the reason relates to them thinking another core maintainer wants to hold off (which might be true, for better or worse.) By sticking to a policy of continous delivery, this can be avoided. It is clear to everyone that when a PR gets made, a release is assumed to be shortly after.
> 
> Your contributors are making PRs because they want the feature/bugfix. You have already indicated you are happy with it by merging, why hold off releasing it to the world?

mkitti:
@johnnychen94 I'll take it into consideration.

Most of it is really about time which is at a premium right now. If you are interested in being a co-maintainer let me know, even if it is just to release more frequently.

Whether to squash or not is dependent for me on the size of the repository. I find it easier to go from commit to PR than to try to expand out a PR into commits that were squashed into it. Github has a feature that tells you the associated pull request for any commit.
image

The main issue I have with squashing is that it rewrites and hides the actual history. This can be problematic and often can hide the reason for certain changes. Version control was introduced to me in a scientific context, so I often see the commit history as the analog to laboratory notebook. Like a commit history a lab notebook can often get quite messy and include erroneous thoughts, but from a scientific perspective it is important to preserve that thought process verbatim. Replacing the contemporaneous notes in the lab notebook with a cleanly organized summary would be tantamount to scientific fraud.

Sometimes the commit history is long, meandering, and is really just detailed debugging session, perhaps like the one in your screenshot, although personally I don't see the problem there either. Honestly in that specific case I forgot to consider squashing. For larger projects, I acknowledge that not squashing is just too unwieldy.

I think there are benefits to not squashing as well:
https://blog.ploeh.dk/2020/10/05/fortunately-i-dont-squash-my-commits/

For this specific pull request, I actually really like that https://github.com/JuliaMath/Interpolations.jl/pull/451/commits/378cd50ef4f20a3bd9b44e0bb9c029581d61c6a9 along with the previous CI run explains why the divide by `8` exists. In other contexts, such as 31f6d54fb8d9eab0511a1c50182025b0a448c9f1, I have squashed at your suggestion. Generally, though my preference is not to squash for the reasons I explained above.

Continuously delivery is fine. If I had the release process more automated, I would probably release more often. It would be really frustrating if the main Julia repository did continuous delivery though. This is really more due to time constraint.

In summary, some of it is due to preference but most of it is to due to time. Both reduce down to me being the sole maintainer at the moment.

Issue #403, opened 2021-02-24
Simple, concrete example in docs


OliverEvans96:
Hello,

There's a whole page of benchmarks front-and-center in the README, but I've been searching for 10 minutes now to figure out how to do a 1D linear interpolation of a set of points similar to scipy's [interp1d](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html#scipy.interpolate.interp1d).

[This page](http://juliamath.github.io/Interpolations.jl/latest/interpolations/) discusses (but doesn't give an example of) the first argument to `interpolate` but not the required options.

[This page](http://juliamath.github.io/Interpolations.jl/latest/control/) discusses the options but not the data (until the section on Parametric Splines, which is clear but not what I'm trying to do).

Also, [The tests](https://github.com/JuliaMath/Interpolations.jl/blob/master/test/b-splines/linear.jl) are abstracted such that they do not contain a simple, concrete example either.

The [equivalent scipy documentation](https://docs.scipy.org/doc/scipy/reference/tutorial/interpolate.html#d-interpolation-interp1d) is abundantly clear and easy to find.

```python
from scipy.interpolate import interp1d

x = np.linspace(0, 10, num=11, endpoint=True)
y = np.cos(-x**2/9.0)
f = interp1d(x, y)
f2 = interp1d(x, y, kind='cubic')
xnew = np.linspace(0, 10, num=41, endpoint=True)

import matplotlib.pyplot as plt
plt.plot(x, y, 'o', xnew, f(xnew), '-', xnew, f2(xnew), '--')
plt.legend(['data', 'linear', 'cubic'], loc='best')
plt.show()
```

![plot](https://docs.scipy.org/doc/scipy/reference/_images/interpolate-1.png)

Sorry to complain, but hopefully it's helpful complaining :)
Oliver

OliverEvans96:
The first [Dierckx README](https://github.com/kbarbary/Dierckx.jl#example-usage) is also great (but I'd rather avoid a fortran wrapper).

Example Usage
-------------

```julia
using Dierckx
```

Fit a 1-d spline to some input data (points can be unevenly spaced):

```julia
x = [0., 1., 2., 3., 4.]
y = [-1., 0., 7., 26., 63.]  # x.^3 - 1.
spl = Spline1D(x, y)
```

Evaluate the spline at some new points:

```julia
spl([1.5, 2.5])  # result = [2.375, 14.625]
[...]

PR #463, opened 2021-09-21
Added simple example with Plots.jl interaction

vini-fda:
As requested by @mkitti on issue #403, I've added a simple plotting example with Plots.jl to the documentation. I've opted to include it in the "convenience contruction" doc file, and created a link to it in the main README.


mkitti:
Looks good. Thank you.

Issue #467, opened 2021-11-21
deduplicate_knots!() doesn't deduplicate?


dodoplus:
check:
```julia
(@v1.6) pkg> st Interpolations
      Status `~/.julia/environments/v1.6/Project.toml`
  [a98d9a8b] Interpolations v0.13.4

julia> xx = [3.6321974345434977, 3.769939196468377, 3.7933987387602532, 3.7933987387602532, 3.7933987387602537, 3.864656907914981];

julia> Interpolations.deduplicate_knots!(xx);

julia> allunique(xx)
false
```

mkitti:
Try applying it twice.

```julia
julia> xx = [3.6321974345434977, 3.769939196468377, 3.7933987387602532, 3.7933987387602532, 3.7933987387602537, 3.864656907914981]
6-element Vector{Float64}:
 3.6321974345434977
 3.769939196468377
 3.7933987387602532
 3.7933987387602532
 3.7933987387602537
 3.864656907914981

julia> Interpolations.deduplicate_knots!(xx)
6-element Vector{Float64}:
 3.6321974345434977
 3.769939196468377
 3.7933987387602532
 3.7933987387602537
 3.7933987387602537
 3.864656907914981

julia> allunique(xx)
false

julia> Interpolations.deduplicate_knots!(xx)
6-element Vector{Float64}:
 3.6321974345434977
 3.769939196468377
 3.7933987387602532
 3.7933987387602537
 3.793398738760254
 3.864656907914981

julia> allunique(xx)
true
```

The issue is that
```julia
julia> xx = [3.6321974345434977, 3.769939196468377, 3.7933987387602532, 3.7933987387602532, 3.7933987387602537, 3.864656907914981]
6-element Vector{Float64}:
 3.6321974345434977
 3.769939196468377
 3.7933987387602532
 3.7933987387602532
 3.7933987387602537
 3.864656907914981

julia> nextfloat(xx[4]) == xx[5]
true
```

dodoplus:
this is still a bug in `deduplicate_knots!()`
It _assumes_ there are no consecutive floats in the knots, but if duplicate knots are possible, there's no reason to assume _almost_ duplicates are just (or more) likely. For example, say I have an exponential function and want to interpolate its _inverse_.

mkitti:
Feel free to propose a fix. As this point, throwing an error is reasonable since it is far from clear what the user expects in this situation. Should we drop a knot?

[...]

Issue #107, opened 2016-02-21
Interpolation over "flat" dimensions


nilshg:
Not sure whether this is an issue or not (there may well be theoretical reasons for why the current state is correct!), but I'm currently interpolating a multidimensional object linearly within a loop, with grids changing in each iteration. In some iterations, the grid can in certain dimensions become "flat", i.e. just a vector containing repeats of the same value. In this case, Interpolations.jl returns NaNs, which at least to me came as a surprise.

When collapsing the grid down to one point (I inserted a line saying `(std(gridvals)>0.1)  || (gridpoints=1)`), an error is raised that dimensions of length 1 are not (yet) supported - is this something that will happen sometime soon? And might it be worthwhile to raise an error if the grid is flat as well, rather than returning NaNs silently, which might be harder to debug?

tlycken:
Hm. That is definietely not the intended behavior. Can you post a minimal example that reproduces the problem?

nilshg:
````
xgrid = collect(linspace(1,10,20))
ygrid = collect(linspace(1,10,20))
func(x,y) = x^2 + y^2
vals = Array(Float64, 20,20)
for x = 1:20, y = 1:20
  vals[x,y] = func(xgrid[x],ygrid[y])
end
using Interpolations
interpolate((xgrid,ygrid),vals, Gridded(Linear()))  # produces expected results
````
but then:
````
ygrid = collect(linspace(1,1,20))
for x = 1:20, y = 1:20
  vals[x,y] = func(xgrid[x],ygrid[y])
end
interpolate((xgrid,ygrid), vals, Gridded(Linear()))  # gives only NaNs
````

This is on Julia 0.4.3 with the latest Interpolations master. Can you reproduce this on your machine?

tlycken:
Yes, I can reproduce.

The simple answer to why this happens, is that since you have 0 distance between your y-points, the interpolation algorithm (which basically does `Δz / Δy` to interpolate in the y-direction) divides by zero, producing a `NaN`, which then poisons the rest of the calculations.

Regarding support for length-1 vectors I refer to @timholy, who implemented the `Gridded` interpolations; I don't know how difficult it would be, but I suppose it might not be too hard, in which case a PR is definitely welcome if you'd like to give it a try. (I can promise to give it a try now that I know it's a feature that someone wants, but I can't give you a timeline - don't hold your breath. And Tim is usually super busy...).

[...]

PR #468, opened 2021-12-08
Safer knot deduplication

mateuszbaran:
This small PR ensures that more pathological cases of duplicated knots (for example `[1.0, 1.0, 1.0, nextfloat(1.0), nextfloat(1.0)]`) are correctly handled.

### DIFF
--- a/src/gridded/gridded.jl
+++ b/src/gridded/gridded.jl
@@ -86,9 +86,11 @@ check_gridded(::Any, ::Tuple{}, ::Tuple{}) = nothing
 degree(flag::Gridded) = flag.degree
 
 """
-    Interpolations.deduplicate_knots!(knots)
+    Interpolations.deduplicate_knots!(knots; move_knots = false)
 
     Makes knots unique by incrementing repeated but otherwise sorted knots using `nextfloat`.
+    If keyword `move_knots` is true, then `nextfloat` will be applied successively until knots
+    are unique. Otherwise, a warning will be issued.
 
     # Example
 
@@ -106,6 +108,14 @@ degree(flag::Gridded) = flag.degree
     0.0
     20.0
     20.000000000000004
+
+    julia> Interpolations.deduplicate_knots!([1.0, 1.0, 1.0, nextfloat(1.0), nextfloat(1.0)]; move_knots = true)
+    5-element Vector{Float64}:
+    1.0
+    1.0000000000000002
+    1.0000000000000004
+    1.0000000000000007
+    1.0000000000000009
     ```
 """
 function deduplicate_knots!(knots; move_knots::Bool = false)
--- a/src/gridded/gridded.jl
+++ b/src/gridded/gridded.jl
@@ -114,7 +114,7 @@ function deduplicate_knots!(knots)
         if i == 1
             continue
         end
-        if knots[i] == last_knot
+        if knots[i] == last_knot || knots[i] <= knots[i-1]
             @inbounds knots[i] = nextfloat(knots[i-1])
         else
             last_knot = @inbounds knots[i]
--- a/src/gridded/gridded.jl
+++ b/src/gridded/gridded.jl
@@ -108,14 +108,18 @@ degree(flag::Gridded) = flag.degree
     20.000000000000004
     ```
 """
-function deduplicate_knots!(knots)
+function deduplicate_knots!(knots; move_knots::Bool = false)
     last_knot = first(knots)
     for i = eachindex(knots)
         if i == 1
             continue
         end
-        if knots[i] == last_knot || knots[i] <= knots[i-1]
+        if knots[i] == last_knot || (move_knots && (@inbounds knots[i] <= knots[i-1]))
             @inbounds knots[i] = nextfloat(knots[i-1])
+        elseif @inbounds knots[i] <= knots[i-1] 
+            @warn "Successive repeated knots detected. Consider using `move_knots` keyword to Interpolations.deduplicate_knots!" last_knot knots[i-1] knots[i]
+            @inbounds knots[i] = nextfloat(knots[i-1])
+            move_knots = true
         else
             last_knot = @inbounds knots[i]
         end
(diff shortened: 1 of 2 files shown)


mateuszbaran:
Hm... Julia 1.7 has a regression in type inference related to `ndims`. I think I can only fix it by forcing more method specialization. What do you think?

mkitti:
I am still not convinced that we are honoring the user's intentions here. In your example we have two unique knots: `1.0` and `nextfloat(1.0)`. The interpolated values at each of the unique knots should be one of the values given for those knots. Let's say the values are `[1,2,3,4,5]`.

```julia
julia> A = [1.0, 1.0, 1.0, nextfloat(1.0), nextfloat(1.0)]
5-element Vector{Float64}:
 1.0
 1.0
 1.0
 1.0000000000000002
 1.0000000000000002

julia> B = deduplicate_knots!(copy(A))
5-element Vector{Float64}:
 1.0
 1.0000000000000002
 1.0000000000000004
 1.0000000000000007
 1.0000000000000009

julia> V = [1,2,3,4,5]
5-element Vector{Int64}:
 1
 2
 3
 4
 5

itp = interpolate((B,), V, Gridded(Linear()))

julia> collect(zip(A, itp.(A)))
5-element Vector{Tuple{Float64, Float64}}:
 (1.0, 1.0)
 (1.0, 1.0)
 (1.0, 1.0)
 (1.0000000000000002, 2.0)
 (1.0000000000000002, 2.0)
```

Above `1.0` is mapped to `1` which is fine. That's the first value I gave to the knot at `1.0`.
`nextfloat(1.0)` however, is mapped to `2.0`. However, the values I assigned to `nextfloat(1.0)` are either `3`, `4`, or `5`.

My expectation is that `itp(nextfloat(1.0)) in V[3:5]` be true, but it is not.

We may thus need a two argument version of `deduplicate_knots!` where the values are also given and some of them are ommitted.

The behavior of `itp` below is closer to what I might expect from deduplicating this.

```julia
julia> V = [1,4,5]
3-element Vector{Int64}:
 1
 4
 5

julia> B = [1.0, nextfloat(1.0), nextfloat(nextfloat(1.0))]
3-element Vector{Float64}:
 1.0
 1.0000000000000002
 1.0000000000000004

julia> itp = interpolate((B,), V, Gridded(Linear()))
3-element interpolate((::Vector{Float64},), ::Vector{Int64}, Gridded(Linear())) with element type Float64:
 1.0
 4.0
 5.0

julia> itp.(A)
3-element Vector{Float64}:
 1.0
 4.0
 5.0

julia> collect(zip(A, itp.(A)))
3-element Vector{Tuple{Float64, Float64}}:
 (1.0, 1.0)
 (1.0000000000000002, 4.0)
 (1.0000000000000004, 5.0)
```

`nextfloat(1.0)` now maps to one of the values originally associated with that knot, `4.0`. If there were further values, then values greater than `nextfloat(1.0)` might map to 5.0. Here we had to not only change the knots, but we also had to change the values and drop some of the knot-value pairs.

mateuszbaran:
Right, my solution is not perfect (though it's fine for my use case). Actually I'm not sure what the best solution would be. Consider knots `[prevfloat(1.0), 1.0, 1.0, nextfloat(1.0)]` and values `[0.0, -1.0, 1.0, 0.0]`. It's impossible to correct it because there is no "right" value to pick for knot 1.0. The sloppy solution is doing what I've done in this PR. If we can remove knots such that the range of values is retained (for example we could drop the second knot if values in my example were `[0.0, 0.5, 1.0, 0.0]`, but not the third one) I think we can reasonably do so. Otherwise the user has to fix the data themselves.

mkitti:
In general, I think the user should probably fix the data themselves. I consider `deduplicate_knots!` to be merely a convenience function. In the latest example, the value at the knot `1.0` should be either `-1.0` or `1.0`. The circumstances force us to drop a knot to be the most consistent.

N5N3:
As for `ndims`, `eltype` and so on, I think we'd better replace the following:
https://github.com/JuliaMath/Interpolations.jl/blob/eefe448a0a1e5e2cbe7d66bb440f841e4bef66b0/src/Interpolations.jl#L150-L161

with:
```julia
itptype(::Type{<:AbstractInterpolation{T,N,IT}}) where {T,N,IT<:DimSpec{InterpolationType}} = IT
itptype(itp::AbstractInterpolation) = itptype(typeof(itp))
ndims(::Type{<:AbstractInterpolation{T,N,<:DimSpec{InterpolationType}}}) where {T,N} = N
ndims(itp::AbstractInterpolation) = ndims(typeof(itp))
eltype(::Type{<:AbstractInterpolation{T,N,<:DimSpec{InterpolationType}}}) where {T,N} = T
eltype(itp::AbstractInterpolation) = eltype(typeof(itp))
```
I delete `gridtype` as it's broken (`GT` is undefined)

mkitti:
@N5N3, I'm unclear how this is related. Could you open a new issue or PR?

mkitti:
To move this PR forward, could we implement this behavior as an keyword option?

mateuszbaran:
This PR is a bit stuck on unrelated Julia 1.7 failures so I was planning to go back to it once that is solved in some way. I'm also not quite sure at the moment how exactly this should work.

mkitti:
Let me worry about 1.7.

mkitti:
I ended up going with a warning:
```julia
julia> Interpolations.deduplicate_knots!([1.0, 1.0, 1.0, nextfloat(1.0), nextfloat(1.0)])
┌ Warning: Successive repeated knots detected. Consider using `move_knots` keyword to Interpolations.deduplicate_knots!
│   last_knot = 1.0
│   knots[i - 1] = 1.0000000000000004
│   knots[i] = 1.0000000000000002
└ @ Interpolations C:\Users\kittisopikulm\.julia\dev\Interpolations\src\gridded\gridded.jl:120
5-element Vector{Float64}:
 1.0
 1.0000000000000002
 1.0000000000000004
 1.0000000000000007
 1.0000000000000009
```

mkitti:
Fix #467

mateuszbaran:
Thanks for finishing it! I was still considering adding the node removal feature here but I was busy with other work.

jaemolihm:
@mkitti Does this PR fix #107 ?

PR #470, opened 2021-12-13
Add support for WENO4 interpolation

tiagopereira:
This PR is not currently working. I would like to add support for interpolation with a 4th order Weighted Essentially Non-Oscillatory (WENO) scheme, following [Janett et al (2019)](https://ui.adsabs.harvard.edu/abs/2019A%26A...624A.104J/abstract). I wrote the standalone package [WENO4.jl](https://github.com/tiagopereira/WENO4.jl) (not yet registered) that has been tested to work as it should. This PR is an attempt at bringing the same functionality into Interpolations.jl.

My experience with Julia is limited, and I don't fully understand the abstractions and architecture of Interpolations.jl, so I apologise for the many mistakes in the current PR and would appreciate some help at getting this code working and sufficiently general. This implementation follows closely what I saw in `monotonic.jl`, but I couldn't get my head around all the abstractions. This is the type of usage I was hoping it would get when working:

```julia
using Interpolations

itp = interpolate(x, y, interpolate(x, y, WENO4Interpolation())
new_values = itp.(new_x)
```

The idea is to follow existing practice of getting a way to generate an interpolant for all input data values and nodes, and then a separate function to evaluate this interpolant at a required point. I have tested and the functions inside, in particular `interpolate` (to build the interpolant) and the function to evaluate the interpolant, currently 

```julia
function (itp::WENO4Interpolation)(x::Number)
```

works if I rewrite it as the standalone function

```julia
function evaluate(itp::WENO4Interpolation, x::Number)
```
However, it is not working in the present formulation because I don't really understand the relation between the `interpolate` function and the function with the same name as the `struct`, here called  `WENO4Interpolation`.

Currently the functions used perform about twice as slow as in the standalone package, which I attribute to the need of generating intermediate arrays. This could also be improved.

### DIFF
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -35,6 +35,7 @@ export
     # extrapolation/extrapolation.jl
     # monotonic/monotonic.jl
     # scaling/scaling.jl
+    # weno/weno.jl
 
 using LinearAlgebra, SparseArrays
 using StaticArrays, WoodburyMatrices, Ratios, AxisAlgorithms, OffsetArrays
@@ -493,6 +494,7 @@ include("lanczos/lanczos.jl")
 include("lanczos/lanczos_opencv.jl")
 include("iterate.jl")
 include("chainrules/chainrules.jl")
+include("weno/weno.jl")
 
 function __init__()
     @require Unitful="1986cc42-f94f-5a68-af5c-568840ba703d" include("requires/unitful.jl")
--- a/src/weno/weno.jl
+++ b/src/weno/weno.jl
@@ -1,0 +1,200 @@
+#=
+    Interpolate using weighted essentially non-oscillatory (WENO) techniques.
+=#
+
+export WENO4Interpolation
+
+
+"""
+    WENOInterpolationType
+
+Abstract class for all types of WENO interpolation.
+"""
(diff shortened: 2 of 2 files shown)


mkitti:
`interpolate` is a general method that will eventually construct an instance of the interpolation object. `(itp::WENO4Interpolation)(x...)` is used when you try to use an _instance_ of the struct as a function.

```julia
julia> struct Greeting 
           salutation::String
       end # define a struct

julia> hello = Greeting("Hello") # construct a Greeting instance with salutation == "Hello"
Greeting("Hello")

julia> (g::Greeting)(name) = println(g.salutation * ", " * name) # using an instance

julia> hello("Tiago") # Use the instance as a function (e.g. a Functor)
Hello, Tiago

julia> hola = Greeting("Hola") # construct another Greeting instance with salutation == "Hola"
Greeting("Hola")

julia> hola("Marcos")
Hola, Marcos

julia> Greeting() = Greeting("Hi") # define a constructor for Greeting that takes no arguments... and defaults salutation to "Hi"
Greeting

julia> default = Greeting() # use the new constructor
Greeting("Hi")

julia> default("Mark")
Hi, Mark
```

PR #471, opened 2021-12-14
Fix inference regression on 1.7

N5N3:
Since `AbstractInterpolation` <: `AbstractArray`, there's no need to define `ndims` and `eltype` ourself.
`gridtype` is broken, (and unused), so I remove it.
`itptype` is modified to avoid the recursive call of `supertype`.

### DIFF
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -147,18 +147,13 @@ julia> bounds(itp)
 bounds(itp::AbstractInterpolation) = map(twotuple, lbounds(itp), ubounds(itp))
 bounds(itp::AbstractInterpolation, d) = bounds(itp)[d]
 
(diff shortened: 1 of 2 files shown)


mkitti:
Also tracking https://github.com/JuliaLang/julia/issues/43368

N5N3:
This PR should be enough as a temporary fix.
Do we need test for `itptype`? looks like it is called only once at:https://github.com/JuliaMath/Interpolations.jl/blob/0f4dd95bda2e8cf60fd69245990921bcd6d66d94/src/b-splines/b-splines.jl#L157

mkitti:
Can you write a test for the one line you added?

N5N3:
test added

mkitti:
Merging

PR #472, opened 2021-12-14
Inference improvement.

N5N3:
Fix #469. test added. (I only test it on master)

PR #480, opened 2022-02-23
Fix inference on `symmatrix` (patch for #472)

N5N3:
This PR hardcoded the output size of `symmatrix` for L < 32 to avoid `sqrt`/`power`, thus also work on 1.6 and 1.7.
For `L > 32`, we'll touch the inference threshold of `Tuple` anyway. Thus more methods seems meaningless.
(I'm not sure why #472 passed previously on 1.7.0. Quite strange.)


src/b-splines/indexing.jl
```diff
@@ -221,3 +219,17 @@ function symmatrix(h::NTuple{L,Any}) where L
         SMatrix{N,N}(h[i] for i in l)
     end
 end
+
+# Hardcode the output size.
+for N in 4:7
+    L = N*(N+1)÷2
+    @eval symsize(::Val{$L}) = $N
```
mkitti, line 226: Why not use [`@generated`](https://docs.julialang.org/en/v1/manual/metaprogramming/#Generated-functions) here rather than `@eval`?

N5N3: Do you mean:
```julia
@generated function f(::Val{L}) where {L}
           N = floor(Int, sqrt(2L))
           N*(N+1) == 2L || error()
           return :($N)
       end
```
?

mkitti: That looks good if it works.

PR #490, opened 2022-05-11
Fix scaled interpolation for Unitful ranges.

contradict:


mkitti:
It looks like we are intentionally removing units from `coordlookup`, is that right?

PR #504, opened 2022-07-27
Add native GPU support.

N5N3:
~Only 1d and 2d `BSpline` Interpolations are tested though.~
Test has been added for `Gridded` and `Lanczos`.
~Since this overload needs `GPUArraysCore.jl`, we need to drop < 1.6 compatibility if we don't want to add more `@require`.~
`Base.getindex` is overloaded with a `InterpGetIndex` wrapper in the latest design. Thus there's no need to depend on `GPUArraysCore.jl`. (We can keep < 1.6 compatibility now)

### DIFF
--- a/docs/src/devdocs.md
+++ b/docs/src/devdocs.md
@@ -55,6 +55,8 @@ concrete subtypes:
 General `AbstractArray`s may be indexed with `WeightedIndex` indices,
 and the result produces the interpolated value. In other words, the end result
 is just `itp.coefs[wis...]`, where `wis` is a tuple of `WeightedIndex` indices.
+To make sure this overloading is effective, we wrap the `coefs` with `InterpGetindex`，
+i.e. `InterpGetindex(itp.coefs)[wis...]`
 
 Derivatives along a particular axis can be computed just by substituting a component of `wis` for one that has been designed to compute derivatives rather than values.
 
@@ -107,7 +109,7 @@ Interpolations.WeightedAdjIndex{2, Float64}(1, (0.6000000000000001, 0.3999999999
 julia> wis[3]
 Interpolations.WeightedAdjIndex{2, Float64}(1, (0.30000000000000004, 0.7))
 
-julia> A[wis...]
+julia> Interpolations.InterpGetindex(A)[wis...]
 8.7
 ```
 
@@ -125,7 +127,7 @@ This computed the value of `itp` at `x...` because we called `weightedindexes` w
 
 !!! note
     Remember that prefiltering is not used for `Linear` interpolation.
-    In a case where prefiltering is used, we would substitute `itp.coefs[wis...]` for `A[wis...]` above.
+    In a case where prefiltering is used, we would substitute `InterpGetindex(itp.coefs)[wis...]` for `InterpGetindex(A)[wis...]` above.
 
 To compute derivatives, we *also* pass additional functions like [`Interpolations.gradient_weights`](@ref):
 
@@ -142,13 +144,13 @@ julia> wis[2]
 julia> wis[3]
 (Interpolations.WeightedAdjIndex{2,Float64}(1, (0.8, 0.19999999999999996)), Interpolations.WeightedAdjIndex{2,Float64}(1, (0.6000000000000001, 0.3999999999999999)), Interpolations.WeightedAdjIndex{2,Float64}(1, (-1.0, 1.0)))
 
-julia> A[wis[1]...]
+julia> Interpolations.InterpGetindex(A)[wis[1]...]
 1.0
 
-julia> A[wis[2]...]
+julia> Interpolations.InterpGetindex(A)[wis[2]...]
 3.000000000000001
 
-julia> A[wis[3]...]
+julia> Interpolations.InterpGetindex(A)[wis[3]...]
 9.0
 ```
 In this case you can see that `wis` is a 3-tuple-of-3-tuples. `A[wis[i]...]` can be used to compute the `i`th component of the gradient.
@@ -168,3 +170,29 @@ The code to do this replacement is a bit complicated due to the need to support 
 It makes good use of *tuple* manipulations, sometimes called "lispy tuple programming."
 You can search Julia's discourse forum for more tips about how to program this way.
 It could alternatively be done using generated functions, but this would increase compile time considerably and can lead to world-age problems.
+
+### GPU Support
+At present, `Interpolations.jl` supports interpolant usage on GPU via broadcasting.
+
+A basic work flow looks like:
+```julia
+julia> using Interpolations, Adapt, CUDA # Or any other GPU package
+
+julia > itp = Interpolations.interpolate([1, 2, 3], (BSpline(Linear()))); # construct the interpolant object on CPU
+
+julia> cuitp = adapt(CuArray{Float32}, itp); # adapt it to GPU memory
+
+julia > cuitp.(1:0.5:2) # call interpolant object via broadcast
+
+julia> gradient.(Ref(cuitp), 1:0.5:2)
+```
+
+To achieve this, an `ITP <: AbstractInterpolation` should define it's own `Adapt.adapt_structure(to, itp::ITP)`, which constructs a new `ITP` with the adapted
+fields (`adapt(to, itp.fieldname)`) of `itp`. The field adaption could be skipped
+if we know that it has been GPU-compatable, e.g. a `isbit` range.
+
+!!! note
+    Some adaptors may change the storage type. Please ensure that the adapted `itp`
+    has the correct element type via the method `eltype`.
+
+Also, all GPU-compatable `AbstractInterpolation`s should define their own `Interpolations.root_storage_type`. This function allows us to modify the broadcast mechanism by overloading the default `BroadcastStyle`. See [Customizing broadcasting](https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting) for more details.
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -269,25 +269,15 @@ Base.:(/)(wi::WeightedArbIndex, x::Number) = WeightedArbIndex(wi.indexes, wi.wei
 
 ### Indexing with WeightedIndex
 
+# We inject `WeightedIndex` as a non-exported indexing point with a `InterpGetindex` wrapper.
+# `InterpGetindex` is not a subtype of `AbstractArray`. This ensures that the overload applies to all array types.
 struct InterpGetindex{N,A<:AbstractArray{<:Any,N}}
     coeffs::A
     InterpGetindex(itp::AbstractInterpolation) = InterpGetindex(coefficients(itp))
     InterpGetindex(A::AbstractArray) = new{ndims(A),typeof(A)}(A)
 end
-Base.@propagate_inbounds Base.getindex(A::InterpGetindex{N}, I::Vararg{Union{Int,WeightedIndex},N}) where {N} =
+@inline Base.getindex(A::InterpGetindex{N}, I::Vararg{Union{Int,WeightedIndex},N}) where {N} =
     interp_getindex(A.coeffs, I, ntuple(d->0, Val(N))...)
-
-# TODO: should we drop the following injection?
-# We inject indexing with `WeightedIndex` at a non-exported point in the dispatch heirarchy.
-# This is to avoid ambiguities with methods that specialize on the array type rather than
-# the index type.
-Base.to_indices(A, I::Tuple{Vararg{Union{Int,WeightedIndex}}}) = I
-if VERSION < v"1.6.0-DEV.104"
-    @propagate_inbounds Base._getindex(::IndexLinear, A::AbstractVector, i::Int) = getindex(A, i)  # ambiguity resolution
-end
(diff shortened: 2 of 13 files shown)


mkitti:
I'm a bit wary of dependency creep. Is there a way to do this via a subpackage?

johnnychen94:
The additional dependencies are quite small (loading time from 0.74s to 0.79s in my machine). But they request bumping Julia compat to 1.6, which is okay as well (I think).

Either living in here or in the new interpolationsGPU would be great!

cc: @timholy @ChantalJuntao (from https://github.com/JuliaImages/ImageTransformations.jl/pull/156)

src/b-splines/b-splines.jl
```diff
@@ -66,7 +66,7 @@ However, for customized control you may also construct them with
 where `T` gets computed from the product of `TWeights` and `eltype(coefs)`.
 (This is equivalent to indicating that you'll be evaluating at locations `itp(x::TWeights, y::TWeights, ...)`.)
 """
-struct BSplineInterpolation{T,N,TCoefs<:AbstractArray,IT<:DimSpec{BSpline},Axs<:Tuple{Vararg{AbstractUnitRange,N}}} <: AbstractInterpolation{T,N,IT}
+struct BSplineInterpolation{T,N,TCoefs<:AbstractArray,IT<:DimSpec{BSpline},Axs<:Indices{N}} <: AbstractInterpolation{T,N,IT}
```
mkitti, line 69: Nothing has actually changed here, right?

src/b-splines/b-splines.jl
```diff
@@ -98,7 +101,7 @@ function BSplineInterpolation(::Type{TWeights}, A::AbstractArray{Tel,N}, it::IT,
     else
         T = typeof(zero(TWeights) * first(A))
     end
-    BSplineInterpolation{T,N,typeof(A),IT,typeof(axs)}(A, fix_axis.(axs), it)
+    BSplineInterpolation{T,N}(A, fix_axis.(axs), it)
```
mkitti, line 104: Could you explain this?

src/scaling/scaling.jl
```diff
@@ -2,7 +2,7 @@ export ScaledInterpolation, eachvalue
 
 import Base: iterate
 
-struct ScaledInterpolation{T,N,ITPT,IT,RT} <: AbstractInterpolationWrapper{T,N,ITPT,IT}
+struct ScaledInterpolation{T,N,ITPT,IT,RT<:NTuple{N,AbstractRange}} <: AbstractInterpolationWrapper{T,N,ITPT,IT}
```
mkitti, line 5: Is this necessary?

mkitti (commented):
Just a few questions ...

src/scaling/scaling.jl
N5N3: The current added code assumes that ranges here is GPU compatible.
I think we’d better add this restriction to constructor.

src/b-splines/b-splines.jl
N5N3: fix_axis below might change the input type.
The current version would convert ‘fix_axis.(axs)’ back to ‘typeof(axs)’.

src/b-splines/b-splines.jl
N5N3: Yes, just fuse some cleaner type alias.

src/scaling/scaling.jl
mkitti: The only problem is that this is a potentially breaking change, although I'm not sure if I have a concrete example yet.

A non-breaking change would be to create this type constant and use that.
```julia
const RangedScaledInterpolation{T, N, ITPT, IT, RT <: NTuple{N,AbstractRange}} = ScaledInterpolation{T,N,ITPT,IT,RT}
```

N5N3: In fact, all methods belows assumes that `ranges` all have `step` defined, so I think this change should be safe.
But of course we can discuss this in another thread. I'll move this restriction to `adapt` side.

src/extrapolation/extrapolation.jl
```diff
@@ -71,8 +71,8 @@ end
         eflag = tcollect(etpflag, etp)
         xs = inbounds_position(eflag, bounds(itp), x, etp, x)
         g = @inbounds gradient(itp, xs...)
-        skipni = t->skip_flagged_nointerp(itp, t)
-        SVector(extrapolate_gradient.(skipni(eflag), skipni(x), skipni(xs), Tuple(g)))
+        skipni = Base.Fix1(skip_flagged_nointerp, itp)
+        SVector(map(extrapolate_gradient, skipni(eflag), skipni(x), skipni(xs), Tuple(g)))
```
mkitti, line 75: Does the map versus broadcast make a difference here?

mkitti (approved):
This looks pretty good to me. Could you add some developer documentation about how to use Adapt when someone adds a new interpolation tyoe?

src/extrapolation/extrapolation.jl
N5N3: IIUC, `Tuple(g)` itself should have skipped the "nointerp" dimensions, thus `skipni(eflag)`, `skipni(x)`, `skipni(xs)`, `Tuple(g)` should share the same length.
So this change will give the same result.
The only difference is that L75 is still inlinable with `extrapolate_gradient.()` even if `ndims(itp) > 32`, (as it's `ntuple`-based).
But i think there's no practical usage for such a high-dimensional problem, so `map` is ok here.

mkitti:
In particular, can you please update this document:
https://github.com/JuliaMath/Interpolations.jl/blob/master/docs/src/devdocs.md

N5N3:
@mkitti devdoc has been updated (including the new `InterpGetindex` wrapper).
The `Adapt` based transformation itself should be familiar to every GPU end users.
So I just emphasize that 'eltype' may change during adaption.

docs/src/devdocs.md
```diff
@@ -55,6 +55,8 @@ concrete subtypes:
 General `AbstractArray`s may be indexed with `WeightedIndex` indices,
 and the result produces the interpolated value. In other words, the end result
 is just `itp.coefs[wis...]`, where `wis` is a tuple of `WeightedIndex` indices.
+To make sure this overloading is effective, we wrapper the `coefs` with `InterpGetindex`，
```
mkitti, line 58:
```suggestion
To make sure this overloading is effective, we wrap the `coefs` with `InterpGetindex`，
```

docs/src/devdocs.md
```diff
@@ -168,3 +170,29 @@ The code to do this replacement is a bit complicated due to the need to support
...
+
+To achieve this, an `ITP <: AbstractInterpolation` should define it's own `Adapt.adapt_structure(to, itp::ITP)`, which constructs a new `ITP` with the adapted
+fields (`adapt(to, itp.fieldname)`) of `itp`. The field adaption could be skipped
+if we know that it has been GPU-compatable, e.g. a bit range.
+
+!!! note
+    Some adaptor might change the storage type. Please ensure that the adapted `itp`
+    has the corrent `eltype`.
+
+Also, all GPU-compatable `AbstractInterpolation`s should define their own `Interpolations.root_storage_type`. This function help us injecting into the broadcast mechanism by overloading the default `BroadcastStyle`. See [Customizing broadcasting](https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting) for more details.
```
mkitti, line 198:
```suggestion
Also, all GPU-compatable `AbstractInterpolation`s should define their own `Interpolations.root_storage_type`. This function allows us to modify the broadcast mechanism by overloading the default `BroadcastStyle`. See [Customizing broadcasting](https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting) for more details.
```

docs/src/devdocs.md
```diff
@@ -168,3 +170,29 @@ The code to do this replacement is a bit complicated due to the need to support
 It makes good use of *tuple* manipulations, sometimes called "lispy tuple programming."
 You can search Julia's discourse forum for more tips about how to program this way.
 It could alternatively be done using generated functions, but this would increase compile time considerably and can lead to world-age problems.
+
+### GPU Support
+At present, `Interpolations.jl` supports interpolant usage on GPU via broadcast mechanism.
```
mkitti, line 175:
```suggestion
At present, `Interpolations.jl` supports interpolant usage on GPU via broadcasting.
```

docs/src/devdocs.md
```diff
@@ -168,3 +170,29 @@ The code to do this replacement is a bit complicated due to the need to support
...
+julia> cuitp = adapt(CuArray{Float32}, itp); # adapt it to GPU memory
+
+julia > cuitp.(1:0.5:2) # call interpolant object via broadcast
+
+julia> gradient.(Ref(cuitp), 1:0.5:2)
+```
+
+To achieve this, an `ITP <: AbstractInterpolation` should define it's own `Adapt.adapt_structure(to, itp::ITP)`, which constructs a new `ITP` with the adapted
+fields (`adapt(to, itp.fieldname)`) of `itp`. The field adaption could be skipped
+if we know that it has been GPU-compatable, e.g. a bit range.
```
mkitti, line 192:
```suggestion
To achieve this, an `ITP <: AbstractInterpolation` should define its own `Adapt.adapt_structure(to, itp::ITP)`.
`adapt_structure` constructs a new `ITP` with the adapted fields of `itp` as defined by `adapt(to, itp.fieldname)`. The field adaption could be skipped if we know that it is been GPU-compatable, e.g. a bit range.
```

Also what do you mean by "bit range" here. Do you mean bit type where `isbitstype` is true?

docs/src/devdocs.md
```diff
@@ -168,3 +170,29 @@ The code to do this replacement is a bit complicated due to the need to support
...
+
+!!! note
+    Some adaptor might change the storage type. Please ensure that the adapted `itp`
+    has the corrent `eltype`.
```
mkitti, line 196:
```suggestion
    Some adaptors may change the storage type. Please ensure that the adapted `itp`
    has the correct element type via the method `eltype`.
```

src/Interpolations.jl
```diff
@@ -269,6 +269,15 @@ Base.:(/)(wi::WeightedArbIndex, x::Number) = WeightedArbIndex(wi.indexes, wi.wei
...
 
+struct InterpGetindex{N,A<:AbstractArray{<:Any,N}}
+    coeffs::A
+    InterpGetindex(itp::AbstractInterpolation) = InterpGetindex(coefficients(itp))
+    InterpGetindex(A::AbstractArray) = new{ndims(A),typeof(A)}(A)
+end
+Base.@propagate_inbounds Base.getindex(A::InterpGetindex{N}, I::Vararg{Union{Int,WeightedIndex},N}) where {N} =
+    interp_getindex(A.coeffs, I, ntuple(d->0, Val(N))...)
+
+# TODO: should we drop the following injection?
```
mkitti, line 280: Can you elaborate? Is this fully obsolete now?

src/gpu_support.jl
```diff
@@ -0,0 +1,78 @@
...
+    coefs′ = adapt(to, itp.coefs)
+    T′ = update_eltype(T, coefs′, itp.coefs)
+    BSplineInterpolation{T′,N}(coefs′, itp.parentaxes, itp.it)
+end
+
+function update_eltype(T, coefs′, coefs)
+    ET = eltype(coefs′)
+    ET === eltype(coefs) && return T
+    WT = tweight(coefs′)
+    if isempty(coefs)
```
mkitti, line 14: Expand tests to check when `isempty(coefs)`

src/gpu_support.jl
```diff
@@ -0,0 +1,78 @@
...
+        T′ = Base.promote_op(*, WT, ET)
+    else
+        T′ = Base.promote_op(*, WT, ET)
+        if !isconcretetype(T′)
+            T′ = typeof(zero(WT) * convert(ET, first(coefs)))
```
mkitti, line 19: Expand tests to check when `isconcretetype(T')` is `false`.

mkitti (changes requested):
I did some editing on the documentation.

Also I notice there are some cases that are not covered by the tests. Could you expand test coverage?

docs/src/devdocs.md
N5N3: Yes, has been replaced with "a `isbit` range"

src/Interpolations.jl
N5N3: `Base._getindex` is an internal api anyway. Since `Interpolations.jl` itself no-longer use it. It makes sense to me to dop it.

mkitti:
Is this ready to merge?

N5N3:
Yes, CI passed. The test also passed locally with 1.6.7.

ChantalJuntao:
> ~Only 1d and 2d `BSpline` Interpolations are tested though.~ Test has been added for `Gridded` and `Lanczos`. ~Since this overload needs `GPUArraysCore.jl`, we need to drop < 1.6 compatibility if we don't want to add more `@require`.~ `Base.getindex` is overloaded with a `InterpGetIndex` wrapper in the latest design. Thus there's no need to depend on `GPUArraysCore.jl`. (We can keep < 1.6 compatibility now)

This is kind of late, but was it OK to merge this without tests for 3D? And I thought that Cl's tests don't cover GPU code?

mkitti:
Feel free to submit another pull request. As far as I know there are no GPUs to run CI on here.

PR #512, opened 2022-08-15
Add Previous and Next for Constant interpolation 

### LINKED ISSUE #511: Manual misses Previous and Next for Constant interpolation
It seems the features from #338 are not mentioned in the manual.

judober:
Fix #511 

I thought of adding `Constant{Mode}()` to the docstring of `interpolate` but was not shure if this is apropriate. Thereby I noted that these docstring have different entries which seem inconsistent/incomplete. E.g. once there is `BSpline(NoInterp())` and once only `NoInterp()`. Also these do not reflect the posibility of using `Gridded`.

### DIFF
--- a/docs/src/control.md
+++ b/docs/src/control.md
@@ -119,7 +119,7 @@ NoInterp()
 whereby the coordinate of the selected input vector MUST be located on a grid point. Requests for off grid
 coordinates results in the throwing of an error.
 
-For [`Constant`](@ref) there are the additional parameters. Use `Constant{Previous}()` in order to perform a previous
+For [`Constant`](@ref) there are additional parameters. Use `Constant{Previous}()` in order to perform a previous
 neighbor interpolation. Use `Constant{Next}()` for a next neighbor interpolation.
 Note that rounding can be an issue, see [#473](https://github.com/JuliaMath/Interpolations.jl/issues/473).
 
--- a/docs/src/control.md
+++ b/docs/src/control.md
@@ -119,9 +119,9 @@ NoInterp()
 whereby the coordinate of the selected input vector MUST be located on a grid point. Requests for off grid
 coordinates results in the throwing of an error.
 
-For `Constant()` there are the additional options `Constant{Previous}()` in order to perform a previous
+For [`Constant`](@ref) there are the additional options `Constant{Previous}()` in order to perform a previous
 neighbor interpolation and `Constant{Next}()` for a next neighbor interpolation.
-Note that rounding can be an issue, see [#473](https://github.com/JuliaMath/Interpolations.jl/issues/473)
+Note that rounding can be an issue, see [#473](https://github.com/JuliaMath/Interpolations.jl/issues/473).
 
 `missing` data will naturally propagate through the interpolation,
 where some values will become missing. To avoid that, one can
--- a/docs/src/control.md
+++ b/docs/src/control.md
@@ -119,6 +119,10 @@ NoInterp()
 whereby the coordinate of the selected input vector MUST be located on a grid point. Requests for off grid
 coordinates results in the throwing of an error.
 
+For `Constant()` there are the additional options `Constant{Previous}()` in order to perform a previous
+neighbor interpolation and `Constant{Next}()` for a next neighbor interpolation.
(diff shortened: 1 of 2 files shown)


mkitti:
I think expanding the doc string for `Constant` is sufficient, but I will think about discoverability more later.

mkitti:
Documentation preview: https://juliamath.github.io/Interpolations.jl/previews/PR512

mkitti:
I'm not sure why the preview is not working. I'll investigate later.

src/b-splines/constant.jl
```diff
@@ -1,8 +1,32 @@
...
+Default option for `Constant` that performes *nearest-neighbor* interpolations.
+Can optionally be specified as
+```
+Constant() === Constant{Nearest}()
+```
+"""
 struct Nearest <: ConstantInterpType end
+
+"""
+Option for `Constant` that performes *previous-neighbor* interpolations.
```
mkitti, line 15:
```suggestion
Parameter for `Constant` that performs *previous-neighbor* interpolations.
```

src/b-splines/constant.jl
```diff
@@ -1,8 +1,32 @@
...
+Applied through 
+´´´
+Constant{Previous}()
+´´´
+"""
 struct Previous <: ConstantInterpType end
+
+"""
+Option for `Constant` that performes *next-neighbor* interpolations.
```
mkitti, line 24:
```suggestion
Parameter for `Constant` that performs *next-neighbor* interpolations.
```

src/b-splines/constant.jl
```diff
@@ -31,6 +55,11 @@ end
 """
 Constant b-splines are *nearest-neighbor* interpolations, and effectively
 return `A[round(Int,x)]` when interpolating.
+
+`Constant{Previous}` interpolates to the previous value thus equivalent
+to `A[floor(Int,x)]`.
+`Constant{Next}` interpolates to the next value thus equivalent
+to `A[ceil(Int,x)]`.
```
mkitti, line 62:
```suggestion
Constant b-splines are *nearest-neighbor* interpolations, and effectively
return `A[round(Int,x)]` when interpolating without scaling.

`Constant{Previous}` interpolates to the previous value and is thus equivalent
to `A[floor(Int,x)]` without scaling.
`Constant{Next}` interpolates to the next value and is thus equivalent
to `A[ceil(Int,x)]` without scaling.
```

docs/src/control.md
```diff
@@ -111,6 +119,10 @@ NoInterp()
 whereby the coordinate of the selected input vector MUST be located on a grid point. Requests for off grid
 coordinates results in the throwing of an error.
 
+For [`Constant`](@ref) there are the additional options `Constant{Previous}()` in order to perform a previous
+neighbor interpolation and `Constant{Next}()` for a next neighbor interpolation.
+Note that rounding can be an issue, see [#473](https://github.com/JuliaMath/Interpolations.jl/issues/473).
+
```
mkitti, line 125:
```suggestion
For [`Constant`](@ref) there are the additional parameters. Use `Constant{Previous}()` in order to perform a previous
neighbor interpolation. Use `Constant{Next}()` for a next neighbor interpolation.
Note that rounding can be an issue, see [#473](https://github.com/JuliaMath/Interpolations.jl/issues/473).

```

src/b-splines/constant.jl
```diff
@@ -1,8 +1,32 @@
 export Nearest, Previous, Next
 
 abstract type ConstantInterpType end
+
+"""
+Default option for `Constant` that performes *nearest-neighbor* interpolations.
```
mkitti, line 6:
```suggestion
Default parameter for `Constant` that performs *nearest-neighbor* interpolation.
```

mkitti (changes requested):
Just a few edits. Thanks.

docs/src/control.md
judober: I removed _the_ additional ...

src/b-splines/constant.jl
judober: I added a remark for the inaccuracies when scaling (as far as I understand them).

judober:
Anything left to do here?

mkitti:
Looks good to me

PR #513, opened 2022-09-01
Create Invalidations.yml

ranocha:
This is based on https://github.com/julia-actions/julia-invalidations. Adding such checks came up in https://discourse.julialang.org/t/potential-performance-regressions-in-julia-1-8-for-special-un-precompiled-type-dispatches-and-how-to-fix-them/86359. I suggest to add this check here since this package is widely used as a dependency.

See also SciML/MuladdMacro.jl#26 and SciML/MuladdMacro.jl#29

### DIFF
--- a/.github/workflows/Invalidations.yml
+++ b/.github/workflows/Invalidations.yml
@@ -1,0 +1,40 @@
+name: Invalidations
+
+on:
+  pull_request:
+
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: always.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  evaluate:
+    # Only run on PRs to the default branch.
+    # In the PR trigger above branches can be specified only explicitly whereas this check should work for master, main, or any other default branch
+    if: github.base_ref == github.event.repository.default_branch
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


mkitti:
I'll merge it. How does this work?

ranocha:
Here is an example when this test passes: https://github.com/SciML/MuladdMacro.jl/runs/8121563802?check_suite_focus=true

And here is a failing example: https://github.com/ranocha/InvalidationsTest.jl/runs/8134758303?check_suite_focus=true

Other than that, it will just add a new CI run (that is relatively cheap). It will fail when a PR introduces new invalidations (and also report the numbers of invalidations for each PR). From there on, you can decide whether you want to ignore it or chase down the invalidations & fix them.

Do you have any specific questions about this?

ranocha:
For example, this package and its dependencies currently cause 500 invalidations: https://github.com/JuliaMath/Interpolations.jl/runs/8136468587?check_suite_focus=true#step:9:1

mkitti:
I guess I'll have to run SnoopCompiler myself to figure which 500?

ranocha:
Yes, exactly. Fixing invalidations is quite a manual process in my experience anyway, so you need to run things locally to figure out what's going on and how to fix things. The docs of SnoopCompile.jl are a good place to learn about this: https://timholy.github.io/SnoopCompile.jl/stable/snoopr/

PR #522, opened 2022-10-03
Adding AkimaMonotonicInterpolation

wg030:
I added the new interpolation type `AkimaMonotonicInterpolation` and tested the implementation by comparing the results with the results of working code for Akima splines.
I also ran the tests for `@testset "Monotonic"` successfully and checked the plots created by _test/plots/monotonic/sanity.jl_

### DIFF
--- a/src/monotonic/monotonic.jl
+++ b/src/monotonic/monotonic.jl
@@ -306,23 +306,20 @@ function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
     n = length(x)
     Δ = Vector{TCoeffs}(undef, n-1)
     m = Vector{TCoeffs}(undef, n)
-
     for k in 1:n-1
         Δk = (y[k+1] - y[k]) / (x[k+1] - x[k])
         Δ[k] = Δk
     end
     Γ = [3Δ[1] - 2Δ[2]; 2Δ[1] - Δ[2]; Δ; 2Δ[n-1] - Δ[n-2]; 3Δ[n-1] - 2Δ[n-2]]
-
     for k in 1:n
         δ = abs(Γ[k+3] - Γ[k+2]) + abs(Γ[k+1] - Γ[k])
-        if δ > 0
+        if δ > zero(δ)
             α = abs(Γ[k+1] - Γ[k]) / δ
             m[k] = (1-α) * Γ[k+1] + α * Γ[k+2]
         else
             m[k] = 0.5 * Γ[k+1] + 0.5 * Γ[k+2]
         end
     end
-
     return (m, Δ)
 end
 
--- a/src/monotonic/monotonic.jl
+++ b/src/monotonic/monotonic.jl
@@ -178,7 +178,7 @@ function interpolate(::Type{TWeights}, ::Type{TCoeffs1},::Type{TCoeffs2},::Type{
     m, Δ = calcTangents(TCoeffs1, knots, A, it)
     c = Vector{TCoeffs2}(undef, n-1)
     d = Vector{TCoeffs3}(undef, n-1)
-    for k ∈ 1:n-1
+    for k ∈ eachindex(c)
         if TInterpolationType == LinearMonotonicInterpolation
             c[k] = zero(TCoeffs2)
             d[k] = zero(TCoeffs3)
@@ -249,10 +249,9 @@ function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
     n = length(x)
     Δ = Vector{TCoeffs}(undef, n-1)
     m = Vector{TCoeffs}(undef, n)
-    for k in 1:n-1
-        Δk = (y[k+1] - y[k]) / (x[k+1] - x[k])
-        Δ[k] = Δk
-        m[k] = Δk
+    for k ∈ eachindex(Δ)
+        Δ[k] = (y[k+1] - y[k]) / (x[k+1] - x[k])
+        m[k] = Δ[k]
     end
     m[n] = Δ[n-1]
     return (m, Δ)
@@ -264,13 +263,12 @@ function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
     n = length(x)
     Δ = Vector{TCoeffs}(undef, n-1)
     m = Vector{TCoeffs}(undef, n)
-    for k in 1:n-1
-        Δk = (y[k+1] - y[k]) / (x[k+1] - x[k])
-        Δ[k] = Δk
+    for k ∈ eachindex(Δ)
+        Δ[k] = (y[k+1] - y[k]) / (x[k+1] - x[k])
         if k == 1   # left endpoint
-            m[k] = Δk
+            m[k] = Δ[k]
         else
-            m[k] = (Δ[k-1] + Δk) / 2
+            m[k] = (Δ[k-1] + Δ[k]) / 2
         end
     end
     m[n] = Δ[n-1]
@@ -283,11 +281,10 @@ function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
     n = length(x)
     Δ = Vector{TCoeffs}(undef, n-1)
     m = Vector{TCoeffs}(undef, n)
-    for k in 1:n-1
-        Δk = (y[k+1] - y[k]) / (x[k+1] - x[k])
-        Δ[k] = Δk
+    for k ∈ eachindex(Δ)
+        Δ[k] = (y[k+1] - y[k]) / (x[k+1] - x[k])
         if k == 1   # left endpoint
-            m[k] = Δk
+            m[k] = Δ[k]
         else
             m[k] = (oneunit(TTension) - method.tension) * (y[k+1] - y[k-1]) / (x[k+1] - x[k-1])
         end
@@ -306,12 +303,11 @@ function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
     n = length(x)
     Δ = Vector{TCoeffs}(undef, n-1)
     m = Vector{TCoeffs}(undef, n)
-    for k in 1:n-1
-        Δk = (y[k+1] - y[k]) / (x[k+1] - x[k])
-        Δ[k] = Δk
+    for k ∈ eachindex(Δ)
+        Δ[k] = (y[k+1] - y[k]) / (x[k+1] - x[k])
     end
     Γ = [3Δ[1] - 2Δ[2]; 2Δ[1] - Δ[2]; Δ; 2Δ[n-1] - Δ[n-2]; 3Δ[n-1] - 2Δ[n-2]]
-    for k in 1:n
+    for k ∈ eachindex(m)
         δ = abs(Γ[k+3] - Γ[k+2]) + abs(Γ[k+1] - Γ[k])
         if δ > zero(δ)
             α = abs(Γ[k+1] - Γ[k]) / δ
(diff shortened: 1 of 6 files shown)


src/monotonic/monotonic.jl
```diff
@@ -9,11 +9,14 @@
         "Cardinal"              cubic cardinal splines, uses tension parameter which must be between [0,1]
                                 cubin cardinal splines can overshoot for non-monotonic data
                                 (increasing tension decreases overshoot)
+        "Akima"                 monotonic - tagents are dertermined at each given point locally,
```
mkitti, line 12:
```suggestion
        "Akima"                 monotonic - tangents are determined at each given point locally,
```

spelling

src/monotonic/monotonic.jl
```diff
@@ -279,6 +296,33 @@ function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
...
+
+    n = length(x)
+    Δ = Vector{TCoeffs}(undef, n-1)
+    m = Vector{TCoeffs}(undef, n)
+    for k in 1:n-1
+        Δk = (y[k+1] - y[k]) / (x[k+1] - x[k])
+        Δ[k] = Δk
+    end
+    Γ = [3Δ[1] - 2Δ[2]; 2Δ[1] - Δ[2]; Δ; 2Δ[n-1] - Δ[n-2]; 3Δ[n-1] - 2Δ[n-2]]
+    for k in 1:n
```
mkitti, line 314:
```suggestion
    for k in eachindex(m)
```

src/monotonic/monotonic.jl
```diff
@@ -9,11 +9,14 @@
         "Cardinal"              cubic cardinal splines, uses tension parameter which must be between [0,1]
                                 cubin cardinal splines can overshoot for non-monotonic data
                                 (increasing tension decreases overshoot)
+        "Akima"                 monotonic - tagents are dertermined at each given point locally,
+                                results are close to manual drawn curves
```
mkitti, line 13: What does "manually drawn curves" mean? Is that a quote from the article?

src/monotonic/monotonic.jl
```diff
@@ -279,6 +296,33 @@ function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
...
+    # based on Akima (1970),
+    # "A New Method of Interpolation and Smooth Curve Fitting Based on Local Procedures",
+    # doi:10.1145/321607.321609.
+
+    n = length(x)
+    Δ = Vector{TCoeffs}(undef, n-1)
+    m = Vector{TCoeffs}(undef, n)
+    for k in 1:n-1
+        Δk = (y[k+1] - y[k]) / (x[k+1] - x[k])
+        Δ[k] = Δk
```
mkitti, line 311:
```suggestion
        Δ[k] = (y[k+1] - y[k]) / (x[k+1] - x[k])
```

I think this can be a single line.

src/monotonic/monotonic.jl
```diff
@@ -279,6 +296,33 @@ function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
     return (m, Δ)
 end
 
+function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
+    y :: AbstractVector{TEl}, method::AkimaMonotonicInterpolation) where {TCoeffs, TEl}
+
+    # based on Akima (1970),
+    # "A New Method of Interpolation and Smooth Curve Fitting Based on Local Procedures",
+    # doi:10.1145/321607.321609.
+
```
mkitti, line 305: Check if `x` and `y` are the same the length. Throw an `ArgumentError` otherwise. Maybe this is already done elsewhere?

src/monotonic/monotonic.jl
```diff
@@ -279,6 +296,33 @@ function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
...
+    Δ = Vector{TCoeffs}(undef, n-1)
+    m = Vector{TCoeffs}(undef, n)
+    for k in 1:n-1
+        Δk = (y[k+1] - y[k]) / (x[k+1] - x[k])
+        Δ[k] = Δk
+    end
+    Γ = [3Δ[1] - 2Δ[2]; 2Δ[1] - Δ[2]; Δ; 2Δ[n-1] - Δ[n-2]; 3Δ[n-1] - 2Δ[n-2]]
+    for k in 1:n
+        δ = abs(Γ[k+3] - Γ[k+2]) + abs(Γ[k+1] - Γ[k])
+        if δ > zero(δ)
```
mkitti, line 316:
```suggestion
        if δ ≉ zero(δ)
```

That is `\napprox<tab>`. We should be careful of very small values of `δ` as well. `δ` should be nonnegative.

mkitti (changes requested):
A few minor spelling corrections, comments, and suggestions.

src/monotonic/monotonic.jl
wg030: Yes, the phrase "manually drawn cruves" comes indeed from a sentence in the article:

> Comparison indicates that the curve obtained by this new method is closer to a manually drawn curve than those drawn by other mathematical methods.

wg030: So I think it's ok here, too. But I will change "manual" to "manually", of course!

src/monotonic/monotonic.jl
wg030: I had the same thought when I copied the `calcTangents` from another method in order to modify it for the Akima interpolation, but I also tought there was a good reason for that since all other `calcTangents` methods do it this way!

src/monotonic/monotonic.jl
wg030: I usually use `eachindex`, too. But it has not been used here by all other `calcTangents` methods either and I simplied copied the code from one method to modify it for the Akima method.
Maybe `1:n` is used to make the code easier too read and/or because the package is designed in such a way that it does not give you any advantage in terms of performance?

src/monotonic/monotonic.jl
wg030: You are right here, the check is performed somewhere else:
The not exported `calcTangents` method is called while calling the exported `interpolate` method, which in turn calls `check_monotonic`. Hence when we call `interpolate( [1,2,3,4,5], [1,2,3,4,5,6], AkimaMonotonicInterpolation())` we will get the error

> ERROR: DimensionMismatch: knot vector must have the same axes as the corresponding array

Therefore everything is already fine.

wg030:
@mkitti Thanks for reviewing my contribution!
I wrote down my thoughts on your comments. For now, I will correct the spelling misstakes and we can discuss the other suggestions.

src/monotonic/monotonic.jl
mkitti: I don't see a reason for it. I'll change the others later, or you can. Please change this.

src/monotonic/monotonic.jl
mkitti: I'll change the others later. Please change this.

src/monotonic/monotonic.jl
```diff
@@ -279,6 +296,33 @@ function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
...
 end
 
+function calcTangents(::Type{TCoeffs}, x::AbstractVector{<:Number},
+    y :: AbstractVector{TEl}, method::AkimaMonotonicInterpolation) where {TCoeffs, TEl}
+
+    # based on Akima (1970),
+    # "A New Method of Interpolation and Smooth Curve Fitting Based on Local Procedures",
+    # doi:10.1145/321607.321609.
+
+    n = length(x)
```
mkitti, line 306:
```suggestion
```

We can remove this line if we use `eachindex`.

src/monotonic/monotonic.jl
wg030: Thanks for pointing me to the right place in the documentation.
I will have a closer look at that one now, too, and then give me thoughts and that suggestion.

wg030: Alright, I checked out how the `Base.isapprox` function works now.
I figured that due to the statement

> Note that `x ≈ 0` (i.e., comparing to zero with the default tolerances) is equivalent to `x == 0` since the default `atol` is `0`.

we have a situation here where `δ > zero(δ)` is logically and numerically equivalent to `δ ≉ zero(δ)` because we know that  `δ = abs(Γ[k+3] - Γ[k+2]) + abs(Γ[k+1] - Γ[k])` yields a nonnegative value.

---

Therefore I looked at the performance using the `BenchmarkTools` package. Here are the results:

`julia> @benchmark rand() > 0`:

> BenchmarkTools.Trial: 1000000 samples with 1000 evaluations.
>  Range (min … max):  6.000 ns … 333.200 ns  ┊ GC (min … max): 0.00% … 0.00%
>  Time  (median):     6.200 ns               ┊ GC (median):    0.00%
>  Time  (mean ± σ):   6.695 ns ±   1.945 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%
> 
>   █▇▇▄▄   ▄▃  ▁▃   ▁                                  ▄▄ ▄▄▁▂ ▄
>   ██████▇████████▇▇█▇▆▆▅▆▅▅▅▅▅▅▆▅▅▄▅▄▅▅▄▄▄▄▄▄▄▅▄▄▄▃▄▄▆██▅████ █
>   6 ns         Histogram: log(frequency) by time      12.7 ns <
> 
>  Memory estimate: 0 bytes, allocs estimate: 0.

---

> `julia> @benchmark rand() ≉ 0`:
> 
> BenchmarkTools.Trial: 1000000 samples with 1000 evaluations.
>  Range (min … max):  7.100 ns … 172.600 ns  ┊ GC (min … max): 0.00% … 0.00%
>  Time  (median):     7.300 ns               ┊ GC (median):    0.00%
>  Time  (mean ± σ):   8.034 ns ±   2.332 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%
> 
>   █▆▇▄  ▁▃▃  ▂   ▁                          ▄▄▁▅▄ ▂       ▂ ▁ ▄
>   █████▆███████▇▇█▇▅▆▅▅▅▅▅▅▆▅▅▅▅▄▅▄▅▄▅▄▅▄▄▅▆███████▅▇▅█▆█▇█▇█ █
>   7.1 ns       Histogram: log(frequency) by time      15.8 ns <
> 
>  Memory estimate: 0 bytes, allocs estimate: 0.

---

So, looking at the results in terms of performance no change should be made here.
Is there any good reason that is more important than the performance so that we should still change this statement?

mkitti: Have you checked
```!(Γ[k+3] ≈ Γ[k+2] && Γ[k+1] ≈ Γ[k])```

mkitti (approved):

src/monotonic/monotonic.jl
wg030: Alright,
I created a set of 1_000 random `x` and `y` vectors of length 20 and ran a test function that calls `calcTangents` on all that 1_000 pairs of `(x,y)`. I did that for the following four cases using the same random vectors:
1. `δ > zero(δ)`
2. `!(Γ[k+3] ≈ Γ[k+2] && Γ[k+1] ≈ Γ[k])`
3. `Γ[k+3] ≉ Γ[k+2] && Γ[k+1] ≉ Γ[k]`
4. `Γ[k+3] ≈ Γ[k+2] && Γ[k+1] ≈ Γ[k]` + switching the contents of if and else

Here are the min, median and mean times for 10_000 samples with 1 evaulation of the test function:
1. min:  2.151 ms, median:  2.269 ms, mean:  2.472 ms
2. min:  2.248 ms, median:  2.362 ms, mean:  2.562 ms
3. min:  2.331 ms, median:  2.446 ms, mean:  2.624 ms
4. min:  2.232 ms, median:  2.342 ms, mean:  2.529 ms

We can see that `δ > zero(δ)` is a little superior to all other variants.

wg030:
> ready to merge?

Today I thought about adding some convenience constructors for all those monotonic interpolations types, too, which I might do in the next few days.
But if you want to merge this PR already, go ahead, it is already mergeable!

mkitti:
> > ready to merge?
> 
> Today I thought about adding some convenience constructors for all those monotonic interpolations types, too, which I might do in the next few days. But if you want to merge this PR already, go ahead, it is already mergeable!

Could you sketch out the thoughts for convenience constructors in an issue so we can discuss?

wg030:
> Could you sketch out the thoughts for convenience constructors in an issue so we can discuss?

Sure! I would suggest that we merge this one and I will open a new issue for the discussion of adding convenience constructors outlining my idea!

mkitti:
merged

mkitti:
I need to register it.

dkarrasch:
If that's the case, then you could file an issue or comment on the last commit to master and suggest registering the current state. The registration process is highly automated and efficient, but it requires a maintainer to launch it.

wg030:
> If that's the case, then you could file an issue or comment on the last commit to master and suggest registering the current state. The registration process is highly automated and efficient, but it requires a maintainer to launch it.

Ah ok, thanks for the clearfication. I think i got it now. It's bascially an automated process which is triggered by the maintainer by commenting 

> @JuliaRegistrator register

to the latest commit, isn't it? At least that's what happened here and now `(@v1.8) pkg> up` seems to be up to date with the content of this github repo.

PR #526, opened 2022-10-07
switch from StaticArrays to StaticArraysCore

sjkelly:
This package primarily uses the SVector as a return type, instead of using the operations on said SVector. Therefore StaticArraysCore provides immense improvements to loads time:

before:

```
julia> @time using Interpolations
  1.171881 seconds (2.33 M allocations: 172.010 MiB, 2.03% compilation time: 8% of which was recompilation)
```

after:

```
julia> @time using Interpolations
  0.312613 seconds (387.59 k allocations: 28.954 MiB, 7.65% compilation time: 9% of which was recompilation)
```

I would argue though that an NTuple would be sufficent return type in most cases. Though that is a breaking change and can remain an idea for a major release.


sjkelly:
Okay I realize while fixing the docs builds that StaticArraysCore removed some constructors and converts. There are some holes in coverage that we will need to plug to make sure that all calls to `SVector` are okay.

mkitti:
We should revisit this.

mkitti:
We may be using it internally as well. I'll have to double check.

mkitti:
I think this is a good idea, but this is going to need more refactoring for it to work.

PR #549, opened 2023-04-26
use an extension instead of Requires on 1.9

KristofferC:


mkitti:
I believe the invalidation CI failure is more related to SnoopCompile than anything else at the moment.

mkitti:
nightly failures are due to Zygote not precompiling

mkitti (approved):

timholy:
> I believe the invalidation CI failure is more related to SnoopCompile than anything else at the moment.

Yep. Fixed now.

PR #568, opened 2023-11-23
Some recursion tuning to allow more eager inference.

N5N3:
Before julia 1.11, indirect recursion might cause unwanted type widening and leading to bad inference.
`itp(...)`, `gradient(itp, ...)`, `hessian(itp, ...)` would therefore become unstable if `itp` has some `NoInterp` dim.
This PR fix it, MWE has been added to Test.


mkitti:
Thanks

PR #579, opened 2024-01-29
Allow integers only for getindex on AbstractInterpolation

mkitti:
Drop deprecated getindex with non-integers


mkitti:
I am anticipating complaints about this. I am considering moving this functionality to a compatibility package and/or package extension.

PR #602, opened 2024-08-27
Add 3rd and 4th order convergent cubic convolution interpolation methods with continuous first derivates in any number of dimensions, including extrapolation support

NikoBiele:
Hi interpolators!

I noticed there had been a wish for gridded spline interpolations for a while.
I do not know how to accomplish gridded spline interpolations, but I had an implementation of a cubic convolution algorithm which I have translated. It implements "Cubic Convolution Interpolation for Digital Image
Processing" by Robert G. Keys (1981) IEEE Transactions On Acoustics, Speech, And Signal Processing, vol. ASSP-29, no. 6, December 1981.

I have attempted to implement it according to the package standards with a constructor and an interpolator (it also supports extrapolation). It supports interpolation in any number of dimensions, it has continuous first derivatives and feature both a 3rd and 4th order convergent implementation. The fact that this algorithm does not require continuous second derivatives can be an advantage in some cases, if one desires less overshoot than cubic splines achieves. This is especially true for the 3rd order cubic convolution interpolation, and as such it may be viewed as a compromise between splines and linear interpolation.

Kind regards, Nikolaj

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@
 docs/build/
 Manifest.toml
 .vscode
-temp_tester.jl
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 docs/build/
 Manifest.toml
 .vscode
+temp_tester.jl
--- a/src/convolution.jl
+++ b/src/convolution.jl
@@ -1,10 +1,10 @@
-export ConvolutionInterpolation, ConvolutionMethod
+export CubicConvolutionalInterpolation, ConvolutionMethod
 
-struct ConvolutionKernel end
+struct CubicConvolutionalKernel end
 
 struct ConvolutionMethod <: InterpolationType end
 
-function (::ConvolutionKernel)(s)
+function (::CubicConvolutionalKernel)(s)
     s_abs = abs(s)
     if s_abs < 1.0
         return (3/2)*s_abs^3 - (5/2)*s_abs^2 + 1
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@
 docs/build/
 Manifest.toml
 .vscode
+temp_tester.jl
```
mkitti, line 5:
```suggestion
```

Please revert this.

mkitti:
I like where this is heading. Could you increase test coverage, please?

I am wondering about the name `ConvolutionInterpolation` and `ConvolutionalKernel` because this could refer to any interpolation that uses convolution rather than specifically smooth cubic convolution. Could you rename it be `CubicConvolutionalInterpolation` and `CubicConvolutionalKernel`?

src/convolution.jl
```diff
@@ -0,0 +1,132 @@
...
+    if N == 1
+        knots_new = (knots[1][1]-h[1]:h[1]:knots[1][end]+h[1],)
+    else
+        knots_new = (knots[1][1]-h[1]:h[1]:knots[1][end]+h[1], knots[2][1]-h[2]:h[2]:knots[2][end]+h[2])
+    end
+
+    ConvolutionInterpolation{T,N,typeof(coefs),typeof(it),typeof(knots)}(coefs, (knots_new), it, h, ConvolutionKernel())
+end
+
+function create_coefs(knots::Tuple{AbstractVector}, vs::AbstractVector{T}) where T
```
mkitti, line 43: This name seems too general. Perhaps rename is to `create_cubic_convolutional_coefs`?

.gitignore
NikoBiele: Got it. 👍

src/convolution.jl
NikoBiele: Makes sense, done.

NikoBiele:
I also added 3D support and implemented your requested naming updates.

NikoBiele:
I will look into testing in the weekend. :)

NikoBiele:
Tomorrow is dedicated to writing tests. :)

NikoBiele:
I will write some tests now :)

NikoBiele:
In the spirit of the paper mentioned above I derived and tested several of my own kernels! I finally found a really good and smooth one with continuous second derivatives which is 3rd order accurate and like the other ones here it only requires the original values as coefficients, as well as boundary conditions. :) I will implement it here when I find the time. :)

NikoBiele:
I will close this pull request and instead make my own package implementing this methodology. :)

mkitti:
OK. I'm a little confused. Did you finish writing your tests?

NikoBiele:
I added tests above. But since nobody replied I interpreted this as a lack of interest. Also I'm getting very good at deriving these kernels, so I think this idea should have it's own space, so to speak. :)

mkitti:
It's not lack of interest. I just did not know you were ready for the pull request to be reviewed again. Your last comment implied that you were still working on something. If you want to reopen, I can take a look. We could also consider restructuring this as an extension that will load when your package is present.

NikoBiele:
Thanks. I'm sorry that I misinterpreted your intent. No hard feelings at all. But I still thinks that this deserves its own package, as the progress I'm seeing currently suggests that my method will outperform cubic B-splines on important parameters (on equally spaced data), such as setup time (no linear system), accuracy (maximum absolute error, continuity and order of accuracy), simplicity and dimensionality, while having very similar execution time (hopefully on the order of 10 microseconds for 1d). Maybe the extension is a nice option. :)

PR #607, opened 2024-11-23
fix Vararg

oscardssmith:


mkitti:
Is this ambiguity introduced by this change?

```
   Evaluated: isempty(Tuple{Method, Method}[(
checkbounds(::Type{Bool}, itp::AbstractInterpolation, x::Vararg{Union{Number, AbstractVector}, N}) where N @ Interpolations ~/work/Interpolations.jl/Interpolations.jl/src/Interpolations.jl:425,
checkbounds(::Type{Bool}, A::AbstractArray, i::AbstractVector{Bool}) @ Base multidimensional.jl:837
)])
```

PR #608, opened 2024-12-06
Minor documentation fix: do not reference the deprectaed LinearInterpolation

Liozou:
As the commit states, this simply fixes an old reference to `LinearInterpolation` and replaces it with `linear_interpolation`.


mkitti:
Thank you!

PR #615, opened 2025-04-17
method to lift checkbounds ambiguity

ajwheeler:
This fixes the test failure due to ambiguous `checkbounds` methods.  After merging this and #614, all tests should pass on master with julia 1.11.


mkitti:
Looks promising. Thank you.

I will check in later today.

mkitti:
Would tests pass on Julia 1.10 LTS?

ajwheeler:
Confirmed that this fixes the method ambiguity on 1.10.

PR #635, opened 2025-10-24
WIP: reduced precision for lanczos_opencv

bjarthur:
using Float32 for Lanczos is ~2x faster and uses ~1/2x as much memory as the current Float64.

this PR currently uses whatever precision was input as the precision the internal calculations are performed with.  i could also imagine specifying the type used for internal computations in the type (e.g. `struct Lanczos4OpenCV{T} <: AbstractLanczos end`) to separate it from the input.

i'm also curious where there is a more clever way to cast `l4_2d_cs` at compile time so as not to incur runtime penalities.

let me know what you think and i'll add some tests and docs.

```
julia> using Interpolations, BenchmarkTools

julia> x=rand(1_000_000);

julia> @benchmark Interpolations._lanczos4_opencv.(x)
BenchmarkTools.Trial: 231 samples with 1 evaluation per sample.
 Range (min … max):  19.061 ms … 83.561 ms  ┊ GC (min … max): 5.34% … 74.95%
 Time  (median):     21.303 ms              ┊ GC (median):    2.90%
 Time  (mean ± σ):   21.654 ms ±  4.135 ms  ┊ GC (mean ± σ):  4.35% ±  5.10%

                       ▅█▄▃                                    
  ▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▃▃▃▃▅▇█████▄▄▄▃▃▃▃▂▂▁▁▁▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▂▁▁▁▁▂ ▃
  19.1 ms         Histogram: frequency by time        24.8 ms <

 Memory estimate: 61.05 MiB, allocs estimate: 4.

julia> x=rand(Float32, 1_000_000);

julia> @benchmark Interpolations._lanczos4_opencv.(x)
BenchmarkTools.Trial: 387 samples with 1 evaluation per sample.
 Range (min … max):  12.078 ms … 76.608 ms  ┊ GC (min … max): 0.00% … 83.87%
 Time  (median):     12.695 ms              ┊ GC (median):    3.05%
 Time  (mean ± σ):   12.928 ms ±  3.290 ms  ┊ GC (mean ± σ):  4.56% ±  4.80%

     ▂       ▄▅▇██▇▅▂                                          
  ▆▇▇██▄▆▆▄▇██████████▄█▄▇▄▆▁▄▁▇▄▄▁▄▄▄▇▄▆▁▁▄▁▁▁▁▆▁▁▁▁▁▁▄▄▁▁▁▄ ▇
  12.1 ms      Histogram: log(frequency) by time      14.5 ms <

 Memory estimate: 30.53 MiB, allocs estimate: 4.
```

### DIFF
--- a/src/lanczos/lanczos_opencv.jl
+++ b/src/lanczos/lanczos_opencv.jl
@@ -54,7 +54,8 @@ struct Lanczos4OpenCV <: AbstractLanczos end
 
 degree(::Lanczos4OpenCV) = 4
 
-value_weights(::Lanczos4OpenCV, δx) = _lanczos4_opencv(δx)
+value_weights(::Lanczos4OpenCV, δx::T) where {T} =
+        _lanczos4_opencv(float(T), float(T).(l4_2d_cs), δx)
 
 # s45 = sqrt(2)/2
 const s45 = 0.70710678118654752440084436210485
@@ -65,8 +66,7 @@ const s45 = 0.70710678118654752440084436210485
 const l4_2d_cs = SA[1 0; -s45 -s45; 0 1; s45 -s45; -1 0; s45 s45; 0 -1; -s45 s45]
 
 
-function _lanczos4_opencv(δx::T) where T
-    F = float(T)
+function _lanczos4_opencv(::Type{F}, l4_2d_cs, δx) where {F}
     p_4 = π / F(4)
     y0 = -(δx + 3) * p_4
     s0::F, c0::F = sincos(y0)
--- a/src/lanczos/lanczos_opencv.jl
+++ b/src/lanczos/lanczos_opencv.jl
@@ -65,20 +65,21 @@ const s45 = 0.70710678118654752440084436210485
 const l4_2d_cs = SA[1 0; -s45 -s45; 0 1; s45 -s45; -1 0; s45 s45; 0 -1; -s45 s45]
 
 
-function _lanczos4_opencv(δx)
-    p_4 = π / 4
+function _lanczos4_opencv(δx::T) where T
+    F = float(T)
+    p_4 = π / F(4)
     y0 = -(δx + 3) * p_4
-    s0, c0 = sincos(y0)
+    s0::F, c0::F = sincos(y0)
     cs = ntuple(8) do i
         y = (δx + 4 - i) * p_4
         # Improve precision of Lanczos OpenCV4 #451, avoid NaN
         if iszero(y)
-            y = eps(oneunit(y))/8
+            y = eps(oneunit(y))/F(8)
         end
         # Numerator is the sin subtraction identity
         # It is equivalent to the following
         # f(δx,i) = sin( π/4*( 5*(i-1)-δx-3 ) )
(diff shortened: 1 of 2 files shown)


bjarthur:
second commit makes it slightly faster:

```
julia> x=rand(Float32, 1_000_000);

julia> @benchmark Interpolations.value_weights.(Ref(Lanczos4OpenCV()), x)
BenchmarkTools.Trial: 429 samples with 1 evaluation per sample.
 Range (min … max):  10.796 ms … 75.222 ms  ┊ GC (min … max): 0.00% … 85.30%
 Time  (median):     11.479 ms              ┊ GC (median):    3.84%
 Time  (mean ± σ):   11.664 ms ±  3.107 ms  ┊ GC (mean ± σ):  5.43% ±  4.66%

               ▂█▆▃▄▄▅▇▃▁                                      
  ▃▃▃▃▃▃▁▃▃▃▄▅▆██████████▆▃▃▃▄▃▁▁▂▃▁▁▃▃▂▁▁▃▁▂▁▃▁▂▂▁▁▁▁▁▁▁▁▂▂▂ ▃
  10.8 ms         Histogram: frequency by time          13 ms <

 Memory estimate: 30.53 MiB, allocs estimate: 5.
```

bjarthur:
latest commit preserves the original implementation of lanczos_opencv, and adds a new version which can use Float32 internally to mirror openCV identically.

bjarthur:
@mkitti can you please review?

src/lanczos/lanczos_opencv.jl
```diff
@@ -84,3 +92,42 @@ function _lanczos4_opencv(δx)
...
+struct Lanczos4OpenCVFaithful{T} <: AbstractLanczos end
+Lanczos4OpenCVFaithful() = Lanczos4OpenCVFaithful{Float32}()
+
+degree(::Lanczos4OpenCVFaithful) = 4
+
+value_weights(::Lanczos4OpenCVFaithful{T}, δx) where {T} =
+        _lanczos4_opencv_faithful(float(T), float(T).(l4_2d_cs), δx)
+
+# main differences with `_lanczos4_opencv` are (1) the criterion for preventing a
+# division by zero is `< pi/4 * 1e-6` (instead of `iszero`) and (2) the resulting
```
mkitti, line 113: Can you justify where these new constants are coming from? Is it going to work for Float16?

I wonder if we should use `isapprox`?

mkitti (commented):
I have a few questions on how general this is and where the constants are coming from.

mkitti:
Any update here?

src/lanczos/lanczos_opencv.jl
bjarthur: these new constants come directly from the OpenCV implementation.  i don't know about Float16.  the whole point of this PR is to copy the OpenCV algorithm verbatim, including the use of Float32.  so that one can reproduce results from and OpenCV workflow in julia with nearly identical results at the bit level.  the existing non-faithful code here tries to improve on their approach.  that's not what this new method is intended for.

mkitti:
Is this still a WIP?

